PR #5, opened 2018-03-07
Part 1 of many for porting keystone and sessions

brennancheung:


brennancheung:
I think this is ready for review now.  I don't want each PR to be too huge.  Did a lot of refactoring and added the ability for plugins to register their own redux reducers.

brennancheung:
The end-to-end tests are not quite there yet.  I was thinking of doing that and the integration tests in a separate PR.

kevinylee1991:
This'll take me awhile to go through -- i think i need to play around with the new code first and get used to react again before I can give a real review

PR #9, opened 2018-03-28
Users and tenants

brennancheung:
The following was added in this PR:

- Simulator
  - Users
  - Tenants
- Tests
- Generic React table component

### DIFF
--- a/src/openstack-simulator/models/User.js
+++ b/src/openstack-simulator/models/User.js
@@ -7,11 +7,14 @@ const coll = () => context.users
 class User extends ActiveModel {
   constructor (params = {}) {
     super(params)
-    this.roles = params.roles || []
-    this.username = params.username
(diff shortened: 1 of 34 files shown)


src/openstack-simulator/models/User.js
```diff
@@ -7,11 +7,14 @@ const coll = () => context.users
 class User extends ActiveModel {
   constructor (params = {}) {
     super(params)
-    this.roles = params.roles || []
-    this.username = params.username
+    this.displayname = params.displayname || params.name || params.email || ''
```
kevinylee1991: will they get ids later?

brennancheung: `ActiveModel` creates the `id` automatically.  The parent constructor will generate the `id`.

PR #11, opened 2018-04-10
Set up Storybook further and add User examples

brennancheung:
This adds some nice additions to [Storybook](https://storybook.js.org/).

We will need to have some discipline to make sure components aren't too coupled.  Adding users to the Storybook made me realize that we should making components that can work without being connected to side effect functionality.  We can stub out functionality like getting data and submitting.

Also, we should be making frequent use of the [Container](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) architectural pattern.

storybook


brennancheung:
Form validation is definitely on the roadmap but not sure exactly when.  I'm also not sure if we want to use `redux-form` or not.  It does make things nice and uniform but it creates some architectural restrictions and does bind us to `redux` unnecessarily.  I never liked triggering a `dispatch` once every keystroke.  The `recompose` library might also be worth looking into.

PR #13, opened 2018-04-12
Added basic openstack flavors support to the UI, storybook, and simul…

kevinylee1991:
…ator

### DIFF
--- a/src/app/plugins/openstack/api/nova.js
+++ b/src/app/plugins/openstack/api/nova.js
@@ -1,0 +1,20 @@
+import http from '../../../util/http'
+
+const v2Base = '/nova/v2.1'
+const authHttp = http.authenticated.openstack
+
+export const createFlavor = (flavor, tenantId) => {
+  const body = {
+    flavor: {
+      name: flavor.name,
+      ram: 1024,
+      vcpus: 2,
+      disk: 10,
+    }
+  }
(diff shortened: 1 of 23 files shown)


src/app/plugins/openstack/api/nova.js
```diff
@@ -0,0 +1,21 @@
...
+const authHttp = http.authenticated.openstack
+
+export const createFlavor = (flavor, tenantId) => {
+  const body = {
+    flavor: {
+      name: flavor.name,
+      ram: 1024,
+      vcpus: 2,
+      disk: 10,
+      description: 'test description',
```
brennancheung: Don't hardcode this.  Maybe `flavor.description || ''`

src/app/plugins/openstack/components/FlavorsPage.js
```diff
@@ -0,0 +1,38 @@
...
+import FlavorsListContainer from './flavors/FlavorsListContainer'
+
+function mapStateToProps (state, ownProps) {
+  const { flavors } = state.openstack
+  return {
+    flavorsLoaded: flavors.flavorsLoaded,
+  }
+}
+
+const Loader = () => <div>Loading...</div>
```
brennancheung: Let's make this a common component.  We're going to be using it a lot.

PR #58, opened 2018-04-18
Lots of misc functionality for tenants/users/flavors

brennancheung:
Just fleshing out more functionality for Users, Tenants, and Flavors.

### DIFF
--- a/src/app/plugins/openstack/api/keystone.js
+++ b/src/app/plugins/openstack/api/keystone.js
@@ -97,6 +97,8 @@ export function createTenant (project) {
   return authHttp.post(`${v3Base}/projects`, body)
 }
 
+export const deleteTenant = tenantId => authHttp.delete(`${v3Base}/projects/${tenantId}`)
+
 /*
 import {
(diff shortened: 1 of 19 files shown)


src/app/plugins/openstack/api/keystone.js
```diff
@@ -84,6 +84,21 @@ export const createUser = user => {
...
 export const getUsers = () => authHttp.get(`${v3Base}/users`).then(x => x.users)
 
+export function createTenant (project) {
+  const body = {
+    project: {
+      enabled: project.enabled,
+      name: project.name,
+      description: project.description,
+      domain_id: 'default',
+      is_domain: false
```
kevinylee1991: no trailing comma -- not sure if we actually care though

brennancheung: I generally add them but forgot to here.  Trailing commas are optional in Standard.js.  No set rule for it.  Developer is free to go either way.  I'm not concerned about being uniform about this.  It doesn't have any real impact.

PR #65, opened 2018-04-30
Bootstrap UI selection with a loader

brennancheung:
This PR adds support for selecting different versions of the UI.

It creates **breaking changes**.  Developers will need to add `?dev=true` the first time to update to the new process.

See the `README.md` below for what this is about.

Deploying these different versions is still pending.  Right now we only have `?branch=master` working in CI/CD.

We will need to come up with a process to deploy specific releases to S3.

I believe this approach takes a balanced approach to being flexible and simple.

If anyone has other requirements, concerns, or ideas for improvements, let's discuss.

This `loader.js` is deliberately written in *old school* JS and exists outside of the bundling process.  It is deployed as a static file without any compilation steps.


tobensg:
It looks good to me.

PR #69, opened 2018-05-08
Added padding to main content div to account for navbar

kevinylee1991:
Added it to the css file instead of js because I assume that MuiThemeProvider theme needs to be defined outside of the component we try to use withStyles on.

Other way could be extracting _main-content div as its own react component. Which would be the better implementation?

### DIFF
--- a/src/app/app.css
+++ b/src/app/app.css
@@ -23,6 +23,10 @@ body {
   grid-area: main;
 }
 
+#_main-content {
+  padding-top: 64px;
+}
+
 #main-footer {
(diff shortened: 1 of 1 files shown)


brennancheung:
Not sure the best approach yet.  Are you seeing the problem on your end?  I'm not seeing it anymore on my end and I don't have this branch.  I'm worried we might be doubling the padding.

kevinylee1991:
My side still has the problem -- it requires this css change. Where does the padding come from on your UI?

PR #70, opened 2018-05-08
Networks

kevinylee1991:
Currently only takes a name property (can add support for cidr, etc later).

One question I have is where should helper functions be located? Ex. cidr validations, IP address validations. Also where should form validations go in general?

### DIFF
--- a/src/app/plugins/openstack/api/neutron.js
+++ b/src/app/plugins/openstack/api/neutron.js
@@ -1,0 +1,17 @@
+import http from '../../../util/http'
+
+const v2Base = '/neutron/v2.0'
+const authHttp = http.authenticated.openstack
+
+export const createNetwork = (network) => {
+  const body = {
+    network: {
+      name: network.name,
+    }
+  }
+  return authHttp.post(`${v2Base}/networks`, body).then(json => json.network.id)
+}
+
+export const deleteNetwork = (networkId) => authHttp.delete(`${v2Base}/networks/${networkId}`)
+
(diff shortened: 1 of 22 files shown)


src/app/plugins/openstack/api/neutron.js
```diff
@@ -0,0 +1,17 @@
+import http from '../../../util/http'
+
+const v2Base = '/neutron/v2.0'
```
blake: This should not be hard coded. It should come from the service catalog by looking up the "network" service type, or "neutron" service name.

kevinylee1991: Oh yeah -- that should be on the road map soon. We need to add the service catalog to the simulator before we can make it non-hardcoded.

brennancheung: For now, you can make a function that returns the hard-coded URL.  Later on, we can make it get it from the catalog.  The function should probably be promised based as well.

src/app/plugins/openstack/components/networks/AddNetworkPage.js
```diff
@@ -0,0 +1,32 @@
...
+const mapStateToProps = state => ({})
+
+@withRouter
+@connect(mapStateToProps)
+class AddNetworkPage extends React.Component {
+  handleSubmit = network => {
+    const { dispatch, history } = this.props
+    try {
+      dispatch(addNetwork(network))
+      history.push('/networks')
```
brennancheung: Might need to be `/ui/openstack/networks`.  Verify this works correctly.

kevinylee1991:
squashed/merged through local cli

PR #74, opened 2018-05-11
Added service catalog to the simulator and UI

kevinylee1991:
One thing that may need to change:

Currently the service catalog is stored in state in the 'openstack' property. However, the service catalog stores services outside of openstack too, like qbert, so it should probably be moved up to root level. I need to work on something for the old codebase now but I wanted to send a review first. I haven't been able to completely work out all the details about the way state is stored just yet.

### DIFF
--- a/src/app/core/common/EnhancedTableHead.js
+++ b/src/app/core/common/EnhancedTableHead.js
@@ -11,18 +11,22 @@ class EnhancedTableHead extends React.Component {
   }
 
   render () {
-    const { columns, onSelectAllClick, order, orderBy, numSelected, rowCount } = this.props
+    const { columns, onSelectAllClick, order, orderBy, numSelected, rowCount, showCheckboxes } = this.props
+
+    const headerCheckbox = showCheckboxes ? (
+      <TableCell padding="checkbox">
+        <Checkbox
+          indeterminate={numSelected > 0 && numSelected < rowCount}
+          checked={numSelected === rowCount}
+          onChange={onSelectAllClick}
+        />
+      </TableCell>
+    ) : null
 
     return (
       <TableHead>
         <TableRow>
-          <TableCell padding="checkbox">
-            <Checkbox
-              indeterminate={numSelected > 0 && numSelected < rowCount}
-              checked={numSelected === rowCount}
-              onChange={onSelectAllClick}
-            />
-          </TableCell>
+          {headerCheckbox}
           {columns.map(column => {
             return (
               <TableCell
@@ -61,6 +65,7 @@ EnhancedTableHead.propTypes = {
   order: PropTypes.string.isRequired,
   orderBy: PropTypes.string.isRequired,
   rowCount: PropTypes.number.isRequired,
+  showCheckboxes: PropTypes.bool,
 }
 
 export default EnhancedTableHead
--- a/src/app/core/common/ListTable.js
+++ b/src/app/core/common/ListTable.js
@@ -127,21 +127,23 @@ class ListTable extends React.Component {
   }
 
   renderRow = row => {
-    const { columns } = this.props
-
+    const { columns, showCheckboxes } = this.props
     const isSelected = this.isSelected(row.id)
+
+    const checkboxProps = showCheckboxes ? {
+      onClick: this.handleClick(row.id),
+      role: 'checkbox',
+      tabIndex: -1,
+      selected: isSelected
+    } : {}
+
     return (
-      <TableRow
-        hover
-        onClick={this.handleClick(row.id)}
-        role="checkbox"
-        tabIndex={-1}
-        key={row.id}
-        selected={isSelected}
-      >
-        <TableCell padding="checkbox">
-          <Checkbox checked={isSelected} />
-        </TableCell>
+      <TableRow hover key={row.id} {...checkboxProps}>
+        {showCheckboxes &&
+          <TableCell padding="checkbox">
+            <Checkbox checked={isSelected} />
+          </TableCell>
+        }
         {columns.map((columnDef, colIdx) =>
           this.renderCell(columnDef, row[columnDef.id]))
         }
@@ -182,6 +184,10 @@ class ListTable extends React.Component {
       columns,
       data,
       title,
+      onAdd,
+      onDelete,
+      showCheckboxes,
+      paginate
     } = this.props
 
     const {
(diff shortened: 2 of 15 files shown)


brennancheung:
We will need to incorporate this into graphql's version of the "store" (`apollo-link-state`).  For now, we should make this generic code that we can use as a library that we can incorporate easier.

brennancheung:
Just a heads up, the `openstack-simulator` folder has been renamed in `master`.  This might cause some conflicts.

src/app/core/common/EnhancedTableHead.js
```diff
@@ -61,6 +65,7 @@ EnhancedTableHead.propTypes = {
   order: PropTypes.string.isRequired,
   orderBy: PropTypes.string.isRequired,
   rowCount: PropTypes.number.isRequired,
+  checkboxes: PropTypes.bool,
```
brennancheung: `checkboxes` could be interpreted to mean different things other than just a boolean.  What about `showCheckboxes`?

src/app/core/common/ListTable.js
```diff
@@ -125,26 +125,36 @@ class ListTable extends React.Component {
...
+          onClick={this.handleClick(row.id)}
+          role="checkbox"
+          tabIndex={-1}
+          key={row.id}
+          selected={isSelected}
+        >
+          <TableCell padding="checkbox">
+            <Checkbox checked={isSelected} />
+          </TableCell>
+          {columns.map((columnDef, colIdx) =>
```
brennancheung: This part is common between both of them.  We can probably clean this up so we don't repeat it.  Perhaps put it in a variable?

Another way might be `const checkboxProps = ... || {}; <TableRow hover {...checkboxProps} />`

src/app/core/common/ListTable.js
```diff
@@ -180,6 +190,8 @@ class ListTable extends React.Component {
       columns,
       data,
       title,
+      checkboxes,
+      pagination
```
brennancheung: I think I like `paginate` better, `pagination` sounds like it should include the page size and current page.

I'm thinking `<ListTable paginate showCheckboxes ... />`.  Although with the defaults, I suppose we wouldn't normally include them.

src/app/core/common/ListTable.js
```diff
@@ -189,16 +201,16 @@ class ListTable extends React.Component {
...
+    const shouldShowPagination = pagination && sortedData.length > this.state.rowsPerPage
 
     return (
       <Paper className={classes.root}>
         <EnhancedTableToolbar
           numSelected={selected.length}
           title={title}
-          onAdd={this.handleAdd}
-          onDelete={this.handleDelete}
+          onAdd={this.onAdd ? this.handleAdd : undefined}
```
brennancheung: `this.onAdd && this.handleAdd`

src/app/plugins/openstack/actions/serviceCatalog.js
```diff
@@ -0,0 +1,12 @@
+import {
```
brennancheung: You factoring out redux to GraphQL in a separate PR?

src/app/plugins/openstack/api/helpers.js
```diff
@@ -0,0 +1,19 @@
+import http from '../../../util/http'
+import store from '../../../store'
+
+const authHttp = http.authenticated.openstack
+
+// service is the service name, eg. 'neutron', 'nova', 'glance', 'qbert'
+// iface is one of 'public', 'internal', 'admin'
+export const getBaseUrl = (service, iface) => {
```
brennancheung: We're moving away from redux so I'm not really concerned but accessing the redux `store` directly is an anti-pattern.  Usually you would use `redux-thunk` and inject `getState` into your function.

Ex: `export const getBaseUrl = (service, iface) => (dispatch, getState) => { ... }`

src/app/plugins/openstack/api/helpers.js
```diff
@@ -0,0 +1,19 @@
...
+  const catalog = store.getState().openstack.serviceCatalog.serviceCatalog
+  const _service = catalog.find(s => { return s.name === service })
+  const _endpoint = _service.endpoints.find(endpoint => { return endpoint.interface === iface })
```
brennancheung: Use shorthand:

`_service.endpoints.find(endpoint => endpoint.interface === iface)`

src/app/plugins/openstack/components/api-access/ApiAccessListContainer.js
```diff
@@ -0,0 +1,68 @@
...
+      <div>
+        <ApiAccessList
+          catalog={catalog}
+        />
+      </div>
+    )
+  }
+}
+
+const GraphqlCrudContainer = ({ query, actions, children, ...props }) => {
```
brennancheung: Let's make a common component called `GraphqlQueryContainer`.  I also have `CRUDListContainer` but it contains too much functionality for this use case.

src/app/plugins/openstack/components/flavors/FlavorsListContainer.js
```diff
@@ -98,7 +98,7 @@ const GraphqlCrudContainer = ({ query, actions, children, ...props }) => {
 const GraphqlFlavorsList = () => (
   <GraphqlCrudContainer query={GET_FLAVORS}>
```
brennancheung: Rename to `GraphqlQueryContainer` and it should be common and imported instead of being duplicated in every file.

src/app/plugins/openstack/reducers/serviceCatalog.js
```diff
@@ -0,0 +1,26 @@
...
+  SET_SERVICE_CATALOG
+} from '../actions/serviceCatalog'
+
+const initialState = {
+  serviceCatalogLoaded: false,
+  serviceCatalog: [],
+}
+
+const reducer = (state = initialState, action) => {
+  const { type, payload } = action // eslint-disable-line no-unused-vars
```
brennancheung: We're removing redux but I don't think we need the `eslint-disable` since you are using all the variables.  Be careful not to just blindly copy and paste code.  It will lead to a messy codebase.

src/openstack-simulator/api/keystone/index.js
```diff
@@ -29,6 +32,8 @@ router.delete('/v3/projects/:projectId', tokenValidator, deleteProject)
 
 router.get('/v3/regions', tokenValidator, getRegions)
 
+router.get('/v3/auth/catalog', tokenValidator, getCatalog)
```
brennancheung: We probably don't need to add the simulated REST endpoints if the UI will only use GraphQL going forward.

src/openstack-simulator/context.js
```diff
@@ -49,6 +51,12 @@ class Context {
   getTenants = () => {
     return Tenant.getCollection().map(x => x.asGraphQl())
   }
+
+  getServiceCatalog = () => {
```
brennancheung: Why are we setting `this.serviceCatalog`?  We should probably only access the catalog using `getServiceCatalog`.  Any time we can avoid setting imperative state and remain more functional there's that less state we need to worry about.

That would simplify it to: `getServiceCatalog = () => Catalog.getCatalog()`

PR #76, opened 2018-05-19
Add tenant delete mutation

brennancheung:
A little more refactoring on the generic CRUD table stuff.  I made it so that it clears the selected items after items are deleted.  Before it would just keep the selection.

I also did some experimenting and figured out how to update the Apollo cache after the delete mutations.  It's a little verbose.  I couldn't find much documentation about it.  I'll see if I can create some helper utilities to clean it up a little bit more.

### DIFF
--- a/src/app/plugins/openstack/components/TenantsPage.js
+++ b/src/app/plugins/openstack/components/TenantsPage.js
@@ -1,39 +1,19 @@
 import React from 'react'
 import { compose, graphql, withApollo } from 'react-apollo'
-import { gql } from 'apollo-boost'
 
+import DisplayError from 'core/common/DisplayError'
 import Loader from 'core/common/Loader'
 import TenantsListContainer from './tenants/TenantsListContainer'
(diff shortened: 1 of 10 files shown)


src/app/plugins/openstack/components/TenantsPage.js
```diff
@@ -1,39 +1,19 @@
...
 
 const TenantsPage =
   ({ data, loading, error, client }) => {
-    console.log(client)
     return (
       <div>
         <h1>Tenants Page</h1>
         {loading && <Loader />}
-        {error && <h3>TODO: error handler</h3>}
+        {error && <DisplayError error={error} />}
```
kevinylee1991: Eventually the error will need to be parsed, maybe use some global placeholder function for now to avoid having to replace later?

brennancheung: This is the global placeholder.  We just need to change `DisplayError` once we want better error reporting.

PR #78, opened 2018-05-22
Edit README.MD based on folder renaming

HF-liu:
I edited README.MD based on renaming of "server" and "devserver", and fixed a few typo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,20 +4,37 @@ This is the new home for all things Platform9 UI related.
 
 
 # Getting started
+**Initializing the App**
 
 `npm install`
 
+**Starting the Server (previous OpenStack Simulator)**
 
-# Loader
+`npm run server`
 
-The loader bootstraps the app loading process.  It is the very first code to
-run and determines which version of the UI to load.
+**Starting the App**
 
-Normally, the UI is loaded from an S3 bucket.  Multiple versions of the UI are
-stored in different folders and the loader selects which one to use.
+`npm run start`
 
-It does this using a combination of query parameters, `localStorage`, and
-defaults (in that order).
+For local dev version, use `localhost:3000?dev=true` as the browser URL. 
+
+
+# Key Elements
+
+## Server
+
+The Server (previous OpenStack Simulator) is used during UI tests served as frontend server and OpenStack environment. The server will expose API's that are equivalent to those of OpenStack. It may not be 100% compatible but it should be enough for testing and development purpose.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,50 +4,52 @@ This is the new home for all things Platform9 UI related.
...
 
-By default it will select the released version (branch mode).
 
-To switch to a specific branch you can use `?branch=[BRANCH]`.
+# Key Elements
 
-To switch to a specific version you can used `?version=vX.Y`.
+## Server
+
+The Server (previous OpenStack Simulator) is used during UI tests served as frontend server and OpenStack environment. The server will expose API's that are equivalent to those of OpenStack. It may not be 100% compatible but it should be enough for testing and development purpose.
```
brennancheung: It's not just limited to UI tests.  It will be the API endpoint for the entire UI.  It will implement a GraphQL endpoint.  Also, it will contain business logic that we don't want to push to the browser.

We should probably list and describe this by its roles:

* A GraphQL API endpoint
* Proxies requests from GraphQL to the OpenStack REST API
* Contains business logic
* Testing simulator mock environment
* Development mock environment

README.md
```diff
@@ -4,50 +4,52 @@ This is the new home for all things Platform9 UI related.
...
+## Loader
 
-# OpenStack Simulator
+The loader bootstraps the app loading process. It is the very first code to run and determines which version of the UI to load.
 
-The OpenStack Simulator is used during UI tests to create a mock OpenStack
-API environment.  The simulator pretends to be OpenStack and expose API's
-that are equivalent.  It may not be 100% compatible but it should be
-enough for testing and development purposes.
+Normally, the UI is loaded from an S3 bucket. Multiple versions of the UI are stored in different folders and the loader selects which one to use.
```
brennancheung: This may be changed.  We're running into some issues with hosting it on S3.  We'll see if we work around it or move to a different solution.

README.md
HF-liu: I edited this part describing the roles.

README.md
```diff
@@ -86,16 +105,16 @@ To run them:
 
 `npm run test:unit`
 
-Unit tests are designed to run very fast.  Anything that can have an external
-side effect is mocked out.  Jest will run the tests in parallel threads.
+Unit tests are designed to run very fast.  Anything that can have an external
```
brennancheung: Remove the trailing whitespace at the end of the lines.

PR #80, opened 2018-05-23
Volume management in Storybook

HF-liu:


src/app/core/common/formatters.js
```diff
@@ -0,0 +1,27 @@
...
+export const formattedValue = (value, unit='') => {
+  const units = ['Bytes', 'KB', 'MB', 'GB', 'TB']
+  const addComma = (num) => (num.toFixed(2).replace(/\B(?=(\d{3})+(?!\d))/g, ','))
+  if (value === undefined || isNaN(value)) return 'Wrong value input.'
+  if (unit === '' || units.indexOf(unit) === -1) {
+    return addComma(value)
+  }
+  if (unit.length>0) {
+    let pos = units.indexOf(unit)
+    while (value>=1024 && pos<units.length-1) {
```
brennancheung: You can use log 1024 and round down instead of this more imperative approach.

src/app/core/common/__tests__/pf9-formatter.test.js
```diff
@@ -0,0 +1,46 @@
...
+    expect(formattedValue(158)).toEqual('158.00')
+    expect(formattedValue(246.21)).toEqual('246.21')
+    expect(formattedValue(12453)).toEqual('12,453.00')
+    expect(formattedValue(10578143.356)).toEqual('10,578,143.36')
+  })
+
+  it('Value with unit', () => {
+    expect(formattedValue(0, 'Bytes')).toEqual('0.00 Bytes')
+    expect(formattedValue(850, 'MB')).toEqual('850.00 MB')
+    expect(formattedValue(1024, 'GB')).toEqual('1.00 TB')
```
brennancheung: This is coming along nicely.

I was thinking this parameter is the output unit you want and you always feed in bytes.  So it should be 1,000 GB, not 1.00 TB.  We will also need something for number of digits after the decimal point.

src/app/core/common/formatters.js
```diff
@@ -0,0 +1,30 @@
+// Converting the value(bytes) to output unit(optional) with commas
+export const formattedValue = (bytes, unit='Bytes', dicimalDigit=2) => {
```
brennancheung: Spelling typo, should be `decimal`.  Let's call it `decimalDigits`.

src/app/core/common/__tests__/pf9-formatter.test.js
```diff
@@ -0,0 +1,43 @@
+import {
```
brennancheung: By convention, the test should be based off the original filename.  So should be `formatters.test.js`.

brennancheung:
This is looking good. `master` has been updated so you will need to rebase this PR in order to merge it.  We can either merge this and create another PR for the rest of the volume functionality or you can continue working in this PR.

I think it might be a little cleaner to open a separate PR though as this PR already has a lot of history in it.

PR #82, opened 2018-05-25
Volume GraphQL

HF-liu:
Volume/Cinder in GraphQL


src/server/presets/base.js
```diff
@@ -19,6 +20,11 @@ function loadPreset () {
   new Flavor({ name: 'm1.medium', ram: 4096, disk: 40, vcpus: 2 })
   new Flavor({ name: 'm1.large', ram: 8192, disk: 80, vcpus: 4 })
   new Flavor({ name: 'm1.xlarge', ram: 16384, disk: 160, vcpus: 8 })
+  new Volume({ name: 'TestVolume1', description: 'Lalala', type: 'SOF', metadata: 'ad', size: 15, sizeUnit: 'GB', bootable: false, status: 'available', tenantId: 'a', tenant: 'DEV1', source: 'Image', host: '', instance: 'It1', instanceId: '', device: 'Nothing', attachedMode: 'rw', readonly: false })
```
brennancheung: Might want a factory function to create these.  Very repetitive here.

src/server/models/Volume.js
```diff
@@ -0,0 +1,59 @@
...
+
+  static getCollection = coll
+  static clearCollection = () => coll().splice(0, coll().length)
+  static findById = findById(coll)
+
+  static findByName = name => Volume.getCollection().find(x => x.name === name)
+
+  asJson = () => {
+    return {
+      ...super.asJson(),
```
brennancheung: Hmm, this is repetitive.  I'm thinking we might want to make a utility for this.

Something that allows an API like `return { ...super.asJson(), ...pluckFields(this, 'name', 'description', 'type', ...) }`.

We might want to do something similar for setting things in the `constructor` but not quite sure the API yet.  We can hold this off for later.

src/graphql/openstack/schemas/Volume.graphql
```diff
@@ -0,0 +1,25 @@
+type Volume {
+	id: ID!
```
brennancheung: wrong indentation

src/server/presets/base.js
brennancheung: Also, `presets/base.js` is designed to be the absolute minimum clean OpenStack environment like when a customer first gets their system.  The volume really belongs on `presets/dev` this is where we put things that we probably don't want to create manually when working on the UI.

src/server/api/cinder/deleteVolume.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable no-unused-vars */
```
brennancheung: Can probably get rid of this.  Try to avoid copy / paste when you are first learning this.  Typing stuff in a few times is what causes you to remember it.  But like I said in another comment, we don't need the REST API calls at all.  I doesn't hurt to have additional functionality here.  It's fine to leave it, just don't implement it on any new features going forward.  We only need the GraphQL endpoint.

src/graphql/openstack/schemas/Volume.graphql
brennancheung: Nevermind my comment about the indentation.  I see the same weird thing happening in my PR as well.  Github is formatting it differently that what is in the file for some strange reason.

src/server/presets/base.js
```diff
@@ -20,11 +20,6 @@ function loadPreset () {
   new Flavor({ name: 'm1.medium', ram: 4096, disk: 40, vcpus: 2 })
   new Flavor({ name: 'm1.large', ram: 8192, disk: 80, vcpus: 4 })
   new Flavor({ name: 'm1.xlarge', ram: 16384, disk: 160, vcpus: 8 })
-  new Volume({ name: 'TestVolume1', description: 'Lalala', type: 'SOF', metadata: 'ad', size: 15, sizeUnit: 'GB', bootable: false, status: 'available', tenantId: 'a', tenant: 'DEV1', source: 'Image', host: '', instance: 'It1', instanceId: '', device: 'Nothing', attachedMode: 'rw', readonly: false })
```
brennancheung: I would add a few to the `dev` preset.  Make something up that looks believable.

PR #83, opened 2018-05-26
Full CRUD of flavors in GraphQL

brennancheung:
This PR converts Flavors from using redux to GraphQL.  I also added the corresponding GraphQL endpoints on the server.


HF-liu:
I will update volume parts as well.

PR #85, opened 2018-05-30
Convert users CRUD to GraphQL

brennancheung:


src/app/plugins/openstack/components/users/AddUserForm.js
```diff
@@ -23,7 +27,7 @@ class AddUserForm extends React.Component {
     if (type === 'text' || type === 'password') {
       return (
         <div key={id}>
-          <TextField id={id} type={type} label={label} value={this.state[id]} onChange={this.setField(id)} />
+          <TextField id={id} type={type} label={label} value={this.state[id]} onChange={this.setField(id)} autoComplete="off" />
```
kevinylee1991: what about autoComplete="new-password"?

brennancheung: Nice, that seems to work (at least in Chrome).  Switching it now.

src/server/context.js
```diff
@@ -95,6 +92,29 @@ class Context {
...
 
+  getUsers = () => User.getCollection().map(x => x.asGraphQl())
+
+  createUser = ({ input }) => {
+    const user = new User(input)
+    return user.asGraphQl()
+  }
+
+  updateUser = (id, { input }) => {
+    const user = User.findById(id)
```
HF-liu: User is not actually updated in this function. We just check whether it exists here.

PR #86, opened 2018-05-31
Openstack client API library

brennancheung:
I'm working on pulling the OpenStack REST API into its own library.  I'm structuring it is such a way to be consumed as a library by other projects and not have any coupling with the UI.  That being said, it will be opinionated and make things more convenient.  It abstracts away some of the complexity and maintains its own state.

It's currently a work in progress but I want to put it out there to get some early feedback.

### DIFF
--- a/src/openstack-client/Keystone.js
+++ b/src/openstack-client/Keystone.js
@@ -1,0 +1,44 @@
+import axios from 'axios'
+
+const constructPasswordMethod = (username, password) => ({
+  user: {
+    name: username,
+    domain: { id: 'default' },
+    password
+  }
+})
+
+class Keystone {
+  constructor (client) {
+    this.client = client
+  }
+
+  get endpoint () {
+    return this.client.options.keystoneEndpoint
+  }
+
+  async authenticate (username, password) {
+    const body = {
+      auth: {
+        identity: {
+          methods: ['password'],
+          password: {
+            ...constructPasswordMethod(username, password)
+          }
+        }
+      }
+    }
+
+    try {
+      const response = await axios.post(`${this.endpoint}/v3/auth/tokens?nocatalog`, body)
+      const token = response.data.token.id
+      this.client.unscopedToken = token
+      return token
+    } catch (err) {
+      // authentication failed
+      return null
+    }
+  }
+}
+
+export default Keystone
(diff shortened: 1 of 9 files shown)


brennancheung:
I think this is good enough to merge so that others can start implementing functionality as well.

Note the changes in `config.js`.

You will need to rename `testing` to `test` since it is originally a typo.  There are also some fields added.  And I made it so that username and password can be passed in from an environment variable to keep secrets a little more secure.

You will need to have the simulator up and running in order to run the tests for it.

I changed the CI tests to bypass the OpenStack client tests for now since it requires the simulator to be running.  Jest doesn't provide nearly the flexibility as Mocha so not sure how well we could have the OpenStack client tests start and stop the simulator.

For now, I think we can just run them manually whenever we work on the client.

src/openstack-client/Keystone.js
```diff
@@ -0,0 +1,106 @@
...
+    const response = await axios.post(this.usersUrl, body, this.client.getAuthHeaders())
+    return response.data.user
+  }
+
+  async deleteUser (userId) {
+    try {
+      await axios.delete(`${this.usersUrl}/${userId}`, this.client.getAuthHeaders())
+      return userId
+    } catch (err) {
+      throw new Error(`Unable to delete non-existant user`)
```
kevinylee1991: non-existent

src/openstack-client/OpenstackClient.js
```diff
@@ -0,0 +1,41 @@
...
+  serialize () {
+    return {
+      keystoneEndpoint: this.options.keystoneEndpoint,
+      unscopedToken: this.unscopedToken,
+      scopedToken: this.scopedToken,
+      catalog: this.catalog,
+    }
+  }
+
+  static hydrate (state) {
```
kevinylee1991: Could you add a comment explaining what this function is used for? I didn't know what hydrate meant so I wasn't sure the intent of the function at first

src/openstack-client/OpenstackClient.js
```diff
@@ -0,0 +1,41 @@
...
+    const options = {
+      keystoneEndpoint: state.keystoneEndpoint
+    }
+    const client = new OpenstackClient(options)
+    client.catalog = state.catalog
+    return client
+  }
+
+  getAuthHeaders (scoped = true) {
```
kevinylee1991: eventually we will need to add the option to add new headers

src/openstack-client/OpenstackClient.js
kevinylee1991: keystoneEndpoint will be hardcoded i believe? We need the keystone endpoint before we get the catalog (we need to be logged in first to get that), so does this function need to be split?

PR #87, opened 2018-06-06
MaterialUI_Logon

HF-liu:
Update logon page with material_UI.


src/app/static/Components/Layouts/Logon.js
```diff
@@ -0,0 +1,229 @@
+import React, { Component, Fragment } from 'react'
```
brennancheung: We should be modifying the existing Login form not creating a new one.

src/app_old/App.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react'
```
brennancheung: `app_old` should not exist at all.

PR #91, opened 2018-06-06
Styled Login Form

HF-liu:
Add new style to current login form.

screen shot 2018-06-06 at 7 59 45 pm

### DIFF
--- a/src/app/plugins/openstack/components/LoginPage.js
+++ b/src/app/plugins/openstack/components/LoginPage.js
@@ -1,8 +1,9 @@
-import React from 'react'
+import React, { Fragment } from 'react'
 import { connect } from 'react-redux'
-import Button from 'material-ui/Button'
 import Session from '../actions/session'
 import { withRouter } from 'react-router'
+import { withStyles } from '@material-ui/core/styles'
+import { Paper, Typography, Grid, TextField, Button, FormControlLabel, Checkbox } from '@material-ui/core'
 
 function mapStateToProps (state, ownProps) {
   const { login } = state.openstack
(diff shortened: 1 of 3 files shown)


src/stories/UI/loginPage.js
```diff
@@ -0,0 +1,148 @@
+import React, { Fragment } from 'react'
+import { withStyles } from '@material-ui/core/styles'
+import { Paper, Typography, Grid, TextField, Button, FormControlLabel, Checkbox } from '@material-ui/core'
+
+const styles = theme => ({
+  root: {
+    marginTop: '5%',
```
brennancheung: Check to see what is provided inside the `theme` object.  It probably has settings for `margin` and `padding`.

The vertical rhythm looks a bit off.

Also, the padding values don't have any units.

brennancheung:
I just realized this is completely separate.  We actually want to use this.  Please replace or hook it into https://github.com/platform9/pf9-ui-plugin/blob/master/src/app/plugins/openstack/components/LoginPage.js.

Make sure to keep the existing functionality so the app works.

src/app/plugins/openstack/components/LoginPage.js
```diff
@@ -46,32 +90,96 @@ export class LoginPage extends React.Component {
...
+            </Paper>
+          </Grid>
+        </Grid>
       </div>
     )
   }
 }
 
-export default withRouter(connect(mapStateToProps)(LoginPage))
+export default withRouter(connect(mapStateToProps)(withStyles(styles)(LoginPage)))
```
brennancheung: Might be able to use `compose` here to clean this line up.

You can see an example here: https://github.com/platform9/pf9-ui-plugin/blob/master/src/app/plugins/openstack/components/tenants/TenantsListPage.js#L22

PR #92, opened 2018-06-08
Added user update

kevinylee1991:
First pass at user update


src/server/helpers.js
```diff
@@ -5,6 +5,15 @@ export const notImplementedYet = (req, res) => res.status(500).send('Not impleme
 // present.
 export const findById = arr => id => (typeof arr === 'function' ? arr() : arr).find(x => x.id === id)
 
+export const updateById = (arr) => {
+  return (id, data) => {
+    let _arr = typeof arr === 'function' ? arr() : arr
```
brennancheung: We might want to make a utility function for this.  It looks like something we will use all over the place.

PR #101, opened 2018-06-13
Add UpdateVolumePage

HF-liu:
* Update `Volume` schema

* Add UpdateVolumePage


src/graphql/openstack/schemas/Volume.graphql
```diff
@@ -38,8 +39,14 @@ input VolumeInput {
   readonly: Boolean
 }
 
+input UpdateInput {
+  name: String!
```
brennancheung: Do these really need to be required?

PR #102, opened 2018-06-13
Fix small issues in ListTable

HF-liu:
* If all items on the current page are deleted, it will go back to the pervious page.
* `SelectAll` will get all the items on the current page.
* Ensure the height of L`ListTable` won't change frequently.
  * Merge `edit` and `delete` button to toolbar.
  * Always show `PaginationControl` to fix the height.
* Change the color to `primary`

screen shot 2018-06-14 at 9 48 42 am

### DIFF
--- a/src/app/core/common/ListTable.js
+++ b/src/app/core/common/ListTable.js
@@ -166,7 +166,7 @@ class ListTable extends React.Component {
       <TableRow hover key={row.id} {...checkboxProps}>
         {showCheckboxes &&
           <TableCell padding="checkbox">
-            <Checkbox checked={isSelected} />
+            <Checkbox checked={isSelected} color="primary" />
           </TableCell>
         }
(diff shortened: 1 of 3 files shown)


src/app/core/common/ListTable.js
```diff
@@ -18,6 +18,9 @@ const styles = theme => ({
   root: {
     width: '100%',
     marginTop: theme.spacing.unit * 3,
+    maxWidth: '94%',
```
brennancheung: This works but I'm wondering if there is anything from within the theme we can use.  I'd like to stay away from hard-coded values if there is alternative that works within the theme.

HF-liu: I wrap it up with `Grid` to avoid using hard-coded values. Though margins of 11/12 might be a little wide, it looks fine.

PR #107, opened 2018-06-19
Rough POC for form HOC

brennancheung:
This is still early.  I'm planning on adding a bit more to this.

The idea is that we have a HOC form wrapper that handles maintaining the state and performing validation.  The render prop function would render the contents of the form.  Utilities are provided to make things as convenient as possible but still let the user ultimately have complete control over custom rendering.

This also adds support for Enzyme which will allow us to test React components in unit tests.


brennancheung:
Ok, this new version is much cleaner.  All the logic and plumbing have been pushed into the form component and the individual controls.  No more worrying about plumbing values to and from.  Validations is just a placeholder for now.  We can add them as we need them.

src/app/core/common/ValidatedForm.js
```diff
@@ -0,0 +1,114 @@
...
+
+  state = {
+    value: this.props.initialValue || {},
+    fields: {}, // child fields inject data here
+    setField: this.setField,
+    defineField: this.defineField,
+  }
+
+  validateField = (field, value, validations, spec) => {
+    console.log(field, value, validations)
```
kevinylee1991: console log

src/app/core/common/__tests__/ValidatedForm.test.js
```diff
@@ -0,0 +1,10 @@
+// import React from 'react'
```
kevinylee1991: not sure if incomplete test or saving for later

src/app/core/common/Checkbox.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { FormControlLabel, Checkbox as BaseCheckbox } from '@material-ui/core'
+import { withFormContext } from 'core/common/ValidatedForm'
+import { pickMultiple, filterFields } from 'core/fp'
+
+class Checkbox extends React.Component {
+  constructor (props) {
+    super(props)
+    const spec = pickMultiple('validations')(props)
```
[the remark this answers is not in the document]
brennancheung: Not quite.  That would set a variable called `validations`.  This returns an object that has `validations` as one of the keys.  I previously had `required` as a sibling in the `spec` but decided that it should be inside of `validations` since it technically is a validation.

src/app/core/common/__tests__/ValidatedForm.test.js
brennancheung: I'm meaning to add to it later.  `jest` was complaining about the empty file when I commented everything out so I have a blank test.

PR #108, opened 2018-06-19
REST APIs for Volume.

HF-liu:
* Add get/getAll/create/update APIs for Volume.

* Add test cases for APIs.


src/openstack-client/__tests__/Volume.test.js
```diff
@@ -0,0 +1,62 @@
...
+    expect(volume.id).toBeDefined()
+
+    const updatedVolume = await client.volume.updateVolume(volume.id, {
+      name: 'NewName'
+    })
+    expect(updatedVolume.name).toBe('NewName')
+
+    // Wait for new volume's status changing to 'available'
+    var wait = ms => new Promise((resolve, reject) => setTimeout(resolve, ms))
+    await wait(7000)
```
brennancheung: Hmm, we should have some kind of loop that polls periodically.  In the simulator this is going to return instantly so no point in slowing tests down.

See what other people are doing, we can even create a utility function that accepts a callback and a `delay` and `maxRetries`.

brennancheung: Perhaps it can take another function that just returns a boolean about whether it should try again.

src/openstack-client/Volume.js
```diff
@@ -0,0 +1,99 @@
...
+      return response.data.volume
+    } catch (err) {
+      console.log(err)
+    }
+  }
+
+  async waitForCreate (delay, maxRetries, id) {
+    const url = `${await this.volumesUrl()}/${id}`
+    let response = await axios.get(url, this.client.getAuthHeaders())
+    let flag = (response.data.volume.status === 'available')
```
brennancheung: The initialization can be moved into the first param of the `for` loop.  `for (let flag = ..., let retries = 0; ...`

brennancheung: Scratch that, too much going on in the `for` loop.  After experimenting with it myself `while` is a little cleaner.

src/openstack-client/__tests__/Volume.test.js
```diff
@@ -25,13 +27,17 @@ describe('Volumes', () => {
     const newVolume = await client.volume.getVolume(volume.id)
     expect(newVolume).toBeDefined()
 
-    // Wait for new volume's status changing to 'available'
-    await client.volume.waitForCreate(1000, 20, newVolume.id)
-    await client.volume.deleteVolume(newVolume.id)
-
-    await client.volume.waitForDelete(1000, 20, newVolume.id)
+    waitUntil(client.volume.checkForCreate, newVolume.id, 1000, 20)
```
brennancheung: `waitUntil` returns a `Promise` so don't forget to `await`.

src/openstack-client/helpers.js
```diff
@@ -33,3 +33,17 @@ export const makeRegionedClient = async () => {
   await client.keystone.getServicesForActiveRegion()
   return client
 }
+
+export const waitUntil = async (condition, params, delay, maxRetries) =>
```
brennancheung: Don't pass in params and keep this as a single object.  You can use "currying" / "partial application" to encapsulate state.

```
const conditionFn = params => () => { /* do something with params */ }
await waitUntil({ condition: conditionFn(params), delay: 2000, maxRetries: 20 })
```

src/openstack-client/Volume.js
```diff
@@ -55,34 +55,37 @@ class Volume {
...
+    //     flag = (response.data.volume.status === 'available')
+    //   }
+    //   if (retries > maxRetries) {
+    //     throw new Error('Creation did not finish within set time.')
+    //   }
+    // }
     return flag
   }
 
   async waitForDelete (delay, maxRetries, id) {
```
brennancheung: Best to keep this in the test, not in the client.

src/openstack-client/Volume.js
```diff
@@ -0,0 +1,60 @@
...
+
+  async getVolume (id) {
+    const url = `${await this.volumesUrl()}/${id}`
+    const response = await axios.get(url, this.client.getAuthHeaders())
+    return response.data.volume
+  }
+
+  // Get volumes with details
+  async getVolumes () {
+    const url = await `${await this.volumesUrl()}/detail`
```
brennancheung: Don't need 2 `await`s here.  String interpolation is not async.

src/openstack-client/helpers.js
```diff
@@ -33,3 +34,38 @@ export const makeRegionedClient = async () => {
...
+      done = await condition()
+    }
+    done ? resolve() : reject(new Error('Task not done within time.'))
+  })
+
+export const sleep = (delay) =>
+  new Promise(resolve => setTimeout(resolve, delay))
+
+// TODO: Make this functions more generic
+export const waitForCreate = params => async () => {
```
brennancheung: This function is specific to volumes.  Rename to `waitForVolumeCreate` and move this function into the test file itself.

src/openstack-client/helpers.js
```diff
@@ -33,3 +34,38 @@ export const makeRegionedClient = async () => {
...
+  const client = await makeRegionedClient()
+  const services = await client.keystone.getServicesForActiveRegion()
+  const url = `${services.cinderv3.admin.url}/volumes/${params}`
+  let response = await axios.get(url, client.getAuthHeaders())
+  let flag = (response.data.volume.status === 'available')
+  return flag
+}
+// TODO: Make this functions more generic
+export const waitForDelete = params => async () => {
```
brennancheung: Same as `waitForCreate`

PR #110, opened 2018-06-20
Add Form Wrapper

HF-liu:


src/app/plugins/openstack/components/volumes/AddVolumePage.js
```diff
@@ -28,10 +29,9 @@ class AddVolumePage extends React.Component {
 
   render () {
     return (
-      <div>
-        <h1>Add volume</h1>
+      <FormWrapper title={'Add Volume'}>
```
brennancheung: `title="Add Volume"` is simpler.

src/app/core/common/FormWrapper.js
```diff
@@ -0,0 +1,66 @@
...
+                </Typography>
+              </Grid>
+              <Grid>
+                <Button
+                  variant="outlined"
+                  component={Link}
+                  to={backUrl}
+                  style={{ marginBottom: '1em' }}
+                >
+                  &lt;&lt;&nbsp;Back to list
```
brennancheung: Let's remove the double `&lt;` and replace it with something that is more directly suited for that.  I think `&larr` makes sense.  There's a good list here: https://www.w3schools.com/charsets/ref_html_entities_l.asp

src/app/plugins/openstack/components/glanceimages/AddGlanceImagePage.js
```diff
@@ -134,24 +136,32 @@ class AddGlanceImagePage extends React.Component {
...
-            </Button>
-            <Typography variant="display1" color="primary">Import a New Image</Typography>
-            <Divider />
-            {this.renderManualImport()}
-            {this.renderGlanceClient()}
-            <Divider />
-            {this.renderCreateNewImage()}
-          </Paper>
-        </Fragment>
+        <Grid container justify="center">
```
brennancheung: Hmm, does the form wrapper not work here?  We should see about cutting down on the code duplication.

PR #117, opened 2018-06-23
All REST APIs for Volume.

HF-liu:


src/openstack-client/Volume.js
```diff
@@ -13,17 +13,67 @@ class Volume {
 
   volumesUrl = async () => `${await this.endpoint()}/volumes`
 
+  async setRegionUrls () {
+    const services = (await this.client.keystone.getServiceCatalog()).find(x => x.name === 'cinderv3').endpoints
```
brennancheung: Is there a use case that you need to get this across multiple regions?  Take a look at `client.keystone.getServicesForActiveRegion`.

src/openstack-client/Volume.js
```diff
@@ -13,17 +13,67 @@ class Volume {
...
+    let baseUrlsByRegion = {}
+    for (let service of services) {
```
brennancheung: While this works, most modern code uses `Array#reduce` to accumulate the results.

src/openstack-client/Volume.js
```diff
@@ -13,17 +13,67 @@ class Volume {
...
+      const response = await axios.get(url, this.client.getAuthHeaders())
+      return response.data.volumes
+    } catch (err) {
+      console.log(err)
+    }
+  }
+
+  async getAllVolumesCount (limit, allTenants, markerId) {
+    let url = `${await this.volumesUrl()}/detail`
+    if (allTenants) {
```
brennancheung: This is a lot of branching.  `limit` is common to all so no need to specify it in every branch.  Also, in cases like this, it's cleaner to just store each part in variable and combine them:

```javascript
const cond1 = cond ? '?var=value' : ''
const url = cond1 + cond2 + cond3
```

src/openstack-client/Volume.js
```diff
@@ -55,6 +105,240 @@ class Volume {
...
+      console.log(err)
+    }
+  }
+
+  // TODO: test case for upload function (Image implement needed)
+  async uploadVolumeAsImage (id, image) {
+    const url = `${await this.volumesUrl()}/${id}/action`
+    try {
+      const response = await axios.post(url, { 'os-volume_upload_image': {
+        'container_format': 'bare',
```
brennancheung: Probably don't need the single quotes.

src/openstack-client/Volume.js
```diff
@@ -15,10 +15,10 @@ class Volume {
 
   async setRegionUrls () {
     const services = (await this.client.keystone.getServiceCatalog()).find(x => x.name === 'cinderv3').endpoints
-    let baseUrlsByRegion = {}
-    for (let service of services) {
-      baseUrlsByRegion[service.region] = service.url
-    }
+    const baseUrlsByRegion = services.reduce((res, service) => {
```
brennancheung: What's res?  For `reduce` the convention is to use `accum` or `memo` for that variable.

src/openstack-client/Volume.js
```diff
@@ -109,21 +100,24 @@ class Volume {
...
-  // TODO: test case for extend function (API is currently down)
+  // TODO: test case for extend function
   async extendVolume (id, size) {
     const url = `${await this.volumesUrl()}/${id}/action`
-    try {
-      const response = await axios.post(url, { 'os-extend': { 'new-size': size } }, this.client.getAuthHeaders())
-      return response.data.volume
-    } catch (err) {
-      console.log(err)
+    if (this.client.activeRegion.startsWith('AWS' || 'aws')) {
```
brennancheung: Hmm, let's move this into a function.  Also, I'm wondering if this is a good enough check and if it won't conflict with anything else.  I think it is fine for now but if you can think of something to make it more specific that might save from some false positives in the future.  Not sure if this is going to be an issue though.

brennancheung: As discussed, let's test for functionality, not vendor.  We can move the vendor specific checks into the functionality tests if that is the only way to verify.  Not sure if there is a way to check but what happens if we add another vendor that doesn't support this functionality.  Is there a way to determine it from something else other than the name of the vendor?  What if AWS adds it to some AWS regions and not others?  This is common for certain of their functionality to only be offered in certain regions.

src/openstack-client/OpenstackClient.js
```diff
@@ -11,8 +11,8 @@ class OpenstackClient {
     }
     this.keystone = new Keystone(this)
     this.glance = new Glance(this)
+    this.cinder = new Cinder(this)
     this.volume = new Volume(this)
```
brennancheung: If I'm reading this correctly, looks like Volume has been renamed so this should be removed?

PR #118, opened 2018-06-27
Add Neutron APIs

HF-liu:


kevinylee1991:
I think this lgtm for the most part -- we should implement the name prefix on all the resources we create in these tests and create a script to periodically delete the ones that remain from failed tests (if that is the route we are going for as discussed in the meeting).

We will probably also need to include some type of random uuid as a suffix so that we can distinguish objects created from the running test from previous ones.

src/openstack-client/helpers.js
```diff
@@ -29,7 +29,7 @@ export const makeScopedClient = async () => {
 export const makeRegionedClient = async () => {
   const client = await makeScopedClient()
   const regions = await client.keystone.getRegions()
-  client.setActiveRegion(regions[0].id)
+  client.setActiveRegion(regions[5].id)
```
brennancheung: What's this `5` all about?  This might change in the future?  Perhaps we should use `Array#find` to select the region by name.

HF-liu: This `5` switched region to `KVM-Neutron`. `Array.find` is a better idea.

PR #120, opened 2018-06-28
Persistent navbar

kevinylee1991:
Looking for suggestions regarding the showNavMenu and showNavBar stuff (left commented code in the review for easy reference) -- are these still needed? Which views with showNavBar be false for, maybe login?

### DIFF
--- a/src/app/App.js
+++ b/src/app/App.js
@@ -19,25 +19,19 @@ class App extends React.Component {
   render () {
     const theme = createMuiTheme({
       palette: {
-        type: this.props.theme
+        type: this.props.theme,
+        primary: {
+          light: '#aee0ff',
+          main: '#4aa3df',
+          dark: '#1e699c',
+          contrastText: '#fff',
+        }
       }
     })
 
     const { pluginManager } = this.context
     const options = pluginManager.getOptions()
-    const { showNavMenu, showSidebar, showFooter } = options
-
-    const renderNavMenu = () => (
-      <div id="_nav-menu">
-        <Navbar links={pluginManager.getNavItems()} />
-      </div>
-    )
-
-    const renderSidebar = () => (
-      <div id="_main-sidebar">
-        TODO: Sidebar
-      </div>
-    )
+    const { showFooter } = options
 
     const renderFooter = () => (
       <div id="_main-footer">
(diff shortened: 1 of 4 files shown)


src/app/core/common/Navbar.js
```diff
@@ -98,22 +132,20 @@ class Navbar extends React.Component {
...
-            <IconButton onClick={this.handleDrawerClose}>
-              <ChevronLeftIcon />
-            </IconButton>
-          </div>
-          <Divider />
-          <MenuList>
-            {links.map(this.renderNavLink)}
-          </MenuList>
+        <div className={classes.drawerHeader}>
+          <IconButton onClick={this.handleDrawerClose}>
```
HF-liu: Maybe we could use one function to replace both `handleDrawerClose` and `handleDrawerOpen`? Like:

```
handleDrawerStatus = () => {
    this.setState(prevState => ({
      open: !prevState.open
    }))
  }
```

src/app/App.js
```diff
@@ -25,33 +25,57 @@ class App extends React.Component {
...
+    //         </div>
+    //         {showFooter && renderFooter()}
+    //       </div>
+    //     </MuiThemeProvider>
+    //   </Router>
+    // )
+
     return (
       <Router>
         <MuiThemeProvider theme={theme}>
```
HF-liu: Picked theme color from Platform9.com:

```
const theme = createMuiTheme({
      palette: {
        type: this.props.theme,
        primary: {
          light: '#66CCF1',
          main: '#00ABE8',
          dark: '#007AB6',
          contrastText: '#fff',
        },
      }
    })
```

src/app/App.js
```diff
@@ -25,33 +25,57 @@ class App extends React.Component {
...
           <div id="_main-container">
-            {showNavMenu && renderNavMenu()}
-            {showSidebar && renderSidebar()}
-            <div id="_main-content">
+            <Navbar links={pluginManager.getNavItems()} >
```
brennancheung: weird space at end, probably going to trip up linting

kevinylee1991:
I tweaked the color hashes just a bit using the ui styleguide as a reference

PR #123, opened 2018-06-28
Add handleRemove to CRUDListContainer.

HF-liu:


src/app/core/common/CRUDListContainer.js
```diff
@@ -33,11 +33,23 @@ class CRUDListContainer extends React.Component {
...
     this.resolveDelete()
   }
 
+  handleRemove = async (id, item, removeQuery, getQuery) => {
+    this.props.client.mutate({
+      mutation: removeQuery,
+      variables: { id },
+      update: cache => {
+        const data = cache.readQuery({ query: getQuery })
+        data[`${item}`] = data[`${item}`].filter(x => x.id !== id)
```
brennancheung: Is `item` a string or object here?  Might not need the string interpolation.

src/app/core/common/CRUDListContainer.js
```diff
@@ -33,11 +33,23 @@ class CRUDListContainer extends React.Component {
   handleDeleteConfirm = () => {
     this.setState({ showConfirmation: false })
     const items = this.state.selectedItems || []
-    items.forEach(item => this.props.onRemove(item.id))
+    items.forEach(item => this.handleRemove(item.id, this.props.str, this.props.removeQuery, this.props.getQuery))
```
brennancheung: Pull out the props with `const { str, removeQuery, getQuery } = this.props` so we don't see `this.props` repeated so many times.

src/app/core/common/CRUDListContainer.js
```diff
@@ -33,11 +33,23 @@ class CRUDListContainer extends React.Component {
...
     this.setState({ selectedItems: [] })
     this.resolveDelete()
   }
 
+  handleRemove = async (id, str, removeQuery, getQuery) => {
+    this.props.client.mutate({
```
brennancheung: Related to the other PR.  I'm working if we can just import the `client` directly instead of passing it down.

src/app/plugins/openstack/components/flavors/FlavorsListContainer.js
```diff
@@ -8,24 +8,14 @@ import FlavorsList from './FlavorsList'
...
-      }
-    })
-  }
-
   render () {
     return (
       <CRUDListContainer
         items={this.props.flavors}
-        onRemove={this.handleRemove}
+        str="flavors"
```
brennancheung: `str` is not descriptive enough for a prop name.

PR #124, opened 2018-07-01
Add handleAdd and handleUpdate to ValidatedForm

HF-liu:


src/app/plugins/openstack/components/flavors/AddFlavorForm.js
```diff
@@ -12,17 +12,23 @@ const initialValue = {
   public: false,
 }
 
-const AddFlavorForm = ({ onSubmit }) =>
-  <ValidatedForm initialValue={initialValue} onSubmit={onSubmit}>
+const AddFlavorForm = ({ client, history }) =>
+  <ValidatedForm
```
brennancheung: Love it!  This is so clean.

src/app/core/common/ValidatedForm.js
```diff
@@ -61,13 +61,57 @@ class ValidatedForm extends React.Component {
...
+        }
+      })
+      history.push(backUrl)
+    } catch (err) {
+      console.error(err)
+    }
+  }
+
+  handleUpdate = inputObj => {
+    const { client, history, workId, updateQuery, backUrl } = this.props
```
brennancheung: Hmm, it's not obvious what `workId` means.  We might want to rename this.

Also, we should probably add `propTypes` to this class describing these fields.  If you use javadoc style docs on the proptypes they should show up nicely in Storybook with a description of the props.

src/app/plugins/openstack/components/flavors/AddFlavorForm.js
```diff
@@ -12,17 +12,23 @@ const initialValue = {
...
+    initialValue={initialValue}
+    client={client}
```
brennancheung: Hmm, it doesn't seem right that `client` and `history` need to be passed down.  It should probably pull these from React `context` somehow or the `ValidatedForm` should just import them.  `AddFlavorForm` should not need to know about `history` and `client` at all.

src/app/plugins/openstack/components/routers/AddRouterForm.js
```diff
@@ -1,60 +1,34 @@
...
+  <ValidatedForm
+    initialValue={initialValue}
+    client={client}
+    history={history}
+    backUrl="/ui/openstack/routers"
+    type="add"
+    addQuery={ADD_ROUTER}
+    getQuery={GET_ROUTERS}
+    str="routers"
+    cacheStr="createRouter"
```
brennancheung: Hmm, this component is getting too bloating with `props` that don't really have anything to do with the Form itself.  We might be able to wrap things more.  I've seen this before in other projects where people keep adding props to a Component until there are 25+ props.  We might want to use some kind of composition pattern so that we can wrap `<ValidatedForm />` with something that implements this logic.  A good example of this is `FormWrapper` with the `backUrl`.

brennancheung: Another alternative is making the graphql query parts a child component or a single component that can be passed in the props.  Not sure yet.

src/app/core/common/ValidatedForm.js
```diff
@@ -83,10 +130,28 @@ class ValidatedForm extends React.Component {
...
+  // GraphQl query to add an object
+  addQuery: PropTypes.object,
+  // GraphQl query to get an object
+  getQuery: PropTypes.object,
+  // GraphQl query to update an object
+  updateQuery: PropTypes.object,
+  // Action to take(add/delete/update)
+  action: PropTypes.string,
+  // Url to go back when the operation ends
+  backUrl: PropTypes.string.isRequired,
```
brennancheung: I see this prop on `FormWrapper` as well.  Wondering if we can make it so we don't have to specify `backUrl` twice.  Architecturally, `ValidatedForm` shouldn't have navigation logic in it.  We can hook into the `onSubmit` somehow and have that trigger the navigation in something that wraps it?  Not super sure yet.

PR #126, opened 2018-07-05
Add logo to navbar and change menu icon color.

HF-liu:
screen shot 2018-07-05 at 10 20 24 am

### DIFF
--- a/src/app/core/common/Navbar.js
+++ b/src/app/core/common/Navbar.js
@@ -11,8 +11,7 @@ import {
   ListItemText,
   MenuItem,
   MenuList,
-  Toolbar,
-  Typography
+  Toolbar
 } from '@material-ui/core'
 import ChevronLeftIcon from '@material-ui/icons/ChevronLeft'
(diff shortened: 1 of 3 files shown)


src/app/core/common/Navbar.js
```diff
@@ -160,16 +162,14 @@ class Navbar extends React.Component {
...
                 aria-label="open drawer"
                 onClick={this.handleDrawerOpen}
                 className={classNames(classes.menuButton, open && classes.hide)}
               >
                 <MenuIcon />
               </IconButton>
-              <Typography variant="title" color="inherit" noWrap>
-                Platform9
-              </Typography>
+              <img src="/images/logo.png" className={classes.logo} align="middle" />
```
brennancheung: Using the absolute path might break down the road.  Just a heads up, how we deploy the app and what URL it resides at might be subject to change.

It might end up at `http://FQDN/ui/v2` or any number of other places.

We might want to have the root path specified in the code as a variable.

`import { rootPath } from `core/globals`

Then use that in the `src`.

PR #127, opened 2018-07-05
Added form wizard step components

kevinylee1991:


src/app/core/common/BackButton.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import { Button } from '@material-ui/core'
+import { withFormContext } from 'core/common/ValidatedForm'
+
+class BackButton extends React.Component {
```
brennancheung: This can be a stateless functional component.

src/app/core/common/FormButtons.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+
+const formButtonsStyle = {
+  marginTop: '20px'
+}
+
+class FormButtons extends React.Component {
```
brennancheung: Same, use functional stateless component

src/app/core/common/FormButtons.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+
+const formButtonsStyle = {
```
brennancheung: Use the material design `withStyles` system.  Also, see if there is a setting in `theme` for padding.  David worked on this recently if you want to hit him up.  In general, we should use settings in the `theme` rather than hard-code numbers.

src/app/core/common/SubmitButton.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { Button } from '@material-ui/core'
+import { withFormContext } from 'core/common/ValidatedForm'
+
+class BackButton extends React.Component {
```
brennancheung: File is called `SubmitButton` but class is called `BackButton`.

These aren't really common components that will be used outside of the form wizard.  I think since these functional stateless components will end up being 1 or 2 lines they should be put in the same file as the wizard component.

src/app/core/common/BackButton.js
brennancheung: `const BackButton = ({ step, onClick }) => step > 1 && (<Button variant="outlined" onClick={handleBack}>Back</Button>)`

No need to return null, can use short-circuit operator.

src/app/core/common/BackButton.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import { Button } from '@material-ui/core'
+import { withFormContext } from 'core/common/ValidatedForm'
```
brennancheung: Not sure it needs the context.

src/app/core/common/ValidatedForm.js
```diff
@@ -46,11 +46,26 @@ class ValidatedForm extends React.Component {
     )
   }
 
+  handleBack = () => {
```
brennancheung: `handleBack` and `handleNext` do not belong here.  This has nothing to do with the form nor validation.  Components should do one thing only and do it well.

This functionality belongs with the Wizard component.

src/app/plugins/openstack/components/volumes/AddVolumeForm.js
```diff
@@ -21,21 +25,31 @@ const AddVolumeForm = () =>
...
+    </FormStep>
+    <FormStep stepNumber="3">
+      <TextField id="size" label="Capacity" type="number" />
+      <Checkbox id="bootable" label="Bootable" />
+      <Checkbox id="attachedMode" label="Attached Mode" />
+      <Checkbox id="readonly" label="Read only?" />
+      <TextField id="metadata" label="Metadata" />
+    </FormStep>
+    <FormButtons>
+      <BackButton />
```
brennancheung: Let `FormButtons` handle all this so we just need to include `<FormButtons />`.

src/app/plugins/openstack/components/volumes/AddVolumeForm.js
```diff
@@ -21,21 +25,31 @@ const AddVolumeForm = () =>
...
-    <TextField id="host" label="Host" />
-    <TextField id="instance" label="Instance" />
-    <TextField id="device" label="Device" />
-    <TextField id="size" label="Capacity" type="number" />
-    <Checkbox id="bootable" label="Bootable" />
-    <Checkbox id="attachedMode" label="Attached Mode" />
-    <Checkbox id="readonly" label="Read only?" />
-    <TextField id="metadata" label="Metadata" />
-    <Button type="submit" variant="raised">Add Volume</Button>
+    <FormStep stepNumber="1">
```
brennancheung: We need to have a `Wizard` component that handles the state, determines the number of steps (no need for that to be a parameter in `NextButton` and `SubmitButtons`).

brennancheung: Also, we don't need to number the step numbers.  The parent Wizard component can deal with it.

src/app/core/common/BackButton.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
```
brennancheung: This is specific to Wizard, not common to everything.  I would just make this an inline stateless functional component.  It doesn't need all the class overhead.

src/app/core/common/FormButtons.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
```
brennancheung: Same.  Make this a functional stateless component.

src/app/core/common/NextButton.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
```
brennancheung: functional stateless component

src/app/core/common/SubmitButton.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
```
brennancheung: functional stateless component

src/app/core/common/SubmitButton.js
```diff
@@ -0,0 +1,19 @@
...
+import { Button } from '@material-ui/core'
+
+class SubmitButton extends React.Component {
+  render () {
+    const { step, steps } = this.props
```
brennancheung: Also, since these passed in props are already in the parent class, I might just make it a `renderSubmitButton` in the parent `Wizard` class.

src/app/core/common/WizardStep.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { withWizardContext } from 'core/common/Wizard'
+
+class WizardStep extends React.Component {
+  render () {
+    const { step } = this.props
+    const stepNumber = parseInt(this.props.stepNumber)
+    return stepNumber === step ? (
```
brennancheung: This can probably be 1 line.

src/app/core/common/Wizard.js
```diff
@@ -0,0 +1,85 @@
...
+    step: 1
+  }
+
+  render () {
+    const numSteps = this.props.children.length
+
+    return (
+      <div>
+        <Provider value={this.state}>
+          <ValidatedForm {...this.props}>
```
brennancheung: Each step will need its own `ValidatedForm`; there may be custom validations for each step of the form to determine if it can advance to the next step.

This is too rigid and inflexible.  I fear it's going to be a bottleneck that technical debt will quickly accumulate because we can't build things with the more suitable abstractions.

Compound components are an advanced React topic but it will be worth it in the long run.  The FrontendMasters course is a good way to learn about it.  I will see about adding more to my video tutorial as well.

src/app/plugins/openstack/components/volumes/AddVolumeForm.js
```diff
@@ -21,21 +21,26 @@ const AddVolumeForm = () =>
...
-    <TextField id="instance" label="Instance" />
-    <TextField id="device" label="Device" />
-    <TextField id="size" label="Capacity" type="number" />
-    <Checkbox id="bootable" label="Bootable" />
-    <Checkbox id="attachedMode" label="Attached Mode" />
-    <Checkbox id="readonly" label="Read only?" />
-    <TextField id="metadata" label="Metadata" />
-    <Button type="submit" variant="raised">Add Volume</Button>
-  </ValidatedForm>
+    <WizardStep stepNumber="1">
```
brennancheung: This might not be needed once we move to a proper compound component, but it might still be, not certain.  In any case, if we want to reorder them, or insert a step in-between, having these be a number makes it more difficult to refactor.  I think we can just used named ids that are more semantic.  `step="general"`, `step="networking"` something like that.

src/app/core/common/FormButtons.js
brennancheung: Still needs to be converted to stateless functional component.

src/app/core/common/Wizard.js
```diff
@@ -0,0 +1,134 @@
...
+    )
+  }
+}
+
+Wizard.propTypes = {
+  onComplete: PropTypes.func,
+  context: PropTypes.object
+}
+
+export default compose(
```
brennancheung: Why doesn't Wizard need these?  If it does use them then we are probably doing something wrong.  We should not have a generic UI component coupled to the data layer or the routing layer.

src/app/core/common/WizardStep.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { withWizardContext } from 'core/common/Wizard'
+
+class WizardStep extends React.Component {
+  constructor (props) {
+    super(props)
+    props.addStep({stepId: props.stepId, validations: props.validations})
```
brennancheung: Let's have the `addStep` also send `props.children`.  That way `Wizard` has what it needs to selectively render the current `Step`.

src/app/plugins/openstack/components/volumes/AddVolumeForm.js
```diff
@@ -11,31 +11,43 @@ const initialValue = {
...
+const basicValidations = { name: 'required' }
+const advancedValidations = { tenant: 'required' }
+const configValidations = { size: 'required' }
+
+// See if the data supplied to WizardStep can be raised up to AddVolumeForm state
+const AddVolumeForm = ({ onComplete }) =>
+  <Wizard onComplete={onComplete} context={initialValue}>
+    {({ context, setContext }) => (
+      <div>
+        <WizardStep stepId="basic" validations={basicValidations}>
```
brennancheung: I don't like putting validations into the `Wizard`.  This is mixing concerns and making the code coupled.  Validations need to happen in `ValidatedForm`.  This is not the responsibility of `Wizard`.

src/app/core/common/WizardStep.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  render () {
+    const { stepId, showWizardStep } = this.props
+    return showWizardStep(stepId) ? (
```
brennancheung: We should get rid of `showWizardStep` altogether.  `Wizard` has the information it needs to make the determination of what the current step is, so it should also be responsible for rendering only the current step.

src/app/core/common/Wizard.js
kevinylee1991: oh yeah i can remove that completely

src/app/core/common/ValidatedForm.js
```diff
@@ -139,15 +144,21 @@ ValidatedForm.propTypes = {
...
   // Id of object to update
-  objId: PropTypes.string
+  objId: PropTypes.string,
+  // Initial values
+  initialValue: PropTypes.object,
+  // Set parent context
+  setContext: PropTypes.func,
 }
 
+export const PropKeys = Object.keys(ValidatedForm.propTypes)
```
brennancheung: What's this for?

src/app/plugins/openstack/components/volumes/AddVolumeForm.js
```diff
@@ -11,31 +11,39 @@ const initialValue = {
...
-    <TextField id="metadata" label="Metadata" />
-    <Button type="submit" variant="raised">Add Volume</Button>
-  </ValidatedForm>
+// See if the data supplied to WizardStep can be raised up to AddVolumeForm state
+const AddVolumeForm = ({ onComplete }) =>
+  <Wizard onComplete={onComplete} context={initialValue}>
+    {({ context, setContext }) => (
+      <div>
+        <WizardStep stepId="basic">
+          <ValidatedForm initialValue={context} setContext={setContext}>
```
brennancheung: What about making this more generic with `<ValidatedForm onSubmit={setContext}>`?

This also provides an API to pull out the GraphQL logic out of `ValidatedForm`.

PR #128, opened 2018-07-05
Add tabs to Volume page.

HF-liu:
screen shot 2018-07-05 at 2 42 03 pm

### DIFF
--- a/src/app/plugins/openstack/components/volumes/AddVolumeForm.js
+++ b/src/app/plugins/openstack/components/volumes/AddVolumeForm.js
@@ -14,7 +14,7 @@ const initialValue = {
 const AddVolumeForm = () =>
   <ValidatedForm
     initialValue={initialValue}
-    backUrl="/ui/openstack/volumes"
+    backUrl="/ui/openstack/storage#volumes"
     action="add"
     addQuery={ADD_VOLUME}
     getQuery={GET_VOLUMES}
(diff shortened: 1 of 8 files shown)


src/app/plugins/openstack/components/volumes/AddVolumeForm.js
```diff
@@ -14,7 +14,7 @@ const initialValue = {
 const AddVolumeForm = () =>
   <ValidatedForm
     initialValue={initialValue}
-    backUrl="/ui/openstack/volumes"
+    backUrl="/ui/openstack/storage#volumes"
```
brennancheung: Hmm, wasn't `backUrl` moved into `FormWrapper`?

src/app/plugins/openstack/components/volumes/StorageIndex.js
```diff
@@ -0,0 +1,59 @@
...
+              value={value}
+              onChange={this.handleChange}
+              indicatorColor="primary"
+              textColor="primary"
+            >
+              <Tab value="/ui/openstack/storage#volumes" label="Volumes" href="#volumes" />
+              <Tab value="/ui/openstack/storage#volumetypes" label="Volume Types" href="#volumetypes" />
+              <Tab value="/ui/openstack/storage#snapshots" label="Volume Snapsshots" href="#snapshots" />
+            </Tabs>
+            { value === '/ui/openstack/storage#volumes' && <TabContainer ><VolumePage /></TabContainer>}
```
brennancheung: Wow, I didn't realize Material-UI had such a poor API for tabs.  I imagined a much cleaner API but let's go with it for now.  We can refactor it in a different PR.

src/app/plugins/openstack/components/volumes/AddVolumeForm.js
HF-liu: `backUrl` in `FormWrapper` is for the back button, and `backUrl` here is for `handleUpdate` and `handleAdd` functions.

PR #130, opened 2018-07-06
Add Region & Tenant selectors (mockup).

HF-liu:
screen shot 2018-07-06 at 1 55 18 pm

### DIFF
--- a/src/app/core/common/Navbar.js
+++ b/src/app/core/common/Navbar.js
@@ -291,13 +291,11 @@ class Navbar extends React.Component {
                 <MenuIcon />
               </IconButton>
               <img src={logoPath} className={classes.logo} align="middle" />
-              {localStorage[LOCAL_STORAGE_NAMESPACE] &&
-                <div className={classes.rightTools}>
-                  {this.renderSelector('Region', [`AWS-US-West-1-Test`, `KVM-Neutron`])}
-                  {this.renderSelector('Tenant', [`Dev Team Tenant`, `Test Tenant`])}
-                  {this.renderAvatar()}
-                </div>
-              }
+              <div className={classes.rightTools}>
+                {this.renderSelector('Region', [`AWS-US-West-1-Test`, `KVM-Neutron`])}
+                {this.renderSelector('Tenant', [`Dev Team Tenant`, `Test Tenant`])}
+                {this.renderAvatar()}
+              </div>
             </Toolbar>
           </AppBar>
           {drawer}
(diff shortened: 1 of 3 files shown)


src/app/core/common/Navbar.js
```diff
@@ -145,6 +164,22 @@ class Navbar extends React.Component {
     this.setState({ anchorEl: null })
   }
 
+  handleTenantClick = event => {
```
brennancheung: There's some duplication going on here.  The anchor can be one of the parameters:

```javascript
handleClick = anchor => event => {
  this.setState({ [anchor]: event.currentTarget })
}
```

You can do the same for the `handleClose`.

src/app/core/common/Navbar.js
```diff
@@ -154,39 +189,85 @@ class Navbar extends React.Component {
     <MenuItem onClick={this.navTo(link.path)} key={link.path}><ListItemText primary={name} /></MenuItem>
   )
 
+  renderSelector = (name, list) => {
+    const { classes } = this.props
+    const selectorAnchor= name === 'Tenant'? this.state.tenantAnchor : this.state.regionAnchor
+    const selectorName = name+'-selector'
+    const clickFunc = name === 'Tenant'? this.handleTenantClick : this.handleRegionClick
```
brennancheung: `fn` is more common than `Func`.  But normally if it is an `onClick` handler it's either `onClick` or `handleClick`.  But you can just inline it down below now.

src/app/core/common/Navbar.js
```diff
@@ -154,39 +189,85 @@ class Navbar extends React.Component {
...
+    const closeFunc = name === 'Tenant'? this.handleTenantClose : this.handleRegionClose
+    return <div className={classes.selector}>
+      <Button
+        aria-owns={selectorAnchor ? selectorName : null}
+        aria-haspopup="true"
+        onClick={clickFunc}
```
brennancheung: `onClick={this.handleClick(name)}`

src/app/core/common/Navbar.js
```diff
@@ -226,7 +307,13 @@ class Navbar extends React.Component {
                 <MenuIcon />
               </IconButton>
               <img src={logoPath} className={classes.logo} align="middle" />
-              {localStorage[LOCAL_STORAGE_NAMESPACE] && avatar}
+              {localStorage[LOCAL_STORAGE_NAMESPACE] &&
```
brennancheung: Is this to detect if they are logged in?  The `localStorage` item still persists even if they are not logged in.  We leave behind things like last logged in tenant / region.  I added code changes that don't render the `Navbar` at all if they are not logged in.  This check might not be necessary depending on what you were intending.  The code is in PR right now.

PR #132, opened 2018-07-10
Add switch to control simulator / REST context

brennancheung:
This PR allows us to switch between using the simulator or REST API on a real DU.

Right now it is just a switch, the REST API context will be fleshed on out more in this PR.


brennancheung:
There's a lot more that needs to be added to this but since `context.js` has been moved and renamed merged conflicts are going to be quite an issue.  I want to get this merged now to avoid that and continue adding the REST API calls in future PRs.

src/server/middleware.js
```diff
@@ -1,4 +1,5 @@
-import Token from './models/Token'
+import context from './context'
+import config from '../../config'
 
 export const requestLogger = (req, res, next) => {
```
kevinylee1991: Maybe we should log to a log file as well?

src/server/middleware.js
```diff
@@ -7,14 +8,30 @@ export const requestLogger = (req, res, next) => {
...
   req.token = fullToken
   next()
 }
 
+// The OpenStack API client needs the X-Auth-Token from the request
+// and which region to use (from config.js) for the service catalog.
+export const injectClientInfo = (req, res, next) => {
+  const authToken = req.headers['x-auth-token']
+  if (context.client) {
+    context.client.unscopedToken = authToken
```
kevinylee1991: unscoped token and scoped token both set to the same token?

brennancheung: Yes, it's kind of a weird artifact of 2 incompatible systems.  The headers only allow 1 `X-Auth-Token` and there is no way for the middleware to know which one it is.

PR #135, opened 2018-07-16
Add Tenant/Role Selector

HF-liu:


src/app/core/common/TenantRoleSelectorContainer.js
```diff
@@ -10,7 +10,7 @@ import { pickMultiple } from 'core/fp'
 import { withFormContext } from 'core/common/ValidatedForm'
 import TenantRoleSelector from 'core/common/TenantRoleSelector'
 
-class TenantRoleContainer extends React.Component {
+class TenantRoleSelectorContainer extends React.Component {
```
brennancheung: I think we can safely shorten this to `TenantRolesContainer`.  The `Selector` part makes it too long and isn't really needed.

PR #138, opened 2018-07-23
Save pagination limit in localStorage

HF-liu:


src/app/core/common/ListTable.js
```diff
@@ -36,7 +39,7 @@ class ListTable extends React.Component {
       order: 'asc',
       orderBy: this.props.columns[0].id,
       page: 0,
-      rowsPerPage: 5,
+      rowsPerPage: session.getUserPreference()['perPage'] || 10,
```
brennancheung: This will have an error when `getUserPreference()` returns `undefined`.  This can be fixed in the implementation of the function though.

src/app/plugins/openstack/actions/__tests__/session.test.js
```diff
@@ -79,6 +79,9 @@ describe('signIn', () => {
     }
   })
 
+  // Ignore console.log messages
+  console.log = jest.fn()
```
brennancheung: Make sure to put it back after the test is done.  We don't want to get rid of `console.log` for other tests.  It is very useful for development and debugging to have them.

src/app/plugins/openstack/actions/session.js
```diff
@@ -54,6 +54,25 @@ const Session = (keystone = Keystone, mocks = {}) => {
   const getLastRegion = username => getStorage(regionStorageKey(username))
   const setLastRegion = (username, tenant) => ctx.setStorage(regionStorageKey(username), tenant)
 
+  const userPreferenceKey = username => `user-preference-${username}`
+  const getUserPreference = () => {
+    const username = getStorage('username')
+    if (!username) {
+      console.log('No user found.')
+      return
```
brennancheung: `return {}` so that `getUserPreference().xyz` throw an error.

src/app/plugins/openstack/actions/session.js
```diff
@@ -54,6 +54,25 @@ const Session = (keystone = Keystone, mocks = {}) => {
...
+    }
+    return getStorage(userPreferenceKey(username)) || {}
+  }
+  const initUserPreference = (username, key, value) => ctx.setStorage(userPreferenceKey(username), { [key]: value })
+  const setUserPreference = (key, value) => {
+    const username = getStorage('username')
+    if (!username) {
+      console.log('No user found.')
+      return
```
brennancheung: `return {}`

src/app/plugins/openstack/actions/session.js
```diff
@@ -79,6 +98,10 @@ const Session = (keystone = Keystone, mocks = {}) => {
     const region = ctx.getPreferredRegion(regions, lastRegion)
     setLastRegion(username, region)
 
+    const userPerference = ctx.getUserPreference(username)
+    const perPage = userPerference ? userPerference['perPage'] : 10
```
brennancheung: `userPreference.perPage` is a bit cleaner.

src/app/core/common/pf9-storage.js
```diff
@@ -1,11 +1,16 @@
...
-  const json = localStorage.getItem(LOCAL_STORAGE_NAMESPACE) || '{}'
-  return JSON.parse(json)
+  if (global.localStorage === undefined) {
+    return {}
+  }
+  return JSON.parse(localStorage.getItem(LOCAL_STORAGE_NAMESPACE) ||'{}')
 }
 
-export const getStorage = key => getInstance()[key]
+export const getStorage = key => {
```
brennancheung: This can be shortened to `export const getStore = key => getInstance[key]`

PR #139, opened 2018-07-24
Add more REST APIs to context

brennancheung:
Adds more REST APIs:
* Improved tenant support
* Improved users support
* Added Flavors CRUD

### DIFF
--- a/src/openstack-client/Keystone.js
+++ b/src/openstack-client/Keystone.js
@@ -57,6 +57,11 @@ class Keystone {
   get tokensUrl () { return `${this.v3}/auth/tokens?nocatalog` }
   get usersUrl () { return `${this.v3}/users` }
 
+  async getProject (id) {
+    const response = await axios.get(`${this.projectsUrl}/${id}`, this.client.getAuthHeaders())
+    return response.data.project
+  }
+
   async getProjects (scoped = false) {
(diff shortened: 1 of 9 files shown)


src/openstack-client/Keystone.js
```diff
@@ -57,6 +57,11 @@ class Keystone {
   get tokensUrl () { return `${this.v3}/auth/tokens?nocatalog` }
   get usersUrl () { return `${this.v3}/users` }
 
+  async getProject (id) {
+    const response = await axios.get(`${this.projectsUrl}/${id}`, this.client.getAuthHeaders())
+    return response.data.project
+  }
+
   async getProjects (scoped = false) {
```
kevinylee1991: Note: the UI currently uses a special platform9 API for getting tenants (gives us information in a batch rather than us having to call several APIs and aggregate to get the same data)

# this call to getTenants returns a pf9-specific version of #{adminUrl}/tenants, which includes user and MFA data
getTenants: -> BetterHttp.get  "#{adminUrl}/PF9-KSADM/all_tenants_all_users"

brennancheung: We may add that later.  Not sure, this may all become irrelevant if/when we switch to using SearchLight.  For now, I'm just going to go with the default OpenStack API.  We can make a change when we know which route we are going.

PR #141, opened 2018-07-27
Add Application schemas and visual parts.

HF-liu:
screen shot 2018-08-06 at 3 24 21 pm


src/app/plugins/openstack/components/applications/ApplicationCard.js
```diff
@@ -0,0 +1,121 @@
+import React from 'react'
+import { withStyles } from '@material-ui/core/styles'
+import {
+  Button,
+  Card,
+  CardContent,
```
brennancheung: Hmm, I see there is also a `CardActions` component in Material-UI.  Wondering if we should adopt it.

PR #142, opened 2018-07-31
Selector & SearchBar

HF-liu:
screen shot 2018-07-31 at 10 52 54 am

### DIFF
--- a/src/app/core/common/EnhancedTableToolbar.js
+++ b/src/app/core/common/EnhancedTableToolbar.js
@@ -41,7 +41,7 @@ const toolbarStyles = theme => ({
   }
 })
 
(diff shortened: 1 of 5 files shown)


src/app/core/common/EnhancedTableToolbar.js
```diff
@@ -41,7 +41,7 @@ const toolbarStyles = theme => ({
   }
 })
 
-const EnhancedTableToolbar = ({ classes, numSelected, title, onAdd, onDelete, onEdit, onSearch, onClear, searchTerm }) => (
+const EnhancedTableToolbar = ({ classes, numSelected, title, onAdd, onDelete, onEdit, onChange, searchTerm }) => (
```
brennancheung: Let's make this `onSearchChange` so it's more obvious to anyone using this component.

PR #143, opened 2018-08-01
Remove Kubernetes clusters since it does not work

brennancheung:
Just commenting out the navigation item for K8s since we aren't using it yet and it's not implemented correctly.


HF-liu:
LGTM.

PR #145, opened 2018-08-06
Show error alert on bad login attempts

brennancheung:
Not the prettiest but we can tweak it more if anyone has suggestions.  I created an `<Alert>` component that behaves similar to the `<Snackbar>`.  I originally tried using it but it's not possible to change the width of it to fit inline.

screen shot 2018-08-06 at 4 34 09 pm

### DIFF
--- a/src/app/core/common/Alert.js
+++ b/src/app/core/common/Alert.js
@@ -1,0 +1,83 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import classNames from 'classnames'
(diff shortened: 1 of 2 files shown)


brennancheung:
@kevinylee1991 `message` takes `PropTypes.node` so it should be able to accept pretty much anything that can be rendered in React.  If it turns out to be easier in the future we can probably have it accept `children` as a prop and render that.  Not sure what else needs this yet but it's only this one instance so we should be able to refactor it super easy.

PR #149, opened 2018-08-09
KeyValues component

brennancheung:


brennancheung:
I'm working on an `<Autocomplete>` component to work with this as well.  I will either add it here or start another PR depending on when this is approved.

brennancheung:
The autocomplete still needs some work on the formatting.  I think I need to look into the `anchorElement` for the `MenuList`.

screen shot 2018-08-09 at 5 44 06 pm

brennancheung:
Unfortunately the `<Selector>` component is a little too specific.  We might want to look into combining the 2 components or making a common base component for greater reuse.

src/app/core/common/KeyValues.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+  }
+
+  // Unfortunate overloading of naming going on here.
+  // field is one of ['key', 'value'], and value is the value of the field.
+  handleChange = field => event => {
+    const value = event.target.value
+    const { onChange } = this.props
+    this.setState(
+      () => ({ [field]: value }),
```
kevinylee1991: is [field] basically like \`${field}\`?

src/app/core/common/AutocompleteBase.js
```diff
@@ -0,0 +1,95 @@
...
+    const matched = this.matchedSuggestions()
+    const { value } = this.state
+    const { suggestions, onChange, initialValue, ...other } = this.props
+
+    return (
+      <div>
+        <TextField
+          value={value}
+          onChange={this.handleChange}
+          onBlur={this.handleClose}
```
kevinylee1991: design choice to ask madhura or designer about: should the suggestions be shown on focus or on click? Should there be some indicator that the particular field has autocomplete, like a down caret for example?

src/app/core/common/KeyValues.js
brennancheung: These are equivalent:

```
const field = 'foo'
const obj = { [field]: 'bar' }
```

is the same as:

`let obj = { foo: 'bar' }`

src/app/core/common/AutocompleteBase.js
```diff
@@ -0,0 +1,95 @@
...
+
+  handleClose = () => {
+    this.setState({ open: false })
+  }
+
+  renderSuggestions = suggestions => {
+    const { open } = this.state
+    if (!open || !suggestions || suggestions.length === 0) { return null }
+    return (
+      <Paper>
```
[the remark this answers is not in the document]
brennancheung: The normal behavior is that it just temporarily covers.  Nothing gets pushed down.  I think this is the standard behavior for most autocomplete textfields.

src/app/core/common/AutocompleteBase.js
brennancheung: Perhaps a down arrow on the far right.  I can add that.

PR #151, opened 2018-08-16
Initial placeholder for dashboard graphs

kevinylee1991:
There's a lot of work to be done here still, but want to leave for later. Let me know if there are any architectural issues etc. that you want me to deal with earlier than later.


src/app/core/common/dashboard_graphs/ServersGraph.js
```diff
@@ -0,0 +1,49 @@
...
+})
+
+@withStyles(styles)
+class ServersGraph extends React.Component {
+  state = {
+    hover: false
+  }
+
+  mouseEnterUpdate = () => {
+    this.setState((state) => {
```
brennancheung: Use shorthand here.  `this.setState(state => ({ hover: true }))`

PR #155, opened 2018-08-23
Simple animated arc for dashboard

kevinylee1991:


src/app/core/common/dashboard_graphs/SvgArc.js
```diff
@@ -15,17 +16,49 @@ const styles = theme => ({
...
+      points: '',
+    }
+
+    const endAngle = percentToAngle(props.percent) || 0
+    const anglePerTick = endAngle / 60
+    const animationTimeMs = props.animationTimeMs
+
+    this.drawArc = (angle, anglePerTick, animationTimeMs) => {
+      let points = describeArc(0, 0, 75, 0, this.state.currentAngle)
+      console.log(this.state.currentAngle)
```
brennancheung: console

src/app/core/common/dashboard_graphs/SvgArc.js
```diff
@@ -15,17 +16,49 @@ const styles = theme => ({
...
+      if (this.state.currentAngle === angle) {
+        this.setState(() => ({ points: points }))
```
brennancheung: Doesn't need to be the `function` version of `setState` if you aren't dependent on the previous state.  You can just `this.setState({ points })`

src/app/core/common/dashboard_graphs/SvgArc.js
```diff
@@ -15,17 +16,49 @@ const styles = theme => ({
...
+        return
+      }
 
-    // This is currently just a placeholder value for the hover
-    const points = this.props.hover ? describeArc(0, 0, 75, 0, 180) : describeArc(0, 0, 75, 0, angle)
+      this.setState((prevState, props) => {
+        const currentAngle = (prevState.currentAngle + anglePerTick) >= angle ? angle : (prevState.currentAngle + anglePerTick)
+        return { points: points, currentAngle: currentAngle }
```
brennancheung: Use shorthand: `return { points, currentAngle }`

src/app/core/common/dashboard_graphs/SvgArc.js
```diff
@@ -15,17 +16,49 @@ const styles = theme => ({
...
+      })
+
+      this.drawArcInterval = setTimeout(() => {
+        this.drawArc(angle, anglePerTick, animationTimeMs)
+      }, (animationTimeMs / 60))
```
brennancheung: Hmm, weird that the delays are variable.  It will either be jagged or updating more often than the browser repaints.  Use `requestAnimationFrame` instead.  You might need to adapt it to know the current elapsed time and use that to figure out how far into the animation it is.

You can also make a HOC to do something like `<AnimateValues ... />{values => <div>...</div>}</AnimateValues>`

PR #157, opened 2018-08-30
Start of decoupling GraphQL from CRUDListContainer

brennancheung:


src/openstack-client/Keystone.js
```diff
@@ -117,13 +117,17 @@ class Keystone {
...
     try {
       const response = await axios.post(this.tokensUrl, body)
-      const unscopedToken = response.headers['x-subject-token']
-      this.client.unscopedToken = unscopedToken
-      return unscopedToken
+      const newToken = response.headers['x-subject-token']
+      // We're not sure if this is a scoped or unscoped token but it should be
+      // fine to just set both of them here since scopedToken will be set later
+      // on if it needs to be.
+      this.client.unscopedToken = newToken
```
kevinylee1991: I think token here is the unscoped token (the token that gets passed into the function)

PR #158, opened 2018-08-31
Changed check-all box logic on lists, and added support for objects w…

kevinylee1991:
…ith no id property

I had to change a few fundamental things to allow objects with no id

### DIFF
--- a/src/app/core/common/CRUDListContainer.js
+++ b/src/app/core/common/CRUDListContainer.js
@@ -25,10 +25,8 @@ class CRUDListContainer extends React.Component {
     return `This will permanently delete the following: ${selectedNames}`
   }
 
-  handleDelete = selectedIds => {
-    this.setState({ showConfirmation: true })
-    const selectedItems = this.props.items.filter(item => selectedIds.includes(item.id))
-    this.setState({ selectedItems })
+  handleDelete = selected => {
+    this.setState({ showConfirmation: true, selectedItems: selected })
     // Stash the promise resolver so it can used to resolve later on in
     // response to user interaction (delete confirmation).
     return new Promise(resolve => { this.resolveDelete = resolve })
@@ -39,16 +37,18 @@ class CRUDListContainer extends React.Component {
   }
 
   handleDeleteConfirm = () => {
+    const { uniqueIdentifier } = this.props
     this.setState({ showConfirmation: false })
     const items = this.state.selectedItems || []
-    items.forEach(item => this.handleRemove(item.id))
+    items.forEach(item => this.handleRemove(item[uniqueIdentifier]))
+
     this.setState({ selectedItems: [] })
     // The user resolves the promise by clicking "confirm".
     this.resolveDelete()
   }
 
   handleRemove = async id => {
-    const { client, getQuery, objType, removeQuery, onRemove } = this.props
+    const { client, getQuery, objType, removeQuery, onRemove, uniqueIdentifier } = this.props
     if (getQuery) {
       // TODO: refactor any code that uses this path so we can decouple
       // GraphQL from CRUDListContainer
@@ -57,7 +57,7 @@ class CRUDListContainer extends React.Component {
         variables: { id },
         update: cache => {
           const data = cache.readQuery({ query: getQuery })
-          data[objType] = data[objType].filter(x => x.id !== id)
+          data[objType] = data[objType].filter(x => x[uniqueIdentifier] !== id)
           cache.writeQuery({ query: getQuery, data })
         }
       })
@@ -103,6 +103,17 @@ class CRUDListContainer extends React.Component {
 CRUDListContainer.propTypes = {
   addUrl: PropTypes.string,
   editUrl: PropTypes.string,
+  /*
+    Some objects have a unique identifier other than 'id'
+    For example sshKeys have unique identifier of 'name' and the APIs
+    rely on using the name as part of the URI. Specify the unique identifier
+    in props if it is different from 'id'
+  */
+  uniqueIdentifier: PropTypes.string,
+}
+
+CRUDListContainer.defaultProps = {
+  uniqueIdentifier: 'id'
 }
 
 export default compose(
--- a/src/app/core/common/ListTable.js
+++ b/src/app/core/common/ListTable.js
@@ -1,3 +1,5 @@
+/* eslint-disable react/no-did-update-set-state */
+
 import React from 'react'
 import PropTypes from 'prop-types'
 import { withStyles } from '@material-ui/core/styles'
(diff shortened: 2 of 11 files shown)


src/app/core/common/ListTable.js
```diff
@@ -35,17 +38,40 @@ const styles = theme => ({
 class ListTable extends React.Component {
   constructor (props) {
     super(props)
-    const { columns, getUserPreferences } = props
+    const { columns, getUserPreferences, data } = props
     const prefs = (getUserPreferences && getUserPreferences()) || {}
 
+    // Add id to properties if non-existent
+    // Some objects like SSH Keys do not have ids and have other unique identifier
+    const _data = data.map((item) => {
```
brennancheung: Use shorthand notation.

src/app/core/common/ListTable.js
```diff
@@ -35,17 +38,40 @@ const styles = theme => ({
...
+      searchTerm: '',
+      data: _data
+    }
+  }
+
+  // Need to set data again in state because when prop is updated,
+  // the constructor is not called again
+  componentDidUpdate (prevProps) {
+    if (prevProps.data !== this.props.data) {
+      const _data = this.props.data.map((item) => {
```
brennancheung: Use shorthand notation.

src/app/core/common/ListTable.js
```diff
@@ -35,17 +38,40 @@ const styles = theme => ({
...
-      selectedAll: false,
-      searchTerm: ''
+      searchTerm: '',
+      data: _data
+    }
+  }
+
+  // Need to set data again in state because when prop is updated,
+  // the constructor is not called again
+  componentDidUpdate (prevProps) {
```
brennancheung: Might want to use `getDerviedStateFromProps` instead.

https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

Also read https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html in case there is another pattern.  It's pretty rare that we need to use either of these component lifecycle methods and its a code smell that something wasn't architected correctly.

src/app/core/common/ListTable.js
```diff
@@ -70,26 +96,48 @@ class ListTable extends React.Component {
...
-    let endIdx = startIdx + rowsPerPage
-    const getId = x => x.id
+    const searchData = searchTerm === '' ? sortedData : this.filterBySearch(sortedData, searchTarget)
+    const paginatedData = paginate ? this.paginate(searchData) : searchData
+
+    let newSelected
+    if (checked) {
+      // Add active paginated rows that are not already selected
+      newSelected = [...selected]
+      for (let row of paginatedData) {
```
brennancheung: This looks like a good use case for `filter`.  Then you can "concat" it onto `newSelected`.

src/app/core/common/CRUDListContainer.js
```diff
@@ -19,9 +19,13 @@ class CRUDListContainer extends React.Component {
     return `This will permanently delete the following: ${selectedNames}`
   }
 
-  handleDelete = selectedIds => {
+  handleDelete = selected => {
+    const { uniqueIdentifier } = this.props
     this.setState({ showConfirmation: true })
-    const selectedItems = this.props.items.filter(item => selectedIds.includes(item.id))
+    const selectedItems = this.props.items.filter((item) => {
+      const selectedIds = selected.map(row => row[uniqueIdentifier])
```
brennancheung: This line is not dependent on `item` so it should be moved outside the `filter` function.

src/app/core/common/CRUDListContainer.js
```diff
@@ -19,9 +19,13 @@ class CRUDListContainer extends React.Component {
     return `This will permanently delete the following: ${selectedNames}`
   }
 
-  handleDelete = selectedIds => {
+  handleDelete = selected => {
+    const { uniqueIdentifier } = this.props
     this.setState({ showConfirmation: true })
-    const selectedItems = this.props.items.filter(item => selectedIds.includes(item.id))
+    const selectedItems = this.props.items.filter((item) => {
```
brennancheung: It's not obvious what is going on.  Add some comments.

src/app/core/common/CRUDListContainer.js
```diff
@@ -19,9 +19,13 @@ class CRUDListContainer extends React.Component {
...
+      const selectedIds = selected.map(row => row[uniqueIdentifier])
+      return selectedIds.indexOf(item[uniqueIdentifier]) > -1
```
brennancheung: Consider `Array#includes`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

src/app/core/common/ListTable.js
```diff
@@ -70,26 +96,48 @@ class ListTable extends React.Component {
...
+        if (selected.indexOf(row) === -1) {
+          newSelected.push(row)
+        }
+      }
+    } else {
+      // Remove active paginated rows from selected
+      newSelected = selected.filter((row) => {
+        return paginatedData.indexOf(row) === -1
```
brennancheung: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

src/app/core/common/ListTable.js
```diff
@@ -70,26 +96,48 @@ class ListTable extends React.Component {
...
+      // Add active paginated rows that are not already selected
+      newSelected = [...selected]
+      for (let row of paginatedData) {
+        if (selected.indexOf(row) === -1) {
+          newSelected.push(row)
+        }
+      }
+    } else {
+      // Remove active paginated rows from selected
+      newSelected = selected.filter((row) => {
```
brennancheung: `selected.filter(row => paginatedData.includes(row))`

src/app/core/common/CRUDListContainer.js
```diff
@@ -86,6 +91,11 @@ class CRUDListContainer extends React.Component {
 CRUDListContainer.propTypes = {
   addUrl: PropTypes.string,
   editUrl: PropTypes.string,
+  uniqueIdentifier: PropTypes.string,
```
brennancheung: Add comments describing the use case for this prop.

src/app/core/common/ListTable.js
```diff
@@ -70,26 +71,46 @@ class ListTable extends React.Component {
     return sorted
   }
 
+  areAllSelected = (data) => {
```
brennancheung: `areAllSelected = data => data.every(row => selected.includes(row))`

src/app/core/common/ListTable.js
brennancheung: Still needs to use `filter` and also `includes` instead of `indexOf`.

PR #159, opened 2018-09-11
Added initial hosts functionality. Added data models for nova hypervi…

kevinylee1991:
…sors and resmgr hosts, and UI view for hypervisors.

Initial hosts work. Would like to confirm some design choices before going forward regarding how host data should be shown to the user (like whether we are going to combine resmgr and nova hypervisor data the same way we did in the old UI, or if we are going for something different), and where data aggregation should take place.

### DIFF
--- a/src/app/plugins/openstack/components/hosts/HostsListContainer.js
+++ b/src/app/plugins/openstack/components/hosts/HostsListContainer.js
@@ -1,0 +1,33 @@
+/* eslint-disable camelcase */
+import React from 'react'
+import PropTypes from 'prop-types'
+import CRUDListContainer from 'core/common/CRUDListContainer'
+import HostsList from './HostsList'
+
+class HostsListContainer extends React.Component {
+  render () {
+    const hosts = this.props.hosts || []
+    return (
+      <CRUDListContainer
+        items={hosts}
+        addUrl="/ui/openstack/hosts/add"
+        editUrl="/ui/openstack/hosts/edit"
+      >
+        {({ onDelete, onAdd, onEdit }) => (
+          <HostsList
+            hosts={hosts}
+            onAdd={onAdd}
+            onDelete={onDelete}
+            onEdit={onEdit}
+          />
+        )}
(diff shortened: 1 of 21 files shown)


src/app/plugins/openstack/components/hosts/HostsListPage.js
```diff
@@ -0,0 +1,23 @@
...
+import { compose } from 'core/fp'
+import HostsListContainer from './HostsListContainer'
+import requiresAuthentication from '../../util/requiresAuthentication'
+import DataLoader from 'core/DataLoader'
+
+const loadHosts = async ({ setContext, context }) => {
+  // const hosts = await context.openstackClient.resmgr.getHosts()
+  // console.log(hosts)
+  const hosts = await context.openstackClient.nova.getHypervisors()
+  console.log(hosts)
```
brennancheung: Remove `console.log`

src/app/plugins/openstack/components/hosts/HostsListContainer.js
```diff
@@ -0,0 +1,34 @@
...
+import CRUDListContainer from 'core/common/CRUDListContainer'
+import HostsList from './HostsList'
+
+class HostsListContainer extends React.Component {
+  render () {
+    const hosts = this.props.hosts || []
+    return (
+      <CRUDListContainer
+        items={hosts}
+        objType="hosts"
```
brennancheung: `objType` probably not needed unless this requires GraphQL.

src/server/models/Catalog.js
```diff
@@ -10,23 +10,23 @@ const Catalog = {
           {
             id: uuid.v4(),
             interface: 'admin',
-            region: 'master',
-            region_id: 'master',
-            url: '/neutron'
+            region: 'kvm-neutron',
+            region_id: 'kvm-neutron',
+            url: 'http://localhost:4444/neutron'
```
brennancheung: Probably best to use the option in `config.js` for this instead of hard-coding.

PR #160, opened 2018-09-11
CRUD volumes refactoring

brennancheung:
* Making some Higher Order Components (HOCs) to make CRUD based operations easier outside of GraphQL.
* Creating `<Picklist>` functionality since we don't have anything that works well in forms.
* Added custom `render` method support for Table `columnDef`s.

Currently this PR still needs some work on the Picklist and Volumes CRUD functionality.

### DIFF
--- a/src/app/core/common/ListTable.js
+++ b/src/app/core/common/ListTable.js
@@ -33,7 +33,6 @@ const styles = theme => ({
 
 // TODO: this component should take an optional sort function in the columns prop (eg. for IP addresses)
 
-@withStyles(styles)
 class ListTable extends React.Component {
   constructor (props) {
     super(props)
@@ -103,12 +102,16 @@ class ListTable extends React.Component {
     let newSelected = []
 
     if (selectedIndex === -1) {
+      // not found
       newSelected = newSelected.concat(selected, row)
     } else if (selectedIndex === 0) {
+      // first
       newSelected = newSelected.concat(selected.slice(1))
     } else if (selectedIndex === selected.length - 1) {
+      // last
       newSelected = newSelected.concat(selected.slice(0, -1))
     } else if (selectedIndex > 0) {
+      // somewhere inbetween
       newSelected = newSelected.concat(
         selected.slice(0, selectedIndex),
         selected.slice(selectedIndex + 1),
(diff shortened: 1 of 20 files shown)


src/app/core/common/ListTable.js
```diff
@@ -178,10 +181,15 @@ class ListTable extends React.Component {
 
   renderCell = (columnDef, contents) => {
     const { cellProps = {} } = columnDef
+    let _contents = contents
+
+    if (typeof contents === 'boolean') { _contents = String(_contents) }
```
kevinylee1991: Trying to think of other content types -- array? Also we may need custom cells in the future, like one that shows a progress bar for example

src/app/plugins/openstack/components/volumes/AddVolumeForm.js
```diff
@@ -3,49 +3,127 @@ import Wizard from 'core/common/Wizard'
...
+    return (
+      <Wizard onComplete={onComplete} context={initialValue}>
+        {({ context, setContext, onNext, activeStepId }) => {
+          return (
+            <div>
+              <WizardStep stepId="source" label="Source" activeStepId={activeStepId}>
+                <Picklist name="sourceType" label="Volume Source" value={sourceType} onChange={this.setField('sourceType')} options={sourceTypes} />
+                {sourceType === 'Snapshot' &&
+                  <ValidatedForm initialValue={context} onSubmit={setContext} triggerSubmit={onNext}>
+                    <DataLoader dataKey="volumeSnapshots" loaderFn={nop}>
```
kevinylee1991: This is cool -- if we need to ever reload the list though, how can we re-trigger the loader function?

src/app/plugins/openstack/components/volumes/AddVolumeForm.js
```diff
@@ -3,49 +3,127 @@ import Wizard from 'core/common/Wizard'
...
+              </WizardStep>
+
+              <WizardStep stepId="config" label="Config" activeStepId={activeStepId}>
+                <ValidatedForm initialValue={context} onSubmit={setContext} triggerSubmit={onNext}>
+                  <Checkbox id="attachedMode" label="Attached Mode" />
+                  <Checkbox id="readonly" label="Read only?" />
+                  <TextField id="metadata" label="Metadata" />
+                </ValidatedForm>
+              </WizardStep>
+              <pre>{JSON.stringify(context, null, 4)}</pre>
```
kevinylee1991: is this for debugging?

src/app/plugins/openstack/components/volumes/AddVolumeForm.js
brennancheung: We can do the re-triggering somewhere outside this if we want.  We just need to update `context[dataKey]` and the views will automatically re-render themselves.  We could provide a `refresh` callback that the code can call if it wants to refresh as well.  It should be pretty flexible.  We can add it in later fairly easily.

src/app/core/common/ListTable.js
brennancheung: I made it so we can also pass in a render function in the `columnDef` so we can have a lot of flexibility.  We can make a library of these renderers and just import them when we need them.  Sorting is another issue though.  We need to make the rendered value be what is sorted.  Right now it is the original value.  But we can cross that bridge when we get there.

src/app/plugins/openstack/components/volumes/AddVolumeForm.js
brennancheung: Yes, that was there for debugging purposes.  I got rid of it now.

PR #161, opened 2018-09-13
Made SSH keys work with OpenStack client and adapted simulator to mat…

kevinylee1991:
…ch client

### DIFF
--- a/src/app/plugins/openstack/components/sshkeys/SshKeysListContainer.js
+++ b/src/app/plugins/openstack/components/sshkeys/SshKeysListContainer.js
@@ -2,22 +2,22 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import CRUDListContainer from 'core/common/CRUDListContainer'
 import SshKeysList from './SshKeysList'
-import { GET_SSH_KEYS, REMOVE_SSH_KEY } from './actions'
+import { addIdsToCollection } from 'util/helpers'
 
 class SshKeysListContainer extends React.Component {
   render () {
+    const sshKeys = this.props.sshKeys ? addIdsToCollection(this.props.sshKeys) : []
+
     return (
       <CRUDListContainer
-        items={this.props.sshKeys}
+        items={sshKeys}
         objType="sshKeys"
-        getQuery={GET_SSH_KEYS}
-        removeQuery={REMOVE_SSH_KEY}
         addUrl="/ui/openstack/sshkeys/add"
         uniqueIdentifier="name"
       >
         {({ onDelete, onAdd }) => (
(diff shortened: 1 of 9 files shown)


src/app/plugins/openstack/components/sshkeys/SshKeysListContainer.js
```diff
@@ -2,22 +2,22 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import CRUDListContainer from 'core/common/CRUDListContainer'
 import SshKeysList from './SshKeysList'
-import { GET_SSH_KEYS, REMOVE_SSH_KEY } from './actions'
+import { addIdsToCollection } from '../../../../util/helpers'
```
brennancheung: Add an alias in the `webpack.config.js` so we don't have such long relative paths.

src/app/plugins/openstack/components/sshkeys/SshKeysListPage.js
```diff
@@ -1,26 +1,20 @@
...
-import Loader from 'core/common/Loader'
+import { compose } from 'react-apollo'
 import SshKeysListContainer from './SshKeysListContainer'
 import requiresAuthentication from '../../util/requiresAuthentication'
-import { GET_SSH_KEYS } from './actions'
-import { addIdsToCollection } from '../../../../util/helpers'
+import DataLoader from 'core/DataLoader'
+
+const loadSshKeys = async ({ setContext, context }) => {
+  const _sshKeys = await context.openstackClient.nova.getSshKeys()
```
brennancheung: What about using `response` instead of `_sshKeys`.

src/app/plugins/openstack/components/sshkeys/SshKeysListPage.js
```diff
@@ -1,26 +1,20 @@
...
-import DisplayError from 'core/common/DisplayError'
-import Loader from 'core/common/Loader'
+import { compose } from 'react-apollo'
 import SshKeysListContainer from './SshKeysListContainer'
 import requiresAuthentication from '../../util/requiresAuthentication'
-import { GET_SSH_KEYS } from './actions'
-import { addIdsToCollection } from '../../../../util/helpers'
+import DataLoader from 'core/DataLoader'
+
+const loadSshKeys = async ({ setContext, context }) => {
```
brennancheung: I've been finding that these loading functions will probably be needed elsewhere.  For example, somewhere else in the UI we need to have the user choose from an SSH key from the existing list when creating an instance.  As a result, I've been creating an `actions.js` file in the same folder so it can be reused easier.

PR #162, opened 2018-09-14
Add/Delete SSH Keys

kevinylee1991:


src/app/plugins/openstack/components/sshkeys/AddSshKeyPage.js
```diff
@@ -1,16 +1,36 @@
...
+import { compose } from 'core/fp'
 import requiresAuthentication from '../../util/requiresAuthentication'
+import { withAppContext } from 'core/AppContext'
 
 class AddSshKeyPage extends React.Component {
+  handleAdd = async sshKey => {
+    const { setContext, context, history } = this.props
+    try {
+      const body = { keypair: sshKey }
+      const response = await context.openstackClient.nova.createSshKey(body)
```
brennancheung: Is it always `keyPair`?  Maybe the client should do this work instead of doing it outside the client.

src/app/plugins/openstack/components/sshkeys/SshKeysListContainer.js
```diff
@@ -13,7 +25,9 @@ class SshKeysListContainer extends React.Component {
         items={sshKeys}
         objType="sshKeys"
         addUrl="/ui/openstack/sshkeys/add"
-        uniqueIdentifier="name"
+        onRemove={this.handleRemove}
+        uniqueIdentifier={sshKeyUniqueIdentifier}
+        // onRemove=
```
brennancheung: Looks like this comment was accidentally left behind.

PR #163, opened 2018-09-15
Set scoped token on login

kevinylee1991:
Scoped token doesn't seem to be set upon logging in -- this change works but not sure if it's ideal.

To reproduce:

Start with fresh ui server and login -- go directly to SSH keys or Hosts page (these go through OpenStack client, which is where the issue exists). Should show unauthorized.

### DIFF
--- a/src/openstack-client/Keystone.js
+++ b/src/openstack-client/Keystone.js
@@ -124,7 +124,7 @@ class Keystone {
       const newToken = response.headers['x-subject-token']
       this.client.unscopedToken = token
       this.client.scopedToken = newToken
-      return newToken
+      return token
     } catch (err) {
       // authentication failed
       return null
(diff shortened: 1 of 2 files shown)


src/openstack-client/Keystone.js
```diff
@@ -124,7 +124,7 @@ class Keystone {
       const newToken = response.headers['x-subject-token']
       this.client.unscopedToken = token
```
brennancheung: This line used to read `this.client.unscopedToken = newToken`.

Not sure why it was changed in https://github.com/platform9/pf9-ui-plugin/commit/70e226f5b18d9045c93c77460b7ce48a6b85c35d#diff-9b4a749e0b23201963fc7ca75f9087d0R125

But I would test to see if that change didn't introduce this problem.

kevinylee1991: The main issue I think was because the scopedToken doesn't get set in the client in the authenticate function run during login, only when loading the app when already logged in. It's a bit confusing for me just going through the code, especially since renewToken does not include the project scope in the request, I think I need to take a closer look

PR #164, opened 2018-09-15
Fix HMR for AddVolumePage

brennancheung:
It turns out the plugin system is creating a level of indirection that Hot Module Reloading can't get through.  I tried to add some workaround but wasn't able to get it to work.

I ended up just adding the `Route` directly and now HMR is working.

It is worth exploring if we actually need this Plugin system and if we can't just hard-code the routes directly.  It would make the codebase simpler and I doubt we're actually going to use this codebase as a plugin architecture.

It might be best just to abandon it.

If we decide to remove the plugin system.  We just need to move the routes over and delete the plugin code.  Should be a simple change.


brennancheung:
I simplified the plugin system a little bit and removed the code that had the old style of using React's `context` system.  Turns out we don't need `context` at all for the plugin system.

I also determined that the plugin system is retaining an old reference to the Components and was not able to find a way to force HMR with the plugin system.

During development we can just hard-code the Route directly.  See https://github.com/platform9/pf9-ui-plugin/pull/164/commits/e3b47267fbf0f2b12d76af91314e2637a108ec72 for an example of how to do this.

Another option would be to use Storybook if we don't need the rest of the system.

For now, I've decided not to abandon the plugin system since we have a workaround to make development work with HMR.

PR #170, opened 2018-09-19
Switched flavors from graphql to openstack client

kevinylee1991:


src/app/plugins/openstack/components/flavors/AddFlavorPage.js
```diff
@@ -1,16 +1,35 @@
...
+import { withRouter } from 'react-router-dom'
+import { compose } from 'core/fp'
 import requiresAuthentication from '../../util/requiresAuthentication'
+import { withAppContext } from 'core/AppContext'
 
 class AddFlavorPage extends React.Component {
+  handleAdd = async flavor => {
+    const { setContext, context, history } = this.props
+    try {
+      const body = { flavor: flavor }
```
brennancheung: Shorthand: `const body = { flavor }`

PR #171, opened 2018-09-21
TenantChooser

brennancheung:
Still seems to be an issue when switching tenants that there are authorization errors.

### DIFF
--- a/src/app/plugins/openstack/components/tenants/TenantChooser.js
+++ b/src/app/plugins/openstack/components/tenants/TenantChooser.js
@@ -1,0 +1,77 @@
+import React from 'react'
+import Selector from 'core/common/Selector'
+import { compose } from 'core/fp'
+import { loadTenants } from './actions'
+import { withAppContext } from 'core/AppContext'
+
+class TenantChooser extends React.Component {
+  state = {
+    tenantSearch: '',
+    currentTenantName: '',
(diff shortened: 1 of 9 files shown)


brennancheung:
Authorization issues are now solved.  Turns out we just need to reload the service catalog.

Reloading the data for the current view when switching tenants was a bigger issue.  Turns out we can just use the DOM's `Event` architecture to simply solve this problem.  When switching tenants we dispatch a `scopeChanged` event on `window` and then `DataLoader` listens for these events and triggers a reload of the data.

src/app/plugins/openstack/components/tenants/TenantChooser.js
```diff
@@ -0,0 +1,93 @@
...
+    })
+  }
+
+  resetTenantScopedContext = (tenant) => {
+    const { setContext } = this.props
+    // Clear any data that should change when the user changes tenant.
+    // The data will then be reloaded when it is needed.
+    setContext({
+      volumes: undefined,
+      volumeSnapshots: undefined,
```
kevinylee1991: instances, flavors, sshKeys as well

brennancheung: Thanks for drawing attention to those.  We can add those in separate PRs as needed.

PR #175, opened 2018-09-25
Fixed DataUpdater to not break if collection is undefined

kevinylee1991:


src/app/core/DataUpdater.js
```diff
@@ -12,7 +12,7 @@ import { withAppContext } from 'core/AppContext'
  * Additionally, it handles loading as well through `DataLoader`.
  */
 class DataUpdater extends React.Component {
-  findById = (arr, id) => arr.find(x => x.id === id)
+  findById = (arr, id) => arr ? arr.find(x => x.id === id) : null
```
brennancheung: Use `(arr || []).find` so it will return `undefined`.  Shouldn't be `null`,

brennancheung:
We can delete this PR now that we have this fix in `master` already.

PR #183, opened 2018-10-03
Create volumes

brennancheung:
This adds the ability to create volumes.  Right not it only implements blank volumes.  Creating volumes from image, snapshot, and other volume can either be added to this PR or a separate one.

There is also a small fix the "App Loading..." screen.  It was preventing the login page from showing.

### DIFF
--- a/src/app/core/common/TextField.js
+++ b/src/app/core/common/TextField.js
@@ -18,9 +18,13 @@ class TextField extends React.Component {
   get restFields () { return filterFields(...withFormContext.propsToExclude)(this.props) }
 
   handleChange = e => {
-    const { id, onChange, setField } = this.props
-    setField(id, e.target.value)
-    if (onChange) { onChange(e.target.value) }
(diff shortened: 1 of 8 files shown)


src/app/core/common/TextField.js
```diff
@@ -18,9 +18,13 @@ class TextField extends React.Component {
...
 
   handleChange = e => {
-    const { id, onChange, setField } = this.props
-    setField(id, e.target.value)
-    if (onChange) { onChange(e.target.value) }
+    const { id, onChange, setField, type } = this.props
+    // HTML specs says that <input type="number"> return strings but it's more useful if we
+    // convert it to a `Number` to reduce type casting all over the place.
+    const strVal = e.target.value
+    const value = (type && type.toLowerCase() === 'number' && strVal !== '') ? Number(strVal) : strVal
```
kevinylee1991: Just a note for future issues, I think sometimes an API may request a number in string form

PR #184, opened 2018-10-03
Create volumes from existing sources

brennancheung:
I created a `<ListTableSelect>` component to allow us to have a picker in the form of a table.

screen shot 2018-10-03 at 4 39 18 pm

Next step is to integrate it into the `AddVolumeForm`.

### DIFF
--- a/src/app/core/common/EnhancedTableHead.js
+++ b/src/app/core/common/EnhancedTableHead.js
@@ -11,11 +11,15 @@ import {
 
 class EnhancedTableHead extends React.Component {
   createSortHandler = property => event => {
-    this.props.onRequestSort(event, property)
+    const { onRequestSort } = this.props
+    if (onRequestSort) {
+      onRequestSort(event, property)
+    }
   }
 
   render () {
     const {
+      blankFirstColumn,
       checked,
       columns,
       numSelected,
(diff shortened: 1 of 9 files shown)


src/app/core/common/EnhancedTableHead.js
```diff
@@ -38,10 +42,13 @@ class EnhancedTableHead extends React.Component {
       </TableCell>
     ) : null
 
+    const firstBlank = blankFirstColumn ? <TableCell padding="checkbox" key="_firstBlank" /> : null
```
kevinylee1991: Is this to create extra space, or is there some deeper meaning to having a blank column? If it's to create space, would using css make more sense?

src/app/core/common/ListTableSelect.js
```diff
@@ -0,0 +1,109 @@
+import React from 'react'
```
kevinylee1991: just a simple note, in the future we will need a multi-select table list

src/app/core/common/EnhancedTableHead.js
brennancheung: This is similar to keeping the first column clear for the checkboxes.  Only in this case it is `radio` buttons.  Thought we should make it more generic but the component is already very hardwired and coupled.  It's going to take a lot of refactoring to make it more generic so I'm just going with this for now.

PR #185, opened 2018-10-08
Qbert API

brennancheung:
Here's the start of the Qbert API.  I still want to add some more tests and rename the client to `api-client` since it is not *just* OpenStack APIs now.


brennancheung:
2nd commit is just a rename.  All the actual code changes are in the 1st commit.

PR #189, opened 2018-10-11
Add Cloud Providers functionality

brennancheung:
This is the start of the cloud providers functionality.  It is still WIP.

I created a number of utility functions to make generating the components much simpler.  This removes a lot of the boilerplate we were previously doing.  If the pattern works out we can refactor and remove a lot of code.


brennancheung:
This should be complete now.

I did some refactoring to make more of our components work with Storybook so we can develop forms in isolation.  This should speed up development and make it easier.

src/app/core/createAddComponents.js
```diff
@@ -0,0 +1,54 @@
...
+      )
+    }
+  }
+
+  const AddPage = compose(
+    withAppContext,
+    withRouter,
+    requiresAuthentication,
+  )(AddPageBase)
+  AddPage.displayName = `Add${name}Page`
```
kevinylee1991: just curious, where does displayName get used? Wasn't sure if the user will see the string anywhere in the ui

brennancheung: Dynamically generated classes don't have their names showing up in debug scenarios.  This gives it a name when looking at stack traces and also if you have the React devtools Chrome extension installed.  Otherwise you get these generic `Anonymous` classes everywhere that makes things difficult to know what is what.

PR #195, opened 2018-10-31
Misc scattered worked for K8s Infrastructure

brennancheung:
There's still work to be done on this.  This is just the first pass at it.

Note, the `loadInfrastructure` action.  This should allow us to complete parts of the data model and have the UI render at least something to begin with.  Then as data from the rest of the APIs come in, we can add more detail to the existing data model.  This will result in the UI being more responsive because it will progressively render.

Nodes and clusters need to be loaded at the same time in order to build the data models that the UI requires.  There's a bit of business logic that would be better served if this was pushed into the API layer in the backend somehow.  That way the required data model fits the API.


src/api-client/Qbert.js
```diff
@@ -39,19 +38,37 @@ class Qbert {
...
   /* Cloud Providers Types */
-
   async getCloudProviderTypes () {
     return this.client.basicGet(`${await this.baseUrl()}/cloudProvider/types`)
   }
 
+  /* Nodes */
+  getNodes = async () => {
+    const nodes = await this.client.basicGet(`${await this.baseUrl()}/nodes`)
+    const scopedNodes = nodes.filter(x => x.projectId === this.client.activeProjectId)
```
kevinylee1991: this is probably unnecessary, the qbert API already scopes the returned nodes

PR #200, opened 2018-11-06
Partial work towards "combinedHosts"

brennancheung:
This a WIP.

Not sure the best location for "combinedHosts" functionality since it spans both OpenStack, Kubernetes, and our own ResMgr.

For the time being I'm putting it with `loadInfrastructure` in the kubernetes section since that's the closest place that has most of the data from APIs already loaded.

CombinedHosts is rather massive and contains a lot of random rules and business logic.  We might want to think about restructuring the code and making it more modular.

### DIFF
--- a/src/app/core/fp.js
+++ b/src/app/core/fp.js
@@ -39,6 +39,17 @@ export const filterFields = (...keys) => obj =>
     {}
   )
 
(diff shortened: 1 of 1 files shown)


brennancheung:
Somehow I think this branch got combined into the hackathon9 branch and is already in master so closing this out.

PR #204, opened 2018-11-07
Info help forms and ValidatedForm refactoring

xantorres:
For now the info is shown just on hover for TextField and Checkbox inside a ValidatedForm, but behavior is easily customizable from the inputs.
image


src/app/core/common/Checkbox.js
```diff
@@ -48,13 +48,18 @@ class Checkbox extends React.Component {
     }
   }
 
+  showInfo = e => {
```
brennancheung: Any way we can extract out the `onHover` / `showInfo` into a common HOC that implements this logic?

src/app/core/common/TextField.js
```diff
@@ -19,13 +19,14 @@ const styles = theme => ({
...
-    const spec = props.required
-      ? {
-        validations: Array.isArray(props.validations)
-          ? [requiredValidator, ...props.validations]
-          : { required: true, ...props.validations }
-      }
-      : pickMultiple('validations')(props)
+    const spec = pickMultiple('validations', 'info')(props)
+
+    if (props.required) {
```
brennancheung: This logic seems to be repeated in each component as well.  Can we pass in a generic validation method as a prop and just have the component use that?  Then if needs be, the component can do additional validation logic.

src/app/core/common/ValidatedForm.js
```diff
@@ -177,17 +203,34 @@ export const withFormContext = Component => props => (
         errors={errors}
         showErrorsOnBlur={showErrorsOnBlur}
         validateField={validateField}
+        showInfo={showInfo}
       />
     )}
   </Consumer>
 )
 
+export const FormInfoPlaceholder = compose(
```
brennancheung: Container seems like it might be a better term than Placeholder.  Placeholder is usually default text that eventually gets replaced.

src/app/core/common/TextField.js
xantorres: I refactored all validation and info hint logic it in a component called `ValidatedFormInput` used by the `withFormContext` HoC, so all this logic is now abstracted there

src/app/core/common/ValidatedForm.js
xantorres: I removed this form and now I'm rendering this in a fixed container inside the `ValidatedForm`

src/app/core/common/Checkbox.js
xantorres: done

src/app/core/createFormComponent.js
```diff
@@ -14,7 +14,7 @@ const createFormComponent = ({ submitLabel, initialValue, displayName, fields })
   }
 
   const Form = ({ onComplete }) => (
-    <ValidatedForm onSubmit={onComplete} initialValue={initialValue}>
+    <ValidatedForm onSubmit={onComplete} initialValues={initialValue}>
```
brennancheung: Hmm, on the fence about this one.  Within the React community, `initialValue` is the normal convention for controlled components.  Guess this makes sense though.  It does add some clarity even if we don't use the same naming convention.

src/app/core/common/validated_form/withFormContext.js
```diff
@@ -0,0 +1,134 @@
...
+      initialValues,
+      values,
+      defineField,
+      setFieldValue,
+      showErrorsOnBlur,
+      validateField,
+      errors,
+      showInfo
+    }) => (
+      <ValidatedFormInput
```
brennancheung: Love it!  This HOC makes the implementation in the Fields themselves much simpler.

PR #206, opened 2018-11-13
Hackathon9

brennancheung:
Developer tools plugin for Platform9's Hackathon 9


src/app/core/fp.js
```diff
@@ -103,3 +105,34 @@ export const emptyObj = Object.freeze({})
...
 export const emptyArr = Object.freeze([])
+
+export const pathOrNull = pathStr => pathOr(null, pathStr.split('.'))
+
+// I didn't see anything in Ramda that would allow me to create a "Maybe"
+// composition so creating a simple version here.
+// With long chains of functions it can get annoying to make sure each one
+// contains a valid value before continuing.  This HOF performs a pipe but
+// only when each function returns something truthy.
+export const pipeWhenTruthy = (...fns) => arg => {
```
xantorres: Maybe we could incorporate something like https://github.com/ramda/ramda-fantasy

brennancheung: I came across that when searching.  I'm on the fence with it.  Unless developers understand monadic composition it can get confusing to them real fast.  Looks like `ramda-fantasy` is no longer maintained but they list alternatives that seem interesting.  Might have to investigate later on.

PR #207, opened 2018-11-13
Bugfix/form validation issue

xantorres:


src/app/core/common/validated_form/ValidatedForm.js
```diff
@@ -119,12 +124,15 @@ class ValidatedForm extends React.Component {
...
+  /**
+   * Validate all fields and return false if any error is found, true otherwise
+   */
   validateForm = () => {
     const { fields } = this.state
     const results = Object.keys(fields).map(field =>
       this.validateField(field)
     )
-    return results.includes(true)
+    return !results.includes(false)
```
brennancheung: We can clean this up with `Array#every`.  `Object.keys(fields).every(field => this.validateField(field))`.  Not sure if binding will stick but maybe even `Object.keys(fields).every(this.validateField)`.

xantorres: I need to fully iterate all fields to show errors on all them, not just the first one, so in this case I can't use "every"

PR #209, opened 2018-11-15
Cluster links

brennancheung:
This PR adds the Kubernetes Dashboard link as well as the CLI.

For the CLI I just used a `<Dialog fullScreen>` instead of the complex tab solution we have now.  The iTerm like tab solution might not be necessary and is a lot of custom code so I opted for the simpler solution for now.

I'm punting on the kubeconfig since I don't like saving the username / password in the client and re-using it.  We already have an auth mechanism so the backend should really provide a proper endpoint for this functionality.

### DIFF
--- a/src/app/plugins/developer/components/ApiHelper.js
+++ b/src/app/plugins/developer/components/ApiHelper.js
@@ -57,8 +57,8 @@ class ApiHelper extends React.Component {
 
     const response = await {
       GET:    () => apiClient.basicGet(finalUrl),
-      POST:   () => apiClient.basicPost(finalUrl, body),
-      PUT:    () => apiClient.basicPut(finalUrl, body),
+      POST:   () => apiClient.basicPost(finalUrl, JSON.parse(body)),
+      PUT:    () => apiClient.basicPut(finalUrl, JSON.parse(body)),
       DELETE: () => apiClient.basicGet(finalUrl),
     }[method]()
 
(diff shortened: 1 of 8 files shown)


src/app/plugins/developer/components/ApiHelper.js
```diff
@@ -67,8 +67,14 @@ class ApiHelper extends React.Component {
 
   setField = key => value => this.setState({ [key]: value })
 
-  handleServiceChange = ({ service, baseUrl }) => {
-    this.setState({ service, baseUrl })
+  handleServiceChange = async ({ service, baseUrl }) => {
+    const { apiClient } = this.props.context
+    this.setState({ service })
```
kevinylee1991: does service need to be set first? or can it be included in the other setStates below?

src/app/plugins/kubernetes/components/infrastructure/ClustersListPage.js
```diff
@@ -1,11 +1,35 @@
+import React from 'react'
 import createCRUDComponents from 'core/createCRUDComponents'
+import DownloadKubeConfigLink from './DownloadKubeConfigLink'
+import KubeCLI from './KubeCLI'
+import { loadInfrastructure } from './actions'
+
+// Prevent the cluster row from being selected
+const stopPropagation = e => e.stopPropagation()
+
+const Link = ({ src, Icon, label }) => (
```
kevinylee1991: this thing could probably be a core component -- the pods/deployments/services gridviews will need this too

src/app/plugins/developer/components/ApiHelper.js
brennancheung: It technically doesn't need to be set separately, but it makes the code easier because `baseUrl` is going to be conditionally set but `service` is always unconditional.

src/app/plugins/kubernetes/components/infrastructure/ClustersListPage.js
brennancheung: Good idea.  I'll make a common component.

PR #213, opened 2018-11-21
Table column improvements

xantorres:
Some minor refactoring and added list table columns selection option (`editableColumns` prop) + column reordering by dragging (`draggableColumns` prop)

By explicitly passing `display: false` on the column definition, the column will be hidden by default but selectable in the Column selector, while when passing `display: 'excluded'`, the column will be excluded in both the table and the column selector.

image


src/app/core/common/list_table/ListTable.js
```diff
@@ -41,6 +33,10 @@ class ListTable extends React.Component {
     const prefs = (getUserPreferences && getUserPreferences()) || {}
 
     this.state = {
+      visibleColumns: columns
+        .filter(column => column.display !== false && column.display !== 'excluded')
```
brennancheung: `column.display` having multiple types (string, boolean) is a bit weird.  It might be simple enough to let slide but I'm wondering if it makes more sense to use 2+ separate booleans and then have a short function to determine if we should ultimately show it.  Might make refactoring and additional changes easier.

src/app/core/common/list_table/ListTable.js
```diff
@@ -166,6 +162,26 @@ class ListTable extends React.Component {
...
 
+  handleColumnsChange = columnId => {
+    this.setState(({visibleColumns}) => ({
+      visibleColumns: visibleColumns.includes(columnId)
+        ? except(columnId, visibleColumns)
+        : [...visibleColumns, columnId]
+    }))
+  }
+
+  handleColumnsSwitch = (srcColumnId, tarColumnId) => {
```
brennancheung: What's `tar`?  target?  `dest` is a bit more common and less ambiguous.

src/app/core/common/list_table/ListTable.js
```diff
@@ -166,6 +162,26 @@ class ListTable extends React.Component {
     })
   }
 
+  handleColumnsChange = columnId => {
+    this.setState(({visibleColumns}) => ({
+      visibleColumns: visibleColumns.includes(columnId)
```
brennancheung: This is a toggle?  Not obvious at a glance.  Also, should probably be singular.  `handleColumnToggle` maybe.

src/app/core/common/list_table/ListTable.js
```diff
@@ -372,6 +399,9 @@ ListTable.propTypes = {
 
   showCheckboxes: PropTypes.bool,
   searchTarget: PropTypes.string,
+
+  editableColumns: PropTypes.bool,
```
brennancheung: `canEditColumns` would be less confusing.  The name makes it sound like it is an array of columns.

src/app/core/common/list_table/ListTable.js
```diff
@@ -372,6 +399,9 @@ ListTable.propTypes = {
...
+  draggableColumns: PropTypes.bool,
```
brennancheung: same

src/app/core/common/list_table/ListTableColumnsPopover.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Checkbox from '@material-ui/core/Checkbox'
```
brennancheung: Should import all these in a single import?

src/app/core/common/list_table/ListTableColumnsPopover.js
```diff
@@ -0,0 +1,97 @@
...
+import { withStyles } from '@material-ui/core/styles'
+import moize from 'moize'
+
+const styles = {
+  root: {
+    padding: '16px 24px 16px 24px',
+    fontFamily: 'Roboto',
+  },
+  title: {
+    marginLeft: '-7px',
```
brennancheung: I think you can leave `px` off and make it an `int` in `@material-ui`.  So `marginLeft: -7`.  `padding` still needs it though since you have multiple.  `px` is the default unit I think.

src/app/core/common/list_table/ListTableColumnsPopover.js
```diff
@@ -0,0 +1,97 @@
...
+  label: {
+    fontSize: '15px',
+    marginLeft: '8px',
+    color: '#4a4a4a',
+  },
+}
+
+class ListTableColumnsPopover extends React.Component {
+  static propTypes = {
+    columns: PropTypes.array.isRequired,
```
brennancheung: Specify the type of the array as well.  (`PropTypes.arrayOf`)

src/app/core/common/list_table/ListTableColumnsPopover.js
```diff
@@ -0,0 +1,97 @@
...
+
+  render () {
+    const { classes, columns, visibleColumns } = this.props
+    return (
+      <FormControl component="fieldset" className={classes.root} aria-label="Columns">
+        <Typography variant="caption" className={classes.title}>
+          Columns
+        </Typography>
+        <FormGroup className={classes.formGroup}>
+          {columns.map(column => {
```
brennancheung: What about doing `filter` then `map` instead of doing the conditional inside the map?

src/app/core/common/list_table/ListTableToolbar.js
```diff
@@ -2,49 +2,56 @@ import React from 'react'
...
-  Typography
-} from '@material-ui/core'
+import { Button, IconButton, Toolbar, Tooltip, Typography } from '@material-ui/core'
 import { lighten } from '@material-ui/core/styles/colorManipulator'
 import AddIcon from '@material-ui/icons/Add'
 import DeleteIcon from '@material-ui/icons/Delete'
 import EditIcon from '@material-ui/icons/Edit'
 import FilterListIcon from '@material-ui/icons/FilterList'
 import SearchBar from 'core/common/SearchBar'
+import ListTableColumnsSelector from 'core/common/list_table/ListTableColumnsSelector'
```
brennancheung: `ListTableColumnSelector` (singular) makes more sense to me.  Could go either way on this one though.

src/app/core/common/list_table/ListTable.js
xantorres: This is the a switching of the columns when dragging, so maybe `handleColumnsSwitch` would be more convenient

src/app/core/common/list_table/ListTableColumnsPopover.js
xantorres: good to know

src/app/core/common/list_table/ListTableColumnsPopover.js
xantorres: good point

PR #215, opened 2018-11-27
More cluster list columns

brennancheung:
This adds some more fields to the clusters list view.

Usage stats are not correct for some reason.  I'm not sure why the totals don't match the old UI.  Still investigating that part.

### DIFF
--- a/src/app/core/common/list_table/ListTable.js
+++ b/src/app/core/common/list_table/ListTable.js
@@ -336,7 +336,7 @@ class ListTable extends React.Component {
               searchTerm={searchTerm}
               columns={columns}
               visibleColumns={visibleColumns}
-              onColumnsChange={canEditColumns && this.handleColumnSwitch}
+              onColumnsChange={(canEditColumns && this.handleColumnSwitch) || undefined}
             />
             <div className={classes.tableWrapper}>
               <Table className={classes.table}>
--- a/src/app/core/common/list_table/ListTable.js
+++ b/src/app/core/common/list_table/ListTable.js
@@ -336,7 +336,7 @@ class ListTable extends React.Component {
               searchTerm={searchTerm}
               columns={columns}
               visibleColumns={visibleColumns}
-              onColumnsChange={(canEditColumns && this.handleColumnSwitch) || undefined}
+              onColumnsChange={canEditColumns ? this.handleColumnSwitch : null}
             />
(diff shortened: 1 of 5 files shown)


brennancheung:
Looks like the stats might actually be correct.  The memory and storage look correct but the CPU was not showing as the same as the old UI.  On further investigation if I keep refreshing the page in the old UI I get a different value for CPU utilization each time.  Looks like it is within the same ballpark (1-2%) so I'm assuming it is actually correct.

I also added a few other columns and created a utility `castFuzzyBool` to help with the "booleans" being all over the place.

src/app/core/common/list_table/ListTable.js
```diff
@@ -336,7 +336,7 @@ class ListTable extends React.Component {
               searchTerm={searchTerm}
               columns={columns}
               visibleColumns={visibleColumns}
-              onColumnsChange={canEditColumns && this.handleColumnSwitch}
+              onColumnsChange={(canEditColumns && this.handleColumnSwitch) || undefined}
```
brennancheung: Xan might be fixing this in another PR.  Not sure his solution so I might want to change this based on what changes elsewhere.

src/app/plugins/kubernetes/components/infrastructure/ClustersListPage.js
```diff
@@ -16,32 +17,81 @@ const renderLinks = links => {
...
+      }
+    }
+  }
+
+  const { compute, memory, disk } = clusterWithStats.usage
+  clusterWithStats.usage.compute.percent = Math.round(100 * compute.current / compute.max)
+  clusterWithStats.usage.memory.percent = Math.round(100 * memory.current / memory.max)
+  clusterWithStats.usage.disk.percent = Math.round(100 * disk.current / disk.max)
+
+  return <pre>{JSON.stringify(clusterWithStats.usage, null, 4)}</pre>
```
brennancheung: This is a temporary placeholder.  We will replace once we have some components that can render these data.  I'm also thinking about splitting this out into separate fields so we can sort them individually.

src/app/core/common/list_table/ListTable.js
xantorres: `canEditColumns ? this.handleColumnSwitch : null` but don't worry, I'll fix and other passed props in my next PR

Issue #219, opened 2018-11-29
Navbar component improvements / bugfixes


brennancheung:
I noticed a few issues with the `Navbar` component and the new hierarchical schema.

1. There isn't enough visual distinction between a link and its `nestedLinks`.

See:

![screen shot 2018-11-28 at 4 46 11 pm](https://user-images.githubusercontent.com/289156/49191938-9beeda80-f32d-11e8-8eb4-7b0919eab189.png)

Clusters, Nodes, and Cloud Providers are all nested under Infrastructure but you can't tell that from the sidenav.  The storybook has icons which makes the indentation more obvious but without icons it is difficult to tell.

2.  Clicking on Infrastructure by itself does not navigate anywhere.  It just expands / contracts the accordion.

3.  The `link.path` for `nestedLinks` do not inherit the path prefix from the plugin.  They become absolute paths instead of relative to the plugin.

4. Switching between routes that have different hashes in the URL (e.g., `infrastructure#clusters` and `infrastructure#nodes` does not change the actual tabs).  The URL changes but it stays on the old component.  I'm not sure yet whether this needs to be fixed in `Navbar` or if `Tabs` needs some kind of route watching.

PR #221, opened 2018-11-29
Navbar improvements and fixes for #219

xantorres:
Fixes for #219

### DIFF
--- a/src/app/core/common/Navbar.js
+++ b/src/app/core/common/Navbar.js
@@ -3,7 +3,7 @@ import { withRouter } from 'react-router'
 import { assoc, flatten, prop } from 'ramda'
 import moize from 'moize'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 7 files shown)


src/app/core/common/Navbar.js
```diff
@@ -1,23 +1,9 @@
...
-  ExpansionPanelSummary,
-  IconButton,
-  InputBase,
-  ListItemIcon,
-  ListItemText,
-  MenuItem,
-  MenuList,
-  Typography,
-} from '@material-ui/core'
+import { Collapse, Divider, Drawer, ExpansionPanel, ExpansionPanelDetails, ExpansionPanelSummary, IconButton, InputBase, ListItemIcon, ListItemText, MenuItem, MenuList, Typography } from '@material-ui/core'
```
brennancheung: With this many items we should problem keep vertical.

PR #222, opened 2018-11-30
Cluster details page

brennancheung:
Issue: #222 

I added an `<InfoPanel>` component that we can reuse elsewhere.

I also noticed the renderProp pattern was getting a bit messy with `<DataLoader`> and requiring more `import`s than I wanted.  I created a `withDataLoader` helper that cleans up the code a bit.

screen shot 2018-12-03 at 4 37 48 pm

screen shot 2018-12-05 at 5 57 11 pm

### DIFF
--- a/src/app/core/fp.js
+++ b/src/app/core/fp.js
@@ -146,3 +146,5 @@ export const ensureFunction = moize(maybeFunc => (...args) => {
   }
   return maybeFunc
 })
+
+export const maybeFnOrNull = fn => value => value ? fn(value) : null
--- a/src/app/plugins/kubernetes/components/infrastructure/ClusterNodes.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/ClusterNodes.js
@@ -4,16 +4,38 @@ import { withAppContext } from 'core/AppContext'
 import { compose } from 'ramda'
 import { loadInfrastructure } from './actions'
 import { withDataLoader } from 'core/DataLoader'
+import { utilizationRenderer } from './NodesListPage'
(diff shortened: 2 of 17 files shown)


src/app/core/fp.js
```diff
@@ -146,3 +146,5 @@ export const ensureFunction = moize(maybeFunc => (...args) => {
   }
   return maybeFunc
 })
+
+export const maybeFnOrNull = fn => value => value ? fn(value) : null
```
xantorres: Wouldn't make more sense to name this `callWhenValue` or something like this? Also not sure of the use case of this but what if a `null`, `0` or `''` is passed as a value? Is that still intended to return a `null`?

src/app/plugins/kubernetes/components/infrastructure/ClusterNodes.js
```diff
@@ -0,0 +1,60 @@
...
+  const cluster = data.find(x => x.uuid === match.params.id)
+  return (
+    <div>
+      <ListTable
+        title="Cluster Nodes"
+        columns={columns}
+        data={cluster.nodes}
+        uniqueIdentifier="uuid"
+      />
+      <pre>{JSON.stringify(cluster.nodes, null, 4)}</pre>
```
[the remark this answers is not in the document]
brennancheung: yes, this is gone now.

src/app/core/fp.js
brennancheung: In this case I wanted to emphasize that `null` was being returned.  This will be used inside of a `render` so returning `0`, `''`, `undefined`, or `false` wouldn't have the same effect.

Issue #216, opened 2018-11-27
Resource utilization with "progress" bars


brennancheung:
We need a generic component that is a "progress" bar.  It takes a value 0.0 to 100.0 and fills in a horizontal rectangle based on this value.

Once we have that we need another component to render the usage stats.

See the attached screenshot for an example of where it will be used (data structure on the left, old UI on the right):

![screen shot 2018-11-26 at 4 28 53 pm](https://user-images.githubusercontent.com/289156/49050606-002c6580-f199-11e8-8eff-1ddc179f50a8.png)

brennancheung:
There's also another place where a similar UI exists.  We might want to combine them in the new UI.

![screen shot 2018-11-26 at 4 53 07 pm](https://user-images.githubusercontent.com/289156/49051246-dde81700-f19b-11e8-9f2e-a5c865bafb8f.png)

I'm leaning towards something somewhere between the 2 UI's.  It should be the same across each.  I'm also thinking we should split resource utilization into 3 different columns.  You can't sort a column when the column has 3 different values.

brennancheung:
Now that I think about it, let's add separate components for both variants.  We can let the user chose which view they prefer by turning columns on and off.

PR #223, opened 2018-11-30
Feature/progress bar

xantorres:
#216 

image

Compact:
image

I also updated `@material-ui/core` to 3.6.0 and added `@material-ui/styles` to be able to use new render props API: https://material-ui.com/css-in-js/basics/#adapting-render-props-api

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -27,14 +27,15 @@
     }],
     "jsx-quotes": 0,
     "react/prop-types": 0,
+    "react/no-unused-prop-types": "warn",
     "no-new": 0,
     "space-infix-ops": 0,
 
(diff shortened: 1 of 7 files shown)


.eslintrc
```diff
@@ -27,14 +27,15 @@
     }],
     "jsx-quotes": 0,
     "react/prop-types": 0,
+    "react/no-unused-prop-types": "warn",
```
xantorres: I was getting eslint complains about a prop that was not being used anywhere in the component, while actually it was being used when rendering the styles (using https://material-ui.com/css-in-js/basics/#adapting-render-props-api) so I thought it would be sensitive to have this setting just as a warning rather than as error.

PR #225, opened 2018-11-30
Added clusters CRUD to simulator, along with some prerequisites like …

kevinylee1991:
…nodes and namespaces list

Tested all operations implemented not through the browser API, but through the CreateModel API since the UI is not fleshed out yet.

I just realized the branch name is misleading

### DIFF
--- a/src/server/models/qbert/Cluster.js
+++ b/src/server/models/qbert/Cluster.js
@@ -1,0 +1,112 @@
+import createModel from '../createModel'
+import faker from 'faker'
+import { getCurrentTime } from '../../util'
+import ResMgrHost from '../resmgr/ResMgrHost'
+import Node from './Node'
+import { attachNodeToCluster } from './Operations'
+import uuid from 'uuid'
+
+const clusterDefaults = {
+  allowWorkloadsOnMaster: 0,
+  appCatalogEnabled: 1,
+  authzEnabled: 1,
+  canUpgrade: false,
+  cloudProperties: {},
+  cloudProviderName: 'someCloudProvider',
+  cloudProviderType: 'local',
+  containersCidr: '10.0.10.0/16',
+  debug: 'false',
+  dockerRoot: '/var/lib',
+  enableMetallb: 0,
+  etcdDataDir: '/var/opt/pf9/kube/etcd/data',
+  externalDnsName: '',
+  flannelIfaceLabel: '',
+  flannelPublicIfaceLabel: '',
+  isKubernetes: 1,
+  isMesos: 0,
+  isSwarm: 0,
+  k8sApiPort: '443',
+  keystoneEnabled: 1,
+  lastOk: '',
+  lastOp: '',
+  masterIp: '10.0.10.20',
+  masterVipIface: 'ens3',
+  masterVipIpv4: '10.0.10.20',
+  metallbCidr: '',
+  name: '',
+  networkPlugin: 'flannel',
+  nodePoolName: 'defaultPool',
+  numMasters: 0,
+  numWorkers: 0,
+  privileged: 1,
+  projectId: '',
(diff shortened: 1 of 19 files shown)


src/server/models/qbert/Cluster.js
```diff
@@ -0,0 +1,112 @@
...
+      }
+    })
+
+    const currentTime = getCurrentTime()
+    return { nodePoolUuid: uuid.v4(), cloudProperties: cloudProperties, cloudProviderUuid: uuid.v4(), created_at: currentTime, lastOk: currentTime, lastOp: currentTime, ...input }
+  },
+  loaderFn: (clusters) => {
+    return clusters.map((cluster) => {
+      const newCluster = { ...cluster }
+      delete newCluster.nodes
```
brennancheung: Why do we need to delete nodes?

src/server/models/qbert/CreateModel.js
```diff
@@ -8,6 +8,7 @@ const createModel = (options={}) => {
     defaults,
     loaderFn, // function that maps data into a different form for API route
     mappingFn, // data coming form API, context needs it in a different format
+    onDeleteFn,
```
brennancheung: `CreateModel` should be consistent with the object name (so lowercase `createModel`).

Also, this looks like a generic utility that has nothing to do with `qbert`.  Perhaps this should be moved to a common location.

src/server/presets/dev.js
```diff
@@ -107,11 +111,26 @@ function loadPreset () {
...
-  new ResMgrHost({ roles: ['pf9-ostackhost'], info: { hostname: 'fake resmgr host' } })
+  const resMgrHost = new ResMgrHost({ roles: ['pf9-ostackhost'], info: { hostname: 'fake resmgr host' } })
+  const resMgrHost2 = new ResMgrHost({ roles: ['pf9-ostackhost'], info: { hostname: 'fake resmgr host 2' } })
 
   // Cloud Providers
   CloudProvider.create({ name: 'fakeCp1', type: 'aws' }, context)
   CloudProvider.create({ name: 'fakeCp2', type: 'openstack' }, context)
+
+  // Clusters
+  const cluster = Cluster.create({ name: 'fakeCluster1', sshKey: 'someKey' }, context)
```
brennancheung: Should the models import `context` themselves instead of having it passed to them?  I guess if we have multiple contexts this makes sense but I'm not aware of a use case where we would need that.

src/server/models/qbert/Cluster.js
```diff
@@ -0,0 +1,112 @@
...
+const options = {
+  dataKey: 'clusters',
+  uniqueIdentifier: 'uuid',
+  defaults: clusterDefaults,
+  mappingFn: (input, context) => {
+    const defaultKeys = Object.keys(clusterDefaults)
+    const cloudProperties = {}
+    if (input.numMasters || input.numWorkers) {
+      // Add nodes property if nodes are going to be created too
+      input.nodes = []
```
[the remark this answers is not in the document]
kevinylee1991: the defaults get merged in after this part runs -- am open to suggestions to make specifying nodes property less confusing

src/server/models/qbert/Cluster.js
kevinylee1991: nodes property is not returned by the real qbert API, but I'm using nodes on the simulator side to track the nodes attached to the clusters. If it is clearer to keep cluster/node association separate from the object I could possibly create a new model that just keeps track of the associations

src/server/presets/dev.js
kevinylee1991: i have thought about this as well since there are benefits to both ways -- passing in context to the functions has been a pattern so far so I decided to follow that

PR #229, opened 2018-12-03
Improve nav menu filtering and allow navigating using keyboard (arrows and enter keys)

xantorres:
Missing: keep section expanded after navigating to a sublink, improve styling

navbar

### DIFF
--- a/src/app/core/common/HotKeysProvider.js
+++ b/src/app/core/common/HotKeysProvider.js
@@ -12,17 +12,29 @@ const HotKeysContext = React.createContext({
 })
 const keyHandlersLens = lensProp('hotKeyHandlers')
 const editableInputs = ['INPUT', 'SELECT', 'TEXTAREA']
-const editEndKeys = ['Enter', 'Escape', 'Tab', 'ArrowUp', 'ArrowDown']
+const pressingSpecialKey = e => e.metaKey || e.altKey || e.shiftKey
+const isEditableInput = e => {
+  const target = e.target || e.srcElement
+  // input, select, and textarea are always considered editable inputs
+  return editableInputs.includes(target.tagName) ||
+    target.isContentEditable
+}
 
 export default class HotKeysProvider extends Component {
   state = {
     hotKeyHandlers: {},
 
+    /**
+     * Assign a key down handler for a specific key
+     * @param key key to bind
+     * @param fn function to trigger
+     * @param options { whileEditing, ctrlKey }
+     */
     setHotKeyHandler: (key, fn, options) => {
       this.setState(over(keyHandlersLens,
         handlers => ({
           ...handlers,
-          [key]: [...ensureArray(handlers[key]), {fn, options}]
+          [key]: [...ensureArray(handlers[key]), { fn, options }]
         })
       ))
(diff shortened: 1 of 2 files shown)


src/app/core/common/HotKeysProvider.js
```diff
@@ -48,19 +49,18 @@ export default class HotKeysProvider extends Component {
...
   }
 
   handleKeyDown = e => {
     const hotkeyHandlers = this.state.hotKeyHandlers[e.key]
-
     if (hotkeyHandlers &&
       !isEmpty(hotkeyHandlers) &&
       !e.metaKey && !e.altKey && !e.shiftKey &&
-      !this.isEditable(e)) {
+      (editEndKeys.includes(e.key) || !this.isEditable(e))) {
```
brennancheung: This is getting messy.  What about moving all the boolean conditionals into their own function?

src/app/core/common/Navbar.js
```diff
@@ -124,12 +134,48 @@ class Navbar extends PureComponent {
...
   }
 
+  handleArrowKeys = direction => () => {
+    const { filterText, activeNavItem } = this.state
+    if (filterText && activeNavItem) {
+      // Highlight next nav item
+      const offset = direction === 'ArrowDown' ? 1 : -1
+      const sectionLinks = this.getSectionLinks()
+      const currentIdx = sectionLinks.findIndex(propEq('name', activeNavItem))
+      const nextIdx = currentIdx + offset >= sectionLinks.length ? 0
```
brennancheung: Getting a bit complicated here.  Is this just for wrapping around?  What about using modulus?

`(currentIdx + offset) % sectionLinks.length`

src/app/core/common/Navbar.js
```diff
@@ -246,35 +321,34 @@ class Navbar extends PureComponent {
...
+            <ExpansionPanelDetails className={classes.navBody}>
+              <MenuList component="nav" className={classes.navMenu}>
+                {(filterText
+                  ? this.getFilteredLinks(section.links)
+                  : section.links
+                ).map(this.renderNavLink)}
+              </MenuList>
+            </ExpansionPanelDetails>
+          </ExpansionPanel>)
+        : <MenuList component="nav" className={classes.navMenu}>
```
brennancheung: There's too many lines between the ternary operator terms.  What about just making it more explicit?

```javascript
{filteredSections.length > 1 && ...}
{filteredSections.length <= 1 && ..}
```

brennancheung: Moving the blocks into functions might work as well and even be cleaner.

xantorres: Yeah I had pending doing a little refactoring and cleanup

src/app/core/common/Navbar.js
xantorres: I want to go to iterate between first/last items when reaching the start/end of the list, that's why I do this. 
I will try to simplify this a little more or at least add a comment explain the goal of this.

xantorres: Good advice, that worked!

PR #237, opened 2018-12-11
Add cluster resource widgets to cluster details and clusters list

brennancheung:
screen shot 2018-12-10 at 4 52 46 pm
screen shot 2018-12-10 at 3 59 46 pm

### DIFF
--- a/src/app/core/common/dashboard_graphs/UsageBar.js
+++ b/src/app/core/common/dashboard_graphs/UsageBar.js
@@ -7,11 +7,12 @@ const UsageBar = ({ precision, stats }) => {
 
   const curStr = current.toFixed(precision)
   const maxStr = max.toFixed(precision)
+  const label = <span><strong>{percent}%</strong> - {curStr} / {maxStr}{units} {type}</span>
   return (
     <ProgressBar
       compact
       percent={percent}
-      label={p => <span><strong>{p}%</strong> - {curStr} / {maxStr}{units} {type}</span>}
+      label={label}
     />
   )
 }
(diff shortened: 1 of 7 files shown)


src/app/core/common/dashboard_graphs/UsageBar.js
```diff
@@ -0,0 +1,34 @@
...
+const UsageBar = ({ precision, stats }) => {
+  const { current, max, percent, type, units } = stats
+
+  const curStr = current.toFixed(precision)
+  const maxStr = max.toFixed(precision)
+  return (
+    <ProgressBar
+      compact
+      percent={percent}
+      label={p => <span><strong>{p}%</strong> - {curStr} / {maxStr}{units} {type}</span>}
```
kevinylee1991: this line confuses me some but it all looks good to me

brennancheung: I might be able to use string interpolation instead but not sure it would make it any cleaner.  Suggestions?

brennancheung: Or is it the fact that it is a function?

xantorres: you can pass a string too, if you want, but you will lose the `<strong>` tags.
What you can also do is to pass a react component directly, without a function, like `<span><strong>{percent}%</strong> - {curStr} / {maxStr}{units} {type}</span>` as the percent variable is already in the scope.
The function way is aimed towards having a separate function outside of the scope to render the label.

PR #238, opened 2018-12-11
Added PreferencesProvider and withPreferences HoC and some refactoring

xantorres:
* Created `PreferencesProvider` component, `withPreferences` (for initialization) and `withScopedPreferences` (to use in components) HoCs used to manage user settings.
* Created session related methods in `AppContext`: `initSession`, `updateSession` and `destroySession` to make session related logic less relying on plugin specific methods.
* Simplified openstack `initialSetup` method by delegating session and preferences logic to `AppContext` and `PreferencesProvider` respectively
* Implemented `withScopedPreferences` to all components using `ListTable`, in order to automatically apply user specific preferences when used.


src/app/core/common/list_table/ListTable.js
```diff
@@ -24,21 +29,23 @@ const styles = theme => ({
   },
 })
 
+export const pluckVisibleColumnIds = columns => pluck('id', reject(
```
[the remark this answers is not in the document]
xantorres: yes, you are right, it started being something and then I changed back and ended up being the same thing but in a more complex way, but I'll make it like it was before

src/app/core/AppContext.js
```diff
@@ -18,6 +12,28 @@ class AppContext extends React.Component {
...
+          unscopedToken,
+          username,
+          loginSuccessful: true
+        }
+      })
+    },
+
+    updateSession: async (path, value) => {
+      return this.state.setContext(
+        assocPath(flatten(['session', path]), value)
```
brennancheung: We can use array spread to eliminate the dependency on `flatten`.  `['session', ...path]`

src/app/plugins/developer/components/ApiHelper.js
```diff
@@ -163,6 +166,10 @@ class ApiHelper extends React.Component {
 
   renderTablePreview = () => {
     const { response, responseLens, fieldMappings } = this.state
+    const {
+      preferences: { visibleColumns, columnsOrder, rowsPerPage },
```
brennancheung: Hmm, not sure if we want `rowsPerPage` to be distinct per table or shared for everything.  I'm fine with it being distinct for now.

src/app/plugins/openstack/components/tenants/TenantChooser.js
```diff
@@ -91,5 +92,6 @@ class TenantChooser extends React.Component {
 }
 
 export default compose(
-  withAppContext,
+  withScopedPreferences('Tenants'),
```
brennancheung: I'm thinking, by convention, we should probably use the exact same key as we're using to store the associated data  in the `appContext` (`camelCase`).

src/app/plugins/openstack/components/volumes/VolumeTypesList.js
```diff
@@ -12,25 +14,31 @@ const columns = [
...
+      columns={columns}
+      data={volumeTypes}
+      onAdd={onAdd}
+      onDelete={onDelete}
+      onEdit={onEdit}
+      searchTarget="name"
+      visibleColumns={visibleColumns}
+      columnsOrder={columnsOrder}
+      rowsPerPage={rowsPerPage}
+      onRowsPerPageChange={rowsPerPage => updatePreferences({ rowsPerPage })}
```
brennancheung: Hmm, I'm wondering if we can abstract out the common props that are used in every table so we don't have to repeat this every time.  Either an HOC or a function that returns some props that we can just spread include `<ListTable {...getCommonProps()} />`

brennancheung: Or maybe just moving this functionality into the `createX` helpers.

src/app/core/AppContext.js
xantorres: no, because I wanted to be able to pass a single string as well, not just a path

src/app/plugins/openstack/components/tenants/TenantChooser.js
xantorres: yeah makes sense

src/app/plugins/openstack/components/volumes/VolumeTypesList.js
xantorres: Or just another component, `<ListTableWithPrefs>` ?

PR #246, opened 2018-12-19
Pods

kevinylee1991:
Todo:
- Need to add the unit tests
- This is a very bare-bones implementation of pods/deployments/services (the real thing does a lot of things behind the scenes that I don't know)


src/server/api/qbert/charts/index.js
```diff
@@ -1,18 +1,23 @@
+/* eslint-disable no-unused-vars */
```
brennancheung: Is this line needed?

src/server/api/qbert/namespaces/getNamespaces.js
```diff
@@ -1,8 +1,10 @@
+/* eslint-disable no-unused-vars */
```
brennancheung: Don't include these unless they are needed?  It's also better to keep linting disables only on the lines that need it.

src/server/models/qbert/Deployment.js
```diff
@@ -0,0 +1,65 @@
...
+        }
+      })
+    }
+
+    return { ...input, name: input.metadata.name, metadata: { ...input.metadata, name: input.metadata.name, namespace: input.namespace, creationTimestamp: getCurrentTime(), uuid: deploymentUuid } }
+  },
+  loaderFn: (deployments) => {
+    return deployments.map((deployment) => {
+      const newDeployment = { ...deployment, metadata: { ...deployment.metadata, uid: deployment.uuid } }
+      delete newDeployment.name
```
brennancheung: Take a look at ramda's `omit`.  https://ramdajs.com/docs/#omit

src/server/models/qbert/Pod.js
```diff
@@ -0,0 +1,55 @@
...
+      hostIp: '',
+    },
+    name: '', // For easy access, will be returned in metadata for API response
+  },
+  // Phase should be 'Pending' when first created
+  // After issuing delete it seems that phase stays as 'Running' up until termination (after 30 sec?)
+  // in metadata there is a deletionGracePeriodSeconds and a deletionTimestamp
+  createFn: (input, context) => {
+    // Remove unneeded inputs
+    if (input.apiVersion) { delete input.apiVersion }
```
brennancheung: Use ramda's `omit`.

src/server/models/qbert/Pod.js
```diff
@@ -0,0 +1,55 @@
...
+      const pod = context.pods.find(x => x.name === input.metadata.name)
+      pod.status.phase = 'Running'
+    }, 5000)
+
+    return { ...input, name: input.metadata.name, metadata: { ...input.metadata, name: input.metadata.name, namespace: input.namespace, creationTimestamp: getCurrentTime() } }
+  },
+  loaderFn: (pods) => {
+    return pods.map((pod) => {
+      const newPod = { ...pod, metadata: { ...pod.metadata, uid: pod.uuid } }
+      delete newPod.name
```
brennancheung: ramda `omit`.  Generally you never use `delete` in functional programming because it is a destructive operation (mutates original).  In this case you already clone, but it's generally good practice to make a copy with the keys removed.

PR #250, opened 2018-12-21
withStyles/webpack issue

xantorres:
Attempt to fix the issue when building the app in production

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusterUsageStats.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusterUsageStats.js
@@ -39,6 +39,8 @@ const clusterUsageStats = (cluster, context) => {
   }
 
   const { compute, memory, disk } = clusterWithStats.usage
+  // Workaround for when max is 0, but this value should never be 0
+  // and if case of so (which would mean there is no max) we should probably just show the value
   clusterWithStats.usage.compute.percent = Math.round(100 * compute.current / (compute.max || 1))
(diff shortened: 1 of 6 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusterUsageStats.js
```diff
@@ -39,9 +39,9 @@ const clusterUsageStats = (cluster, context) => {
   }
 
   const { compute, memory, disk } = clusterWithStats.usage
-  clusterWithStats.usage.compute.percent = Math.round(100 * compute.current / compute.max)
-  clusterWithStats.usage.memory.percent = Math.round(100 * memory.current / memory.max)
-  clusterWithStats.usage.disk.percent = Math.round(100 * disk.current / disk.max)
+  clusterWithStats.usage.compute.percent = Math.round(100 * compute.current / (compute.max || 1))
+  clusterWithStats.usage.memory.percent = Math.round(100 * memory.current / (memory.max || 1))
+  clusterWithStats.usage.disk.percent = Math.round(100 * disk.current / (disk.max || 1))
```
xantorres: A division by 0 was returning a NaN here

kevinylee1991: I think this is fine to avoid NaN, but I also don't think that max for a cluster should ever be 0. Maybe add a comment above with a quick mention?

In the future if there ever is a case where max can be 0, I think it usually means that there is no max (max of infinity), in which case we should adjust the display component to show only the current.

PR #251, opened 2018-12-27
Added kubernetes cloud provider details and attach/detach APIs to sim…

kevinylee1991:
…ulator

### DIFF
--- a/src/server/api/qbert/cloudProviders/getCpDetails.js
+++ b/src/server/api/qbert/cloudProviders/getCpDetails.js
@@ -16,71 +16,27 @@ const randomAwsRegions = (numRegions) => {
 }
 
 const randomOpenstackRegions = (numRegions) => {
-  const regions = times(() => {
-    return { 'RegionName': faker.random.locale() }
-  }, numRegions)
+  const regions = times(() => ({ 'RegionName': faker.random.locale() }), numRegions)
   return {'Regions': regions}
 }
 
 const randomAwsRegionDetails = (regionId) => {
   return {
-    azs: times(() => {
-      return {
-        'RegionName': regionId,
-        'State': 'available',
-        'ZoneName': faker.random.word()
-      }
-    }, 5),
-    domains: times(() => {
-      return {
-        'Name': faker.random.word(),
-        'Id': faker.random.word()
-      }
-    }, 8),
+    azs: times(() => ({ 'RegionName': regionId, 'State': 'available', 'ZoneName': faker.random.word() }), 5),
+    domains: times(() => ({ 'Name': faker.random.word(), 'Id': faker.random.word() }), 8),
     flavors: times(faker.random.word, 15),
-    keyPairs: times(() => {
-      return {
-        'KeyName': faker.random.word(),
-        'KeyFingerprint': faker.random.word()
-      }
-    }, 5),
+    keyPairs: times(() => ({ 'KeyName': faker.random.word(), 'KeyFingerprint': faker.random.word() }), 5),
     operatingSystems: ['centos', 'ubuntu'],
-    vpcs: times(() => {
-      return {
-        'CidrBlock': faker.internet.ip(),
-        'VpcName': faker.random.word()
-      }
-    }, 5)
+    vpcs: times(() => ({ 'CidrBlock': faker.internet.ip(), 'VpcName': faker.random.word() }), 5)
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/server/api/qbert/cloudProviders/getCpDetails.js
```diff
@@ -0,0 +1,151 @@
...
+      'RegionName': faker.random.locale(),
+      'Endpoint': faker.internet.url()
+    }
+  }, numRegions)
+  return {'Regions': regions}
+}
+
+const randomOpenstackRegions = (numRegions) => {
+  const regions = times(() => {
+    return { 'RegionName': faker.random.locale() }
```
brennancheung: Use the shorthand.  `times(() => ({ RegionName: faker.random.locale() }))`

src/server/api/qbert/cloudProviders/getCpDetails.js
```diff
@@ -0,0 +1,151 @@
...
+  }, numRegions)
+  return {'Regions': regions}
+}
+
+const randomAwsRegionDetails = (regionId) => {
+  return {
+    azs: times(() => {
```
brennancheung: Same here, use the shorthand everywhere you can.

src/server/api/qbert/cloudProviders/getTypes.js
```diff
@@ -0,0 +1,7 @@
+import CloudProvider from '../../../models/qbert/CloudProvider'
```
brennancheung: Don't think this file is needed at all.  It doesn't add any functionality.  It just creates an alias for a function.  You can do that with just `export default getCpTypes = CloudProvider.getCpTypes` as well.

src/server/api/qbert/clusters/attachOperations.js
```diff
@@ -0,0 +1,26 @@
...
+import Cluster from '../../../models/qbert/Cluster'
+import Node from '../../../models/qbert/Node'
+import Operations from '../../../models/qbert/Operations'
+
+export const attachNodes = (req, res) => {
+  const { clusterId } = req.params
+  const cluster = Cluster.findById({id: clusterId, context, raw: true})
+  const nodes = req.body
+  for (const node of nodes) {
+    const _node = Node.findById({id: node.uuid, context, raw: true})
```
brennancheung: `nodes.map(...findById...).forEach(attachNodeToCluster(cluster))`

src/server/models/qbert/CloudProvider.js
```diff
@@ -15,4 +15,8 @@ const options = {
 
 const CloudProvider = createModel(options)
 
+CloudProvider.getCpTypes = () => {
```
brennancheung: Use shorthand

src/server/models/qbert/Operations.js
```diff
@@ -3,3 +3,11 @@ export const attachNodeToCluster = (node, cluster) => {
   node.clusterName = cluster.name
   cluster.nodes.push(node)
 }
+
+export const detachNodeFromCluster = (node, cluster) => {
+  node.clusterUuid = null
+  node.clusterName = null
+  cluster.nodes = cluster.nodes.filter((n) => {
```
brennancheung: No need for return.  Arrow functions already return the expression body.

src/server/models/qbert/Operations.js
```diff
@@ -3,3 +3,11 @@ export const attachNodeToCluster = (node, cluster) => {
   node.clusterName = cluster.name
   cluster.nodes.push(node)
 }
+
+export const detachNodeFromCluster = (node, cluster) => {
+  node.clusterUuid = null
```
brennancheung: Kind of a red flag modifying the contents of something passed by reference but I think it will be ok in this case.

src/server/api/qbert/clusters/attachOperations.js
kevinylee1991: for this one, I want to update the original object (with the raw: true part), using map would create a new object right?

brennancheung: Correct.  For the simulator it is probably fine to mutate objects but generally it's considered bad practice with modern functional JS.

kevinylee1991: yeah I see it would make more sense to have a separate updateById function

PR #253, opened 2019-01-02
Scaffold out the AddClusterPage

brennancheung:
This is a very early start to Add Clusters.  There's still a lot of work to be done but I wanted to start early so I can split it into multiple commits.

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/AddClusterPage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/AddClusterPage.js
@@ -1,0 +1,58 @@
+import React from 'react'
+import { withDataLoader } from 'core/DataLoader'
+import FormWrapper from 'core/common/FormWrapper'
+import PicklistField from 'core/common/validated_form/PicklistField'
+import ValidatedForm from 'core/common/validated_form/ValidatedForm'
+import Wizard from 'core/common/Wizard'
+import WizardStep from 'core/common/WizardStep'
+import { compose } from 'ramda'
+import { loadCloudProviders } from './actions'
+import { withAppContext } from 'core/AppContext'
+
+class AddClusterPage extends React.Component {
+  render () {
+    // const { data } = this.props
+    // const cloudProviderOptions = data.map(cp => ({ value: cp.uuid, label: cp.name }))
+    const cloudProviderOptions = [
+      { value: '1', label: 'first cloud provider' },
+      { value: '2', label: 'another cloud provider' },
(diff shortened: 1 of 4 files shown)


src/app/core/DataLoader.js
```diff
@@ -19,25 +21,33 @@ class DataLoaderBase extends React.Component {
...
-            )
+          const promises = loaderFns.map(fn => fn({ setContext, context }))
+          const [promiseFns, syncFns] = partition(propExists('then'), promises) // eslint-disable-line no-unused-vars
+          if (isEmpty(promiseFns)) {
+            // everything is sync so no need to wait
+            this.setState({ loading: false })
           }
-          // loaderFn is sync so loading is done
-          this.setState({ loading: false })
+          Promise.all(promises).then(
```
kevinylee1991: would this part be redundant if promiseFns is empty?

src/app/plugins/kubernetes/components/infrastructure/AddClusterPage.js
```diff
@@ -0,0 +1,105 @@
...
+  manualDeploy: false,
+}
+
+class AddClusterPage extends React.Component {
+  handleSubmit = () => console.log('TODO: AddClusterPage#handleSubmit')
+
+  render () {
+    const { data } = this.props
+    const cloudProviderOptions = projectAs({ value: 'uuid', label: 'name' }, data.cloudProviders)
+    const regionOptions = projectAs({ value: 'id', label: 'name' }, data.regions)
```
kevinylee1991: list of regions will come from the cloud providers API, wouldn't this need to start as empty array?

src/app/plugins/kubernetes/components/infrastructure/AddClusterPage.js
```diff
@@ -0,0 +1,105 @@
...
+    const flavorOptions = projectAs({ value: 'id', label: 'name' }, data.flavors)
```
kevinylee1991: same with the empty array thing

src/app/core/DataLoader.js
brennancheung: Ah, yes.  Good call.

src/app/plugins/kubernetes/components/infrastructure/AddClusterPage.js
brennancheung: I wasn't aware of this.  I'll change how these data are retrieved in the next PR.

PR #256, opened 2019-01-10
Pods/deployments/services UI

kevinylee1991:
Roopak wants me to work on something in the short term for old UI, so I'm putting what I have up now in case you want to take a look through


src/api-client/Qbert.js
```diff
@@ -145,19 +145,37 @@ class Qbert {
   }
 
   async getClusterPods (clusterId) {
-    return this.client.basicGet(`${await this.baseUrl()}/clusters/${clusterId}/k8sapi/api/v1/pods`)
+    try {
+      const data = await this.client.basicGet(`${await this.baseUrl()}/clusters/${clusterId}/k8sapi/api/v1/pods`)
+      return data.items.map(this.convertCluster(clusterId))
+    } catch (err) {
+      console.log(`Error getting cluster pods for clusterId: ${clusterId}`)
+      return []
```
brennancheung: Not sure we should be silencing the errors and returning an empty array.  This is going to lead to us not knowing there is an issue.

Use `console.error` for errors.

src/api-client/Qbert.js
```diff
@@ -145,19 +145,37 @@ class Qbert {
...
   }
 
   async getClusterDeployments (clusterId) {
-    return this.client.basicGet(`${await this.baseUrl()}/clusters/${clusterId}/k8sapi/apis/extensions/v1beta1/deployments`)
+    try {
+      const data = await this.client.basicGet(`${await this.baseUrl()}/clusters/${clusterId}/k8sapi/apis/extensions/v1beta1/deployments`)
+      return data.items.map(this.convertCluster(clusterId))
+    } catch (err) {
+      console.log(`Error getting cluster deployments for clusterId: ${clusterId}`)
+      return []
```
brennancheung: same

src/api-client/Qbert.js
```diff
@@ -145,19 +145,37 @@ class Qbert {
...
   }
 
   async getClusterKubeServices (clusterId) {
-    return this.client.basicGet(`${await this.baseUrl()}/clusters/${clusterId}/k8sapi/api/v1/services`)
+    try {
+      const data = await this.client.basicGet(`${await this.baseUrl()}/clusters/${clusterId}/k8sapi/api/v1/services`)
+      return data.items.map(this.convertCluster(clusterId))
+    } catch (err) {
+      console.log(`Error getting cluster services for clusterId: ${clusterId}`)
+      return []
```
brennancheung: same

src/app/app.css
```diff
@@ -32,3 +32,7 @@ body {
 #sidebar {
   grid-area: sd;
 }
+
+.CodeMirror {
+  border: 1px solid black;
```
brennancheung: Are we able to move this into the component instead of the external css?

src/app/plugins/kubernetes/components/pods/AddPodPage.js
```diff
@@ -0,0 +1,87 @@
...
+import CodeMirror from 'core/common/validated_form/CodeMirror'
+
+export class AddPodForm extends React.Component {
+  state = {
+    clusterId: null,
+    clusterOptions: [],
+    namespaceOptions: [],
+  }
+
+  async componentDidMount () {
```
brennancheung: Use `DataLoader` for initial loading kind of functionality.

src/app/plugins/kubernetes/components/pods/AddPodPage.js
```diff
@@ -0,0 +1,87 @@
...
+    const { context, setContext } = this.props
+    await loadInfrastructure({ context, setContext })
+    // Make sure to use the new reference to context.  It changes after loadInfrastucture.
+    const clusterOptions = this.props.context.clusters.map(c => ({ value: c.uuid, label: c.name }))
+    this.setState({ clusterOptions })
+  }
+
+  setField = key => value => {
```
brennancheung: No need for this, just let `ValidatedForm` handle it for you automatically.

src/app/plugins/kubernetes/components/pods/AddPodPage.js
```diff
@@ -0,0 +1,87 @@
...
+import PicklistField from 'core/common/validated_form/PicklistField'
+import SubmitButton from 'core/common/SubmitButton'
+import TextField from 'core/common/validated_form/TextField'
+import createAddComponents from 'core/helpers/createAddComponents'
+import { loadInfrastructure } from '../infrastructure/actions'
+import { loadPods, createPod } from './actions'
+import CodeMirror from 'core/common/validated_form/CodeMirror'
+
+export class AddPodForm extends React.Component {
+  state = {
```
brennancheung: `ValidatedForm` can maintain the state for you.  No need for this.

src/app/plugins/kubernetes/components/pods/AddPodPage.js
```diff
@@ -0,0 +1,87 @@
...
+      this.setState({ namespaceOptions })
+    }
+  }
+
+  render () {
+    const { clusterOptions, namespaceOptions } = this.state
+    if (!clusterOptions) { return null }
+    const { onComplete } = this.props
+    const codeMirrorOptions = {
+      lineNumbers: true,
```
brennancheung: Move these into props on the `CodeMirror` component.  Assume reasonable defaults where possible to reduce the number of props we need to pass to use this component.  Probably, `mode: 'yaml'` is the only one from what I can tell that would deviate from reasonable defaults.

src/app/plugins/kubernetes/components/pods/AddPodPage.js
```diff
@@ -0,0 +1,87 @@
...
+        }
+      }
+    }
+
+    return (
+      <ValidatedForm onSubmit={onComplete}>
+        <TextField id="name" label="Name" />
+        <PicklistField id="clusterId"
+          label="Cluster"
+          onChange={this.setField('clusterId')}
```
brennancheung: No need for `onChange`, let `ValidatedForm` handle it for you.

src/app/plugins/kubernetes/components/pods/DeploymentsListPage.js
```diff
@@ -0,0 +1,93 @@
...
+class ListPage extends React.Component {
+  state = {
+    activeCluster: '__all__',
+    deployments: null,
+    clusterOptions: [
+      { label: 'all', value: '__all__' },
+    ],
+  }
+
+  async componentDidMount () {
```
brennancheung: Same.  Use `DataLoader`

src/app/plugins/kubernetes/components/pods/DeploymentsListPage.js
```diff
@@ -0,0 +1,93 @@
...
+    const { context, setContext } = this.props
+    await loadInfrastructure({ context, setContext })
+
+    // Make sure to use a new reference to props.context since it has now changed
+    const clusters = this.props.context.clusters.filter(x => x.hasMasterNode)
+    // Need to query for all clusters
+    await loadDeployments({ clusterId: clusters[0].uuid, context, setContext })
+    const clusterOptions = clusters.map(cluster => ({
```
brennancheung: Take a look at the utility function called `projectAs` in `core/fp.js` you can use for this purpose.

See https://github.com/platform9/pf9-ui-plugin/blob/master/src/app/plugins/kubernetes/components/infrastructure/AddClusterPage.js#L25 for an example of using `DataLoader`, `projectAs` and generally a cleaner pattern.

src/app/plugins/kubernetes/components/pods/DeploymentsListPage.js
```diff
@@ -0,0 +1,93 @@
...
+  findClusterName = clusterId => {
+    const cluster = this.props.context.clusters.find(x => x.uuid === clusterId)
+    return (cluster && cluster.name) || ''
+  }
+
+  render () {
+    const { activeCluster, clusterOptions } = this.state
+    const { deployments } = this.props.context
+    if (!deployments) { return <Loader /> }
+    const filteredDeployments = (activeCluster === '__all__' && deployments) ||
```
brennancheung: Use ternary operator here.

Optional, ramda has functions for common scenarios like these:

`deployments.filter(propEq('clusterId', activeCluster))`

Also take a look at ramda's `assoc` function for adding props to an object.

src/app/plugins/kubernetes/components/pods/DeploymentsListPage.js
```diff
@@ -0,0 +1,93 @@
...
+
+  findClusterName = clusterId => {
+    const cluster = this.props.context.clusters.find(x => x.uuid === clusterId)
+    return (cluster && cluster.name) || ''
+  }
+
+  render () {
+    const { activeCluster, clusterOptions } = this.state
+    const { deployments } = this.props.context
+    if (!deployments) { return <Loader /> }
```
brennancheung: No need to check if we need to display `<Loader />`.  `DataLoader` takes care of this for us.

src/app/plugins/kubernetes/components/pods/DeploymentsListPage.js
```diff
@@ -0,0 +1,93 @@
...
+    { id: 'clusterName', label: 'Cluster' },
+    { id: 'created', label: 'Created' },
+  ],
+  dataKey: 'deployments',
+  name: 'Deployments',
+  title: 'Deployments',
+  ListPage: DeploymentsListPage,
+}
+const components = createCRUDComponents(options)
+ListContainer = components.ListContainer
```
brennancheung: Hmm, this doesn't seem right.  Is this because the table contents change based on the cluster?  We might want to see if we can add an abstraction that is a little cleaner, or just implement the functionality outside of `createCRUDComponents`.  That utility is really for the most common scenarios that are repetitive, if a component deviates from it we might want to drop down to a lower level.

How common do you think it is to do something like this?  I can see passing in some kind of wrapping container and filtering functions to `createCRUDComponents` as another possibility.

src/app/plugins/kubernetes/components/pods/PodsListPage.js
```diff
@@ -0,0 +1,94 @@
+import React from 'react'
+import Picklist from 'core/common/Picklist'
+import createCRUDComponents from 'core/helpers/createCRUDComponents'
+import Loader from 'core/common/Loader'
+import { withAppContext } from 'core/AppContext'
+import { loadInfrastructure } from '../infrastructure/actions'
+import { loadPods, deletePod } from './actions'
+
+let ListContainer = null
```
brennancheung: Ok, looks like we need to think about an abstraction for this scenario.

src/app/plugins/kubernetes/components/pods/ServicesListPage.js
```diff
@@ -0,0 +1,94 @@
+import React from 'react'
+import Picklist from 'core/common/Picklist'
+import createCRUDComponents from 'core/helpers/createCRUDComponents'
+import Loader from 'core/common/Loader'
+import { withAppContext } from 'core/AppContext'
+import { loadInfrastructure } from '../infrastructure/actions'
+import { loadServices, deleteService } from './actions'
+
+let ListContainer = null
```
brennancheung: Same

src/app/plugins/kubernetes/components/pods/actions.js
```diff
@@ -0,0 +1,71 @@
...
+    clusterId,
+    name: created.metadata.name,
+    created: created.metadata.creationTimestamp,
+    id: created.metadata.uid
+  }
+  setContext({ pods: [ ...existing, converted ] })
+  return created
+}
+
+export const deletePod = async ({ id, context, setContext }) => {
```
brennancheung: These actions look very simple.  Take a look at `createCRUDActions`.  They allow you to create these functions cleaner.  Also, you will need to add a reference in the api client.  See `qbert.js` in the api client for how to do this.  Don't forget the `bind`.

src/app/plugins/openstack/components/floatingips/actions.js
```diff
@@ -6,8 +6,8 @@ export const loadFloatingIps = async ({ context, setContext, reload }) => {
 }
 
 export const createFloatingIp = async ({ data, context, setContext }) => {
-  const created = await context.apiClient.neutron.createFloatingIp(data)
   const existing = await context.apiClient.neutron.getFloatingIps()
+  const created = await context.apiClient.neutron.createFloatingIp(data)
```
brennancheung: I think we can refactor a bunch of these to use `createCRUDActions`

webpack.config.js
```diff
@@ -63,7 +63,7 @@ module.exports = {
       },
       {
         test: /\.css$/,
-        exclude: /node_modules/,
+        // exclude: /node_modules/,
```
brennancheung: I ran into this problem as well from what I remember I copied the CSS files out so we can include them manually.

Then just do an `import './codemirror.css'` in `App.js`.

src/app/plugins/kubernetes/components/pods/PodsListPage.js
kevinylee1991: this is regarding this right?

await loadPods({ clusterId: clusters[0].uuid, context, setContext })

src/app/plugins/kubernetes/components/pods/AddPodPage.js
kevinylee1991: Having some difficulties with this one -- i may be trying to use loaderFn incorrectly

I tried creating a new function that calls loadInfrastructure and loadPods, as well as setting clusterOptions but the loaderFn didn't get called.

src/app/plugins/kubernetes/components/pods/DeploymentsListPage.js
kevinylee1991: I'm actually not sure, I was following the namespaces list page as an example -- will ask you about this next week

src/app/plugins/kubernetes/components/pods/AddPodPage.js
brennancheung: Make sure you have latest code.  I fixed some issues with DataLoader.

src/app/plugins/kubernetes/components/pods/PodsListPage.js
brennancheung: Not specifically about any line, it's more the pattern and how it doesn't fit this scenario so we need another pattern or to drop down to a lower level.  `createCRUDComponents` is a high level abstraction but the use case here doesn't quite fit.  The `let ListContainer = null` is a clear indication of this.  It's kind of hacky and abusing things to make it fit.

src/app/plugins/kubernetes/components/pods/actions.js
kevinylee1991: Did this for load/list, but create and delete will need extra changes to pass the clusterId / namespace along with the request body. Since that change would require changing every instance of createCRUDActions and the api client, I thought this change would be better in its own commit.

kevinylee1991:
Looking at the ListPages now to see how they should be changes, but I fixed up everything else that had comments on it

src/app/core/common/validated_form/CodeMirror.js
```diff
@@ -0,0 +1,74 @@
...
+
+require('codemirror/mode/yaml/yaml')
+
+const styles = theme => ({
+  formControl: {
+    margin: theme.spacing.unit
+  }
+})
+
+const defaultOptions = {
```
brennancheung: Expose these as props.

src/app/core/common/validated_form/codemirror.css
```diff
@@ -0,0 +1,347 @@
+/* BASICS */
```
brennancheung: Add a comment about where this was taken from and why.

src/app/core/helpers/createCRUDActions.js
```diff
@@ -14,9 +14,9 @@ const createCRUDActions = (options={}) => {
       return created
     },
 
-    list: async ({ context, setContext, reload }) => {
+    list: async ({ clusterId, context, setContext, reload }) => {
```
brennancheung: `clusterId` is too specific for a generic function.  But I'm fine with something called `params` and then passing that in.

src/app/core/helpers/createCRUDComponents.js
```diff
@@ -37,6 +37,7 @@ const createCRUDComponents = options => {
 
   const {
     actions,
+    baseCrudUrl,
```
brennancheung: Why can't the consumer of this function just pass a different `baseUrl`?

src/app/plugins/kubernetes/components/pods/AddDeploymentPage.js
```diff
@@ -0,0 +1,71 @@
...
+    const namespaceOptions = context.namespaces.filter(n => n.clusterId === value).map(n => ({ value: n.name, label: n.name }))
+    this.setState({ namespaceOptions })
+  }
+
+  render () {
+    const { namespaceOptions } = this.state
+    const { context, onComplete } = this.props
+
+    const codeMirrorOptions = {
+      mode: 'yaml',
```
brennancheung: I'm wondering if we should make this a top level prop on the `<CodeMirror>` component.

src/app/plugins/kubernetes/components/pods/AddPodPage.js
```diff
@@ -0,0 +1,71 @@
...
+    const { context, onComplete } = this.props
+
+    const codeMirrorOptions = {
+      mode: 'yaml',
+    }
+
+    const clusterOptions = context.clusters ? projectAs({ value: 'uuid', label: 'name' }, context.clusters) : []
+
+    return (
+      <DataLoader dataKey="clusters" loaderFn={loadInfrastructure}>
```
brennancheung: Using the `withDataLoader` helper will allow you to clean this up and get rid of the "render prop" overhead.

src/app/plugins/kubernetes/components/pods/AddDeploymentPage.js
kevinylee1991: We could do that, but we would lose the ability to easily combine the 2 options objects later so I'm not sure if it would be worth

src/app/core/helpers/createCRUDComponents.js
kevinylee1991: We use tabs for the pods/deployments/services with rely on url hashes while the baseUrl remains as /pods, but these hashes don't work when in the middle of the url. This component just adds /add or /edit to the end of the baseUrl for the add and edit links.

I don't see separating out the hash into its own prop working either because the pods/deployments/services pages will still all share the same baseUrl and will produce the same add and edit URLs.

PR #257, opened 2019-01-11
Refactoring organization

xantorres:


src/app/core/components/listTable/ListTable.js
```diff
@@ -1,19 +1,36 @@
...
-import ListTableHead from './ListTableHead'
-import ListTableToolbar from './ListTableToolbar'
-import MoreMenu from 'core/common/MoreMenu'
-import { compose, ensureFunction, except } from 'core/fp'
+import { withStyles } from '@material-ui/core/styles'
+import { compose, ensureFunction, except } from 'app/utils/fp'
 import { withAppContext } from 'core/AppContext'
+import MoreMenu from 'core/components/MoreMenu'
+import PropTypes from 'prop-types'
 import {
```
brennancheung: Try to keep these on fewer lines if it can fit on one line.

src/app/App.js
```diff
@@ -1,23 +1,26 @@
-import React from 'react'
+import { MuiThemeProvider } from '@material-ui/core/styles'
```
brennancheung: Let's keep imports in their original order, or sorted.  The general strategy we have been using is:

`import React`
`import PropTypes`
sorted imports

Keep multiple imports on 1 line when they can be.

src/app/core/components/Navbar.js
```diff
@@ -1,22 +1,32 @@
-import React, { PureComponent } from 'react'
-import { matchPath, withRouter } from 'react-router'
-import { assoc, flatten, pluck, prop, propEq, propOr } from 'ramda'
-import moize from 'moize'
-import PropTypes from 'prop-types'
 import {
-  Collapse, Divider, Drawer, ExpansionPanel, ExpansionPanelDetails,
-  ExpansionPanelSummary, IconButton, InputBase, ListItemIcon, ListItemText,
-  MenuItem, MenuList, Typography
+  Collapse,
```
brennancheung: I like the old style here.  This takes up too much space.

src/app/App.js
brennancheung: If something needs to be multiple lines I usually save those for the end of the imports.

PR #262, opened 2019-01-21
start of form validation for pods page

kevinylee1991:
Had some difficulties adding form validation to PickListField, I was able to get it working by adding the validations prop to the withFormContext component, but I'm not sure if that was the right thing to do or not.

### DIFF
--- a/src/app/core/components/validatedForm/PicklistField.js
+++ b/src/app/core/components/validatedForm/PicklistField.js
@@ -1,7 +1,20 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import Picklist from 'core/components/Picklist'
-import withFormContext from 'core/components/validatedForm/withFormContext'
+import {
+  FormControl,
+  FormHelperText,
+} from '@material-ui/core'
+import { withStyles } from '@material-ui/core/styles'
+import { withInfoTooltip } from 'app/core/components/InfoTooltip'
+import { compose } from 'app/utils/fp'
+import withFormContext, { ValidatedFormInputPropTypes } from 'core/components/validatedForm/withFormContext'
+
+const styles = theme => ({
+  formControl: {
+    margin: theme.spacing.unit
+  }
+})
 
 /**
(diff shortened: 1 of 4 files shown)


src/app/plugins/kubernetes/components/pods/AddPodPage.js
```diff
@@ -38,15 +38,18 @@ export class AddPodForm extends React.Component {
           label="Cluster"
           onChange={this.handleClusterChange}
           options={clusterOptions}
+          validations={{required: true}}
```
xantorres: prop `required` is enough

src/app/core/components/validatedForm/withFormContext.js
```diff
@@ -113,6 +111,7 @@ const withFormContext = Input =>
               onChange={onChange}
               onMouseEnter={onMouseEnter}
               onBlur={onBlur}
+              validations={validations}
```
xantorres: you don't need to pass this

src/app/core/components/validatedForm/PicklistField.js
```diff
@@ -16,19 +29,20 @@ class PicklistField extends React.Component {
...
           {...restProps}
           name={id}
           label={label}
           options={options}
-          value={value !== undefined ? value : ''}
+          value={value !== undefined ? value: ''}
           onChange={this.handleChange}
         />
-      </div>
+        <FormHelperText>{errorMessage}</FormHelperText>
```
xantorres: Yes, I missed that here, good catch

src/app/core/components/validatedForm/withFormContext.js
kevinylee1991: Somehow if I don't pass this, it gets stripped from the PickListField, any ideas why this would happen?

src/app/plugins/kubernetes/components/pods/AddPodPage.js
kevinylee1991: I tried this but wasn't able to get it working

PR #269, opened 2019-02-01
Cluster actions (attach)

brennancheung:
This adds support for attaching nodes to a cluster.

There are also some misc bug fixes.

There is now an `/admin` API in the simulator to allow the server state to be controlled through an API.  This is used in the Cypress tests to allow state to be configured during test runs.  This is especially important in watch mode.

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/ClusterAttachNodeDialog.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/ClusterAttachNodeDialog.js
@@ -14,11 +14,14 @@ import {
   TableBody,
   TableRow,
   TableCell,
+  Typography,
 } from '@material-ui/core'
 
 // The modal is technically inside the row, so clicking anything inside
 // the modal window will cause the table row to be toggled.
 const stopPropagation = e => {
+  // Except for <a href=""> style links
+  if (e.target.tagName.toUpperCase() === 'A') { return }
   e.preventDefault()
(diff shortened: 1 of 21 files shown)


src/app/plugins/kubernetes/components/infrastructure/ClusterAttachNodeDialog.js
```diff
@@ -0,0 +1,113 @@
...
+    const { data } = this.props
+    const freeNodes = data.filter(x => !x.clusterUuid)
+    return (
+      <Dialog open onClose={this.handleClose} onClick={stopPropagation}>
+        <DialogTitle>Attach Node to Cluster</DialogTitle>
+        <DialogContent>
+          <p>
+            <b>IMPORTANT</b>:
+            Before adding nodes to a cluster, please ensure that you have followed the requirements
+            in <a href="https://platform9.com/support/managed-container-cloud-requirements-checklist/" target="_blank">this article</a> for
```
kevinylee1991: not sure how useful this really is for us, but at some point in the old UI blake had asked me to add rel="noopener" for every single outbound link, should we add it in the new UI too you think?

brennancheung: I have the `target="_blank"` for this reason but I'll create a separate PR where we refactor support links into its own simple component.

PR #273, opened 2019-02-11
App catalog page, CardTable component and some refactoring

xantorres:
Checklist:
- [x] Fix api calls for Charts
- [x] Basic App Catalog page
- [x] Improve CardTable component with pagination, filtering and sorting

image


src/app/core/components/appCatalog/ApplicationCard.js
```diff
@@ -0,0 +1,164 @@
...
+    padding: theme.spacing.unit
+  }
+})
+
+class ApplicationCard extends React.Component {
+  handleAddToEnv = () => {
+    this.props.handleAddToEnv()
+  }
+
+  handleDeploy = () => {
```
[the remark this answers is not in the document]
brennancheung: Also make sure these are required in the `propTypes`.

src/app/plugins/kubernetes/components/apps/actions.js
```diff
@@ -0,0 +1,44 @@
+import { flatten, prop } from 'ramda'
+
+export const loadClusters = async ({ data, context, setContext, reload }) => {
```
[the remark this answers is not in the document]
xantorres: actually, I'm not using this function anymore, so I will remove it

src/app/core/components/appCatalog/ApplicationCard.js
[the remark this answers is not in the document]
xantorres: yes, adding some logic soon

src/server/presets/dev.js
```diff
@@ -124,37 +126,41 @@ function loadPreset () {
...
-
-  // Nodes
-  // Nodes must be linked to a resMgrHost id or else the UI will break
-  const node = Node.create({
-    data: { name: 'fakeNode1', api_responding: 1, isMaster: 1, uuid: resMgrHost.id },
-    context,
-    raw: true
-  })
-  Node.create({ data: { name: 'fakeNode2', uuid: resMgrHost2.id }, context })
+  range(1, 4).forEach(clusterIdx => {
```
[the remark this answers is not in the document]
brennancheung: Slightly unrelated.. I'm thinking we may want to split up the presets.  I'm thinking of having the Cypress tests drive the simulator to create different configs for different scenarios so having functions that we can delegate to to create different configs would work well.  Might also work well while we are developing too.

src/app/core/components/cardTable/CardTable.js
```diff
@@ -0,0 +1,231 @@
...
+      orderBy: null,
+      direction: 'asc',
+      page: 0,
+      selected: [],
+      searchTerm: '',
+      filterValues: {}
+    }
+  }
+
+  handleSearch = value => {
```
brennancheung: Optional if you like it better.  I know the JS community in general hates "abusing" the short-circuit operations but I find it much easier as a 1-liner than 6 lines.

`handleSearch = searchTerm => this.props.searchTarget && this.setState({ searchTerm })`

src/app/core/components/cardTable/CardTable.js
```diff
@@ -0,0 +1,231 @@
...
+      'sortWith',
+      sorting.find(propEq('id', orderBy))
+    )
+    const sortedRows = [...data].sort((a, b) =>
+      sortWith(b[orderBy], a[orderBy])
+    )
+    return this.state.direction === 'desc' ? sortedRows : sortedRows.reverse()
+  }
+
+  paginate = data => {
```
brennancheung: We probably want an HOC for pagination but we can do it in another PR.

src/app/plugins/kubernetes/components/apps/AppCatalogPage.js
```diff
@@ -0,0 +1,90 @@
...
+    },
+    {
+      field: 'relationships.latestChartVersion.data.created',
+      label: 'Created',
+      sortWith: (prevDate, nextDate) =>
+        moment(prevDate).isBefore(nextDate) ? 1 : -1
+    }
+  ]
+
+  get filtersConfig () {
```
brennancheung: Could be much shorter without the getter.

`getFiltersConfig = () => [{ ... }]`

src/app/plugins/openstack/index.js
```diff
@@ -1,57 +1,56 @@
...
 
+import ApiAccessPage from './components/api-access/ApiAccessListPage'
+import AppCatalogPage from 'k8s/components/apps/AppCatalogPage'
+
 import DashboardPage from './components/DashboardPage'
+import AddFlavorPage from './components/flavors/AddFlavorPage'
 
-import AddTenantPage from './components/tenants/AddTenantPage'
-import TenantsListPage from './components/tenants/TenantsListPage'
+import FlavorsListPage from './components/flavors/FlavorsListPage'
```
brennancheung: Let's keep the imports organized by topic.

src/server/presets/dev.js
```diff
@@ -124,37 +127,41 @@ function loadPreset () {
...
-  // Nodes
-  // Nodes must be linked to a resMgrHost id or else the UI will break
-  const node = Node.create({
-    data: { name: 'fakeNode1', api_responding: 1, isMaster: 1, uuid: resMgrHost.id },
-    context,
-    raw: true
-  })
-  Node.create({ data: { name: 'fakeNode2', uuid: resMgrHost2.id }, context })
+  range(4).forEach(clusterIdx => {
+    const cluster = Cluster.create({ data: { name: `fakeCluster${clusterIdx}`, sshKey: 'someKey' }, context, raw: true })
```
brennancheung: Please verify the Cypress integration tests for `clusters.spec.js` still work.  We might need to start splitting the presets into smaller chunks and loading multiple of them.

src/stories/common/CardTable.stories.js
```diff
@@ -0,0 +1,256 @@
...
+    return this.props.children({
+      filtersConfig,
+      sortingConfig,
+      data: filteredData,
+      classes
+    })
+  }
+}
+
+const DefaultListTable = ({ actions }) => (
```
brennancheung: Should this be its own component outside of stories?

src/stories/common/CardTable.stories.js
```diff
@@ -0,0 +1,256 @@
...
+  actions: {
+    display: 'flex',
+    justifyContent: 'flex-end',
+    alignItems: 'center',
+    padding: theme.spacing.unit
+  }
+})
+
+@withStyles(styles)
+class CardTableContainer extends React.Component {
```
brennancheung: Should this be its own component outside of stories?

src/app/plugins/kubernetes/components/apps/AppCatalogPage.js
xantorres: this needs to be called every time as it uses `this.props.context.clusters` which could have changed its reference

src/stories/common/CardTable.stories.js
xantorres: We will see in future implementations if its worth relying on a higher level component like this one that renders the most common implementation of the CardTable and CardItems.

PR #275, opened 2019-02-16
Detach nodes from clusters

brennancheung:
This allows nodes to be detached from clusters through the row action menu.


brennancheung:
What do you guys think about using Material-UI's `Switch` instead of checkboxes for scenarios where we have to toggle rows of items on/off?

https://material-ui.com/demos/selection-controls/#switches

kevinylee1991:
We actually do use switch some certain places in the old UI. I think it could make sense to standardize usage with switch

PR #276, opened 2019-02-21
Scale clusters

brennancheung:


src/app/plugins/kubernetes/components/infrastructure/ClusterScaleDialog.js
```diff
@@ -0,0 +1,76 @@
...
+
+  render () {
+    // Disabling spot worker feature for now.  It sounds like it is going to change
+    // in the near future.
+    const spotFeatureEnabled = false
+
+    const { enableSpotWorkers, sliderValue } = this.state
+    const { row } = this.props
+    const initialValues = pick(['numMasters', 'numWorkers'], row)
+    console.log(row)
```
kevinylee1991: console log

src/app/core/components/validatedForm/ValidatedForm.js
```diff
@@ -155,9 +155,9 @@ class ValidatedForm extends React.Component {
   }
 
   render () {
-    const { classes } = this.props
+    const { classes, id } = this.props
```
kevinylee1991: just curious what purposes does the id serve?

brennancheung: Material-UI's `Dialog` does weird things with forms.  It was breaking the `<button type="submit">` functionality.  Turns out there is a way to use `<form id="abc">` and `<button type="submit" form="abc">` to associate them.

PR #278, opened 2019-02-22
Added kubeconfig download link to clusters list

kevinylee1991:
Had a few points I wanted to run by you guys:

1. We need some way to tell users that the kubeconfig will only work for 24 hours. Ideally this should go through a designer.
2. Wasn't sure if it'd be worth it to create a util function for creating a download. Currently this is our only use for this kind of functionality, although I could certainly see nice-to-have features that could use it (like a csv download of a list table for example)

### DIFF
--- a/src/api-client/Keystone.js
+++ b/src/api-client/Keystone.js
@@ -140,6 +140,21 @@ class Keystone {
     }
   }
 
+  async renewScopedToken () {
+    const body = constructAuthBody('token', this.client.unscopedToken)
+    const projectId = this.client.activeProjectId
+    body.auth.scope = { project: { id: projectId } }
+    try {
+      const response = await axios.post(this.tokensUrl, body)
+      const scopedToken = response.headers['x-subject-token']
+      this.client.scopedToken = scopedToken
+      return scopedToken
+    } catch (err) {
+      // authentication failed
+      return null
+    }
+  }
+
   async getRegions () {
     const response = await axios.get(this.regionsUrl, this.client.getAuthHeaders())
     return response.data.regions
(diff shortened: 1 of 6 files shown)


src/api-client/Keystone.js
```diff
@@ -140,6 +140,21 @@ class Keystone {
...
+    const body = constructAuthBody('token', this.client.unscopedToken)
+    const projectId = this.client.activeProjectId
+    body.auth.scope = { project: { id: projectId } }
+    try {
+      const response = await axios.post(this.tokensUrl, body)
+      const scopedToken = response.headers['x-subject-token']
+      this.client.scopedToken = scopedToken
+      return scopedToken
+    } catch (err) {
+      // authentication failed
```
brennancheung: Add a `console.error` to make it easier to figure out should this ever fail in the future.

brennancheung:
1) For now we can put some text near the download link that mentions it's only valid for 24 hours.
2) We can refactor out common functionality once we see it used more but feel free to define a function in one of the util / helper files if you want.

src/server/api/qbert/kubeConfig/getKubeConfig.js
```diff
@@ -0,0 +1,35 @@
...
+  const { clusterId } = req.params
+  const cluster = Cluster.findById({ id: clusterId, context })
+
+  if (!cluster) {
+    return res.status(404).send({code: 404, message: 'cluster not found'})
+  }
+
+  const yamlString = `apiVersion: v1
+clusters:
+- cluster:
```
brennancheung: I think it is fine in this case since it is time limited and a private IP but be careful of making mock tokens and certs from actual responses.  Jumble and randomize the characters to make it invalid in case anyone tries to use it directly.

PR #280, opened 2019-02-25
Fixed issue where pods/deployments/services errored out when no clust…

kevinylee1991:
…ers existed

### DIFF
--- a/src/app/plugins/kubernetes/components/pods/DeploymentsListPage.js
+++ b/src/app/plugins/kubernetes/components/pods/DeploymentsListPage.js
@@ -23,7 +23,11 @@ const ListPage = ({ ListContainer }) => {
       // Make sure to use a new reference to props.context since it has now changed
       const clusters = this.props.context.clusters.filter(x => x.hasMasterNode)
       // Need to query for all clusters
-      await loadDeployments({ params: { clusterId: clusters[0].uuid }, context, setContext })
+      if (clusters.length) {
+        await loadDeployments({ params: { clusterId: clusters[0].uuid }, context, setContext })
+      } else {
+        setContext({ deployments: [] })
+      }
       const clusterOptions = clusters.map(cluster => ({
         label: cluster.name,
         value: cluster.uuid
--- a/src/app/plugins/kubernetes/components/pods/PodsListPage.js
+++ b/src/app/plugins/kubernetes/components/pods/PodsListPage.js
@@ -23,7 +23,11 @@ const ListPage = ({ ListContainer }) => {
       // Make sure to use a new reference to props.context since it has now changed
       const clusters = this.props.context.clusters.filter(x => x.hasMasterNode)
       // Need to query for all clusters
-      await loadPods({ params: { clusterId: clusters[0].uuid }, context, setContext })
+      if (clusters.length) {
+        await loadPods({ params: { clusterId: clusters[0].uuid }, context, setContext })
+      } else {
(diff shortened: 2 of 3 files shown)


src/app/plugins/kubernetes/components/pods/DeploymentsListPage.js
```diff
@@ -23,7 +23,11 @@ const ListPage = ({ ListContainer }) => {
       // Make sure to use a new reference to props.context since it has now changed
       const clusters = this.props.context.clusters.filter(x => x.hasMasterNode)
       // Need to query for all clusters
-      await loadDeployments({ params: { clusterId: clusters[0].uuid }, context, setContext })
+      if (clusters.length) {
+        await loadDeployments({ params: { clusterId: clusters[0].uuid }, context, setContext })
+      } else {
+        setContext({ deployments: [] })
```
brennancheung: I think how you're using `setContext` will break `DataLoader`.  It will think the data already has been loaded and not attempt to fetch it.

src/app/plugins/kubernetes/components/pods/DeploymentsListPage.js
```diff
@@ -23,7 +23,11 @@ const ListPage = ({ ListContainer }) => {
       // Make sure to use a new reference to props.context since it has now changed
       const clusters = this.props.context.clusters.filter(x => x.hasMasterNode)
       // Need to query for all clusters
-      await loadDeployments({ params: { clusterId: clusters[0].uuid }, context, setContext })
+      if (clusters.length) {
+        await loadDeployments({ params: { clusterId: clusters[0].uuid }, context, setContext })
```
brennancheung: `loadDeployments` and `loadInfrastructure` should happen in the `DataLoader` part.  I think you can control that through the options to `createCrudComponents`.  You should be able to get rid of the `componentDidMount` completely.

The `clusterOptions` can be computed in the `render` function.

src/app/plugins/kubernetes/components/pods/PodsListPage.js
```diff
@@ -23,7 +23,11 @@ const ListPage = ({ ListContainer }) => {
       // Make sure to use a new reference to props.context since it has now changed
       const clusters = this.props.context.clusters.filter(x => x.hasMasterNode)
       // Need to query for all clusters
-      await loadPods({ params: { clusterId: clusters[0].uuid }, context, setContext })
+      if (clusters.length) {
+        await loadPods({ params: { clusterId: clusters[0].uuid }, context, setContext })
```
brennancheung: Same.  Move loading pods into the data loader.

src/app/plugins/kubernetes/components/pods/ServicesListPage.js
```diff
@@ -23,7 +23,11 @@ const ListPage = ({ ListContainer }) => {
       // Make sure to use a new reference to props.context since it has now changed
       const clusters = this.props.context.clusters.filter(x => x.hasMasterNode)
       // Need to query for all clusters
-      await loadServices({ params: { clusterId: clusters[0].uuid }, context, setContext })
+      if (clusters.length) {
+        await loadServices({ params: { clusterId: clusters[0].uuid }, context, setContext })
```
brennancheung: Same

brennancheung:
Take a look at the new `DataLoader` @xantorres created.  It has the ability to load multiple entities and have dependencies between them.

PR #282, opened 2019-02-26
Feature/catalog repositories

xantorres:


src/app/plugins/kubernetes/components/apps/actions.js
```diff
@@ -59,6 +59,28 @@ export const deleteRelease = async ({
...
 }
 
+export const loadRepositories = async ({
+  data,
+  context,
+  setContext,
+  params,
+  reload
+}) => {
+  const repositories = await context.apiClient.qbert.getRepositories().then(prop('data'))
```
brennancheung: Is there anything else useful in that response?  If not, let's change the api-client to just return `data` at the top level.

src/app/plugins/kubernetes/components/apps/RepositoriesListPage.js
```diff
@@ -0,0 +1,55 @@
...
+import { loadInfrastructure } from 'k8s/components/infrastructure/actions'
+import { withAppContext } from 'core/AppContext'
+import Loader from 'core/components/Loader'
+import { compose } from 'ramda'
+import requiresAuthentication from 'openstack/util/requiresAuthentication'
+import { withDataLoader } from 'core/DataLoader'
+
+const ListPage = ({ ListContainer }) => {
+  class ListPage extends React.Component {
+    async componentDidMount () {
```
brennancheung: Can `loadRepositories` be moved into `DataLoader`.  It can take an array to handle multiple loaders.

src/app/plugins/kubernetes/components/apps/actions.js
xantorres: I have to refactor a few places, will do in next PR

src/app/plugins/kubernetes/components/apps/RepositoriesListPage.js
```diff
@@ -1,34 +1,13 @@
...
-      if (!repositories) { return <Loader /> }
-      return (
-        <ListContainer data={repositories} />
-      )
-    }
-  }
-
-  return withAppContext(ListPage)
+  return ({ context: { repositories } }) =>
+    !repositories ? <Loader /> : <ListContainer data={repositories} />
```
brennancheung: `DataLoader` shows the `<Loader />` while data is loading already.

brennancheung: Also, use `props.data.repositories` instead of from the context.

src/app/plugins/kubernetes/components/apps/RepositoriesListPage.js
```diff
@@ -49,7 +28,7 @@ export const options = {
 
 const { ListPage: RepositoriesListPage } = createCRUDComponents(options)
 
-export default compose(
-  requiresAuthentication,
-  withDataLoader({ dataKey: 'clusters', loaderFn: loadInfrastructure }),
-)(RepositoriesListPage)
+export default withDataLoader({
+  dataKey: 'clusters',
```
brennancheung: `dataKey: ['clusters', 'repositories']`

brennancheung: We might want to refactor the API for `DataLoader` at some point to use an object of `{ key: loaderFn, ... }`.  The array form is a bit weird and not very intuitive.

PR #285, opened 2019-02-28
Feature/improve data loader

xantorres:
POC for MultiLoader component and withMultiLoader HoC

### DIFF
--- a/src/app/core/DataLoader.js
+++ b/src/app/core/DataLoader.js
@@ -1,68 +1,120 @@
-import React from 'react'
+import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
-import { all, partition, pick, props } from 'ramda'
-import { ensureArray, exists, propExists } from 'app/utils/fp'
-import { withAppContext } from './AppContext'
+import { pick, intersection, flatten, head } from 'ramda'
+import { ensureArray } from 'app/utils/fp'
 import DisplayError from './components/DisplayError'
 import Loader from './components/Loader'
+import { withAppContext } from 'core/AppContext'
 
-class DataLoaderBase extends React.Component {
+class MultiLoaderBase extends PureComponent {
   state = {
     loading: false,
     error: null,
   }
 
   componentDidMount () {
-    this.listener = window.addEventListener('scopeChanged', () => this.loadData())
-    this.loadData()
+    this.listener = window.addEventListener('scopeChanged', this.loadAll())
+    this.loadAll()
   }
 
   componentWillUnmount () {
     window.removeEventListener('scopeChanged', this.listener)
   }
 
-  doneLoading = () => {
-    const { dataKey, context } = this.props
-    const dataKeys = ensureArray(dataKey)
-    return all(exists, props(dataKeys, context))
+  /**
+   * Returns a map of [loaderKeys, promise] arrays that rely one on another and will wait for the
+   * dependant loaders regardless of the order on which they are called
+   * @param loaders
+   */
+  getLinkedLoaders = loaders => {
+    const loaderPairs = Array.isArray(loaders)
+      ? loaders
+      : Object.entries(loaders)
+
+    const linkedLoaders = loaderPairs.map(([loaderKeys, loaderSpec]) => {
+      const { loaderFn, requires } =
+        typeof loaderSpec === 'function'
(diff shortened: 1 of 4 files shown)


src/app/core/DataLoader.js
```diff
@@ -1,68 +1,111 @@
...
+
+    return children({ data, loading, error, context, reload: this.loadOne })
   }
 }
 
-const DataLoader = withAppContext(DataLoaderBase)
+const MultiLoader = withAppContext(MultiLoaderBase)
+
+const DataLoader = ({ dataKey, loaderFn, children }) => (
+  <MultiLoader loaders={{ [dataKey]: loaderFn }}>{children}</MultiLoader>
```
brennancheung: `dataKey` can be an array.  This will break because of `loaders={{ [['clusters', 'pods']]: loaderFn }}` is not a valid computed property.

src/app/core/DataLoader.js
```diff
@@ -74,22 +117,39 @@ DataLoader.propTypes = {
   dataKey: PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.arrayOf(PropTypes.string),
-  ]).isRequired,
+  ]),
```
brennancheung: Why remove the required for `dataKey` and `loaderFn`?  I can't see any benefit to using the component without these.

src/app/core/DataLoader.js
```diff
@@ -1,68 +1,111 @@
...
-        if (typeof err === 'string') { return err }
-        if (err instanceof Error) { return err.message }
+  /**
+   * Using on the "requires" prop of each loader spec, return a map of promises
+   * that rely one on another and will wait for the dependant loaders regardless of
+   * the order in which they are called
+   * @param loaders
+   */
+  linkedLoaders = loaders => {
+    const linkedLoaders = mapObjIndexed((loaderKey, loaderSpec) => {
```
brennancheung: You're using the same token as a `const` and as a `function` elsewhere.  This is a bit confusing.

src/app/plugins/kubernetes/components/apps/AppCatalogPage.js
```diff
@@ -51,7 +29,9 @@ class AppCatalogPage extends React.Component {
       field: 'clusterId',
       type: 'select',
       label: 'Cluster',
-      onChange: this.handleClusterChange,
+      onChange: async clusterId => {
+        this.props.reload('apps', { clusterId })
```
brennancheung: Is the catalog empty until they choose a cluster.  Wondering if we should default it to a cluster.  I can see it being useful that we have some `context` for `lastUsedCluster`.  I think people are going to want to see different things across the same cluster and having to re-select the cluster on each screen is going to be annoying.

src/app/plugins/kubernetes/components/apps/actions.js
```diff
@@ -43,7 +45,8 @@ export const loadReleases = async ({
 
   const releases = clusterId === '__all__'
     ? await Promise.all(context.clusters.map(loadClusterReleasesFromContext)).then(flatten)
-    : await loadClusterReleasesFromContext(context.clusters.find(cluster => cluster.uuid === clusterId))
+    : await loadClusterReleasesFromContext(context.clusters.find(
```
brennancheung: `propEq`

src/app/plugins/kubernetes/components/apps/AppCatalogPage.js
xantorres: `loadApps` is already loading apps belonging to the first cluster by default

src/app/core/DataLoader.js
xantorres: bad refactor

src/app/core/DataLoader.js
```diff
@@ -1,68 +1,111 @@
...
 
-  loadData = () => {
-    const { loaderFn, setContext, context } = this.props
-    if (!this.doneLoading()) {
-      this.setState({ loading: true })
-      const parseErr = err => {
-        if (typeof err === 'string') { return err }
-        if (err instanceof Error) { return err.message }
+  /**
+   * Using on the "requires" prop of each loader spec, return a map of promises
```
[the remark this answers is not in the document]
xantorres: I changed it, I hope it is clearer now

PR #288, opened 2019-03-06
Redirect navItems to the old UI

brennancheung:
This redirects the current nav item links to the old clarity UI.

During development we can `window.localStorage.disableClarityLinks = true`.

This only works when clicking on nav links.  It does not redirect to Clarity on initial page loads.  We probably don't need to worry about this since it is going to be rare for people to bookmark pages on the new UI before it is fully ported.

### DIFF
--- a/src/app/utils/fp.js
+++ b/src/app/utils/fp.js
@@ -9,6 +9,9 @@ export const isTruthy = x => !!x
 export const exists = x => x !== undefined
 export const propExists = curry((key, obj) => obj[key] !== undefined)
 
+// Works for arrays and strings.  All other types return false.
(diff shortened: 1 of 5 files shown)


src/app/utils/fp.js
```diff
@@ -9,6 +9,9 @@ export const isTruthy = x => !!x
 export const exists = x => x !== undefined
 export const propExists = curry((key, obj) => obj[key] !== undefined)
 
+// Works for arrays and strings.  All other types return false.
+export const notEmpty = arr => !!(arr && arr.length)
```
kevinylee1991: maybe name it notEmptyArray? I was confused what the function did in the other files until I saw it defined

PR #290, opened 2019-03-07
Add nav icons.  Styling changes for sideNav and topNav

brennancheung:
screen shot 2019-03-06 at 5 38 33 pm


src/app/core/components/Toolbar.js
```diff
@@ -67,7 +69,9 @@ const Toolbar = ({ classes, open, handleDrawerOpen }) => (
       <img src={logoPath} className={classes.logo} align="middle" />
       <div className={classes.rightTools}>
         <RegionChooser />
+        &nbsp; &nbsp; &nbsp;
```
kevinylee1991: Would it make more sense to use margin?

PR #293, opened 2019-03-08
Added new properties to localStorage user, fixed region picker, fixed…

kevinylee1991:
… regions and tokens in simulator

Still need to change old UI to use the newer 'user-preferences-<username>' localStorage property with last used tenant/region data.

### DIFF
--- a/src/api-client/Keystone.js
+++ b/src/api-client/Keystone.js
@@ -103,12 +103,36 @@ class Keystone {
     try {
       const response = await axios.post(this.tokensUrl, body)
       const scopedToken = response.headers['x-subject-token']
+      const _user = response.data.token.user
+      const roles = response.data.token.roles
+
+      const matchRole = (roles, roleName) => {
+        return roles.some((role) => role && role.name === roleName)
+      }
+
+      const role = roles.reduce((acc, current) => {
+        if (matchRole([acc, current], 'admin')) {
+          return 'admin'
+        } else if (matchRole([acc, current], '_member')) {
+          return '_member_'
+        } else {
+          return current.name
+        }
+      }, null)
+
+      const user = {
+        username: _user.name,
+        userId: _user.id,
+        tenantId: projectId,
+        role
+      }
       this.client.activeProjectId = projectId
       this.client.scopedToken = scopedToken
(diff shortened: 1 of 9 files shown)


src/api-client/Keystone.js
```diff
@@ -103,12 +103,36 @@ class Keystone {
     try {
       const response = await axios.post(this.tokensUrl, body)
       const scopedToken = response.headers['x-subject-token']
+      const _user = response.data.token.user
+      const roles = response.data.token.roles
+
+      const matchRole = (roles, roleName) => {
```
brennancheung: `const matchRole = (roles, roleName) => roles.some(propEq('name'))`

src/api-client/Keystone.js
```diff
@@ -103,12 +103,36 @@ class Keystone {
...
+        return roles.some((role) => role && role.name === roleName)
+      }
+
+      const role = roles.reduce((acc, current) => {
```
brennancheung: This makes it very hard to understand the logic.  I'm not sure what you are doing here.  But maybe a `map` followed by a `uniq` might make more sense here?

src/api-client/Keystone.js
brennancheung: Also maybe `const containsRoles = (roles, roleName) => pluck('name', roles).contains(roleName)`

brennancheung: contains is a little clearer than match

src/api-client/Keystone.js
kevinylee1991: I added a comment above the code block, but I wasn't able to think of a clearer solution that doesn't add steps

src/api-client/Keystone.js
```diff
@@ -103,12 +104,25 @@ class Keystone {
     try {
       const response = await axios.post(this.tokensUrl, body)
       const scopedToken = response.headers['x-subject-token']
+      const _user = response.data.token.user
+      const roles = response.data.token.roles
+
+      const roleNames = roles.map((role) => role.name)
```
brennancheung: `const roleNames = pluck('name', roles)`

PR #297, opened 2019-03-14
Monitoring

brennancheung:
This is the first version of the Managed Prometheus solution.  Currently it is very alpha.  The backend will be changing to consolidate the API into 1 call instead of 4.

The way the API is currently structured we need to create 4 CRD objects (`prometheuses`, `servicemonitors`, `prometheusrules`, and `alertmanagers`) for each Prometheus Instance in the UI.  They are associated using the standard Kubernetes `labels` / `selector` paradigm.

The UI makes the assumption of a 1-to-1 relation between the different CRDs but that is not the case in the backend.  As a result, deleting is difficult because we don't know which objects to delete.

No simulator endpoints were created due to the fact that the API will be radically changing.

### DIFF
--- a/src/api-client/Qbert.js
+++ b/src/api-client/Qbert.js
@@ -316,6 +316,19 @@ class Qbert {
   async deleteRepositoriesForCluster (clusterId, repoId) {
     return this.client.basicDelete(`${await this.clusterMonocularBaseUrl(clusterId)}/repos/${repoId}`)
   }
+
+  /* Managed Apps */
+  async getPrometheusInstances (clusterId) {
+    return this.client.basicGet(`${await this.baseUrl()}/clusters/${clusterId}/k8sapi/apis/monitoring.coreos.com/v1/prometheuses`)
+  }
+
+  async getPrometheusServiceMonitors (clusterId) {
+    return this.client.basicGet(`${await this.baseUrl()}/clusters/${clusterId}/k8sapi/apis/monitoring.coreos.com/v1/servicemonitors`)
+  }
+
+  async getPrometheusRules (clusterId) {
+    return this.client.basicGet(`${await this.baseUrl()}/clusters/${clusterId}/k8sapi/apis/monitoring.coreos.com/v1/prometheusrules`)
+  }
 }
 
 export default Qbert
--- a/src/api-client/Qbert.js
+++ b/src/api-client/Qbert.js
@@ -1,4 +1,4 @@
-import { hasKeys, keyValueArrToObj } from 'utils/fp'
+import { keyValueArrToObj } from 'utils/fp'
 
 /* eslint-disable camelcase */
 class Qbert {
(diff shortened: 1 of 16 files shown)


src/api-client/Qbert.js
```diff
@@ -316,6 +318,107 @@ class Qbert {
...
+      apiVersion,
+      kind: 'ServiceMonitor',
+      metadata: {
+        name: `${data.name}-service-monitor`,
+        namespace: data.namespace,
+        labels: serviceMonitor,
+      },
+      spec: {
+        endpoints: [
+          { port: data.port }, // TODO: we don't have this data from the UI yet
```
kevinylee1991: is this TODO done?

src/api-client/Qbert.js
```diff
@@ -316,6 +318,107 @@ class Qbert {
...
+      spec: {
+        replicas: data.numInstances,
+        retention: `${data.retention}d`,
+        resources: { requests },
+        serviceMonitorSelector: { matchLabels: serviceMonitor },
+        ruleSelector: { matchLabels: ruleSelector },
+      }
+    }
+
+    // TODO: How do we specifiy "Enable persistent storage" in the API call?  What does this field mean in the
```
kevinylee1991: is this TODO done?

src/app/plugins/kubernetes/components/prometheus/AddPrometheusInstancePage.js
```diff
@@ -0,0 +1,129 @@
...
+                    <TextField id="cpu" label="CPU" info="Expressed in millicores (1m = 1/1000th of a core)" type="number" />
+                    <TextField id="memory" label="Memory" info="MiB of memory to allocate" type="number" />
+                    {enableStorage && <TextField id="storage" label="Storage" info="The storage allocation.  Default is 8 GiB" type="number" />}
+
+                    <PicklistField
+                      id="cluster"
+                      options={clusterOptions}
+                      onChange={this.handleClusterChange}
+                      label="Cluster"
+                      info="Clusters available with RoleBing from admin delegation"
```
kevinylee1991: Is RoleBing a typo?

src/app/plugins/kubernetes/components/prometheus/actions.js
```diff
@@ -0,0 +1,52 @@
...
+  retention: spec.retention,
+  replicas: spec.replicas,
+  metadata,
+  spec,
+})
+
+export const loadPrometheusResources = async ({ context, setContext, reload }) => {
+  if (!reload && context.prometheusInstances) { return context.prometheusInstances }
+
+  const instancesResponse = await context.apiClient.qbert.getPrometheusInstances('e8f1d175-2e7d-40fa-a475-ed20b8d8c66d')
```
kevinylee1991: is this ID supposed to be preset? There are a few more instances of this ID below as well

src/api-client/Qbert.js
brennancheung: Yup, done.  Removing.

src/api-client/Qbert.js
brennancheung: Nope, has not been discussed or spec'ed out yet.  Still needs elaboration from product / backend.

src/app/plugins/kubernetes/components/prometheus/AddPrometheusInstancePage.js
brennancheung: Yes, typo.  Meant to be `RoleBinding`.  Changing.

src/app/plugins/kubernetes/components/prometheus/actions.js
brennancheung: No, we need some way to pass through the currently selected cluster.  That's an entire PR in and of itself.  I'll address that in a different PR and work with @xantorres on the `withCluster` functionality.

PR #298, opened 2019-03-16
Revamp of Loader (now Progress) and avoid remounting after refetching data

xantorres:
The new `<Progress />` component now accepts children that will be rendered after the first loading (first time `loading` prop is set to `false`).
image

Subsequent loads will fade the rendered content rather than remounting the whole children tree (when `overlay` prop is true, otherwise content will just get hidden):
image

There is also an `inline` prop for loading of small inline components
image

Also I have fixed some issues with the `DataLoader` component and `AddClusterPage`

TODO: Refactor some pages that are loading data on `componentDidMount` to use the new `withDataLoader`/`withMultiLoader` HoC's that already implements the new `<Progress />` component


src/app/core/DataLoader.js
```diff
@@ -95,23 +96,20 @@ class MultiLoaderBase extends PureComponent {
...
-      : Object.entries(loaders)
+    const { context, loaders, children, options } = this.props
+    const loaderPairs = this.getLoaderPairs(loaders)
     const dataKeys = flatten(loaderPairs.map(head))
     const data = dataKeys.length === 1 ? context[dataKeys[0]] : pick(dataKeys, context)
-    if (loading || !data) {
-      return <Loader />
-    }
-    return children({ data, loading, error, context, reload: this.loadOne })
+    return <Progress inline={options.inlineProgress} overlay loading={loading || !data}>
```
brennancheung: Is there a benefit to wrapping vs the alternation we used before?  This has 1 more component in the React component hierarchy (React component hierarchy hell) vs the previous way results in less nodes in the hierarchy.

src/app/constants.js
```diff
@@ -1,5 +1,10 @@
 export const appUrlRoot = '/ui'
+export const imageUrlRoot = `${appUrlRoot}/images`
 export const loginUrl = `${appUrlRoot}/openstack/login`
 export const logoutUrl = `${appUrlRoot}/openstack/logout`
 export const dashboardUrl = `${appUrlRoot}/kubernetes/infrastructure`
-export const logoPath = `${appUrlRoot}/images/logo.png`
+
+export const imageUrls = Object.freeze({
```
brennancheung: Is `freeze` a performance optimization?  Curious why just here and why clutter with more verbose code.

src/app/core/components/Picklist.js
```diff
@@ -36,29 +36,29 @@ class Picklist extends React.Component {
   }
 
   render () {
-    const { className, classes, label, name } = this.props
+    const { className, classes, label, name, value, options } = this.props
 
-    const options = this.props.options.map(x =>
-      typeof x === 'string' ? ({ value: x, label: x }) : x
+    const items = options.map(x =>
+      typeof x === 'string' ? ({ value: x, label: x }) : x,
```
brennancheung: trailing comma?

src/app/core/components/Picklist.js
```diff
@@ -36,29 +36,29 @@ class Picklist extends React.Component {
...
     ).map(x => ({
       label: x.label,
       // Hack to work around Material UI's Select ignoring empty string as a value
-      value: x.value === '' ? '__none__' : x.value
+      value: x.value === '' ? '__none__' : x.value,
```
brennancheung: comma

src/app/core/components/Picklist.js
```diff
@@ -36,29 +36,29 @@ class Picklist extends React.Component {
...
     }))
 
     // Hack to work around Material UI's Select ignoring empty string as a value
-    const value = this.props.value === '' ? '__none__' : this.props.value
-
+    const nonEmptyValue = value === '' ? '__none__' : value
+    console.log(value, items)
```
brennancheung: remove `console.log`

src/app/constants.js
xantorres: There is no performance optimization, and I think the verbosity is really negligible.
Freeze simply clearly states that the object is immutable, and since we are exporting constants, it makes it explicit that the `imageUrls` object should not (and in fact _can't_)  be modified.

src/app/core/components/Picklist.js
xantorres: This has been added by `eslint --fix` due to a rule we have with trailing commas
image

src/app/core/components/Picklist.js
xantorres: This has been added by `eslint --fix` due to a rule we have with trailing commas
image
I think it's ok to leave it as it is.

src/app/core/DataLoader.js
xantorres: The advantage is that we will not re-render all children tree every time a load is performed, which will improve performance and memory usage and also make the app look more "natural" since the children component will not cause that weird looking "flickering" caused by the children disappearing and being mounted again.

PR #301, opened 2019-03-25
Fully switch to JSS (@material-ui/styles) to fix glitches and improve stability

xantorres:


xantorres:
> Can you add some more documentation on the what and why for this PR? Does this fix the weird CSS issues with the `ProgressBar` component?
> 
> I did notice that the `ProgressBar` issue was random but it was much more likely if the UI started loading with a page that has the `ProgressBar`. Was the CSS-in-JS system not set up correctly at load time when the component was displayed?

Yes, this should fix all the weird issues. Until now I was using the new css-in-js just in a few places but now I properly installed the new css-in-js as recommended by the material-ui guys and it seems all issues are gone.

PR #302, opened 2019-04-02
Form styles adjustments, migration to ThemeProvider (from MuiThemeProvider) and misc style fixes

xantorres:
image

### DIFF
--- a/src/app/core/components/Picklist.js
+++ b/src/app/core/components/Picklist.js
@@ -1,12 +1,11 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import FormControl from '@material-ui/core/FormControl'
-import InputLabel from '@material-ui/core/InputLabel'
 import MenuItem from '@material-ui/core/MenuItem'
-import Select from '@material-ui/core/Select'
 import { withStyles } from '@material-ui/styles'
 import { compose } from 'app/utils/fp'
 import classnames from 'classnames'
+import TextField from '@material-ui/core/TextField'
 
 /**
  * Picklist is a bare-bones widget-only implmentation.
@@ -36,7 +35,7 @@ class Picklist extends React.Component {
   }
 
   render () {
-    const { className, classes, label, name, value, options } = this.props
+    const { className, classes, label, name, value, options, filled } = this.props
 
     const items = options.map(x =>
       typeof x === 'string' ? ({ value: x, label: x }) : x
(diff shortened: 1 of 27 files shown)


src/app/core/components/Wizard.js
```diff
@@ -82,17 +89,20 @@ class Wizard extends React.Component {
...
       <div>
         <Provider value={this.state}>
           <ProgressTracker steps={steps} activeStep={step} />
           {children({ wizardContext, setWizardContext, onNext: this.onNext })}
           <FormButtons>
-            { this.hasBack() && <BackButton handleBack={this.handleBack} /> }
-            { this.hasNext() && <NextButton handleNext={this.handleNext}>Next</NextButton> }
-            { this.isLastStep() && <NextButton handleNext={this.handleNext}>{submitLabel}</NextButton> }
+            {this.hasBack() &&
+            <BackButton className={classes.button} handleBack={this.handleBack} />}
```
brennancheung: Might look cleaner indented.

Alternatively, putting a newline between each one could work as well.

src/app/plugins/openstack/components/images/AddImagePage.js
```diff
@@ -133,14 +132,14 @@ class AddImagePage extends React.Component {
 
   render () {
     return (
-      <MuiThemeProvider theme={textTheme}>
+      <ThemeProvider theme={textTheme}>
```
brennancheung: I'm wondering if we can refactor out this code and just use `withStyles` instead of creating a new theme just for a single page.  This might be from earlier when we didn't understand Material-UI as well.  Thoughts?

src/app/core/components/Picklist.js
```diff
@@ -50,15 +49,17 @@ class Picklist extends React.Component {
     const nonEmptyValue = value === '' ? '__none__' : value
     return (
       <FormControl className={classnames(classes.formControl, className)}>
-        <InputLabel htmlFor={name}>{label}</InputLabel>
-        <Select
+        <TextField
```
brennancheung: Selects go through `TextField`?  Is this a Material-UI quirk?

src/app/core/components/validatedForm/CheckboxField.js
```diff
@@ -10,11 +10,11 @@ import withFormContext, { ValidatedFormInputPropTypes } from 'core/components/va
 
 const styles = theme => ({
   formControl: {
-    margin: theme.spacing.unit
+    marginTop: theme.spacing.unit,
   }
 })
 
-class Checkbox extends React.Component {
+class CheckboxField extends React.Component {
```
brennancheung: Much more consistent.  👍

src/app/core/components/Picklist.js
xantorres: Yes, it is the only way I found to have it with the `inner` prop

src/app/plugins/openstack/components/images/AddImagePage.js
xantorres: yes, I thought the same thing, but I didn't want to change anything else from the scope of the PR, will do in a separate one

PR #305, opened 2019-04-27
Start of work on theming support

brennancheung:
* Created a plugin to allow customization of the theme values.
* Added color picker component.
* Added a `<KitchenSink />` component where we can add our component library.
* Added common palette colors in property inspector area.

Future Work:
* Add more components to `<KitchenSink />` preview
* Load theme
* Save theme

localhost_3000_ui_themes_configure

### DIFF
--- a/src/app/ThemeManager.js
+++ b/src/app/ThemeManager.js
@@ -1,0 +1,40 @@
+import React from 'react'
+import { withAppContext } from 'core/AppContext'
+import { deepOrange } from '@material-ui/core/colors'
+import { createMuiTheme } from '@material-ui/core/styles'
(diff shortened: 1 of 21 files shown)


src/app/ThemeManager.js
```diff
@@ -0,0 +1,44 @@
...
+    },
+    secondary: deepOrange,
+    sidebar: {
+      background: '#243748',
+      text: '#aee0ff',
+    }
+  }
+}
+
+class ThemeManager extends React.Component {
```
xantorres: Would it make sense to connect this with `withScopedPreferences` HoC to make the settings preserved upon page reloads?

PR #308, opened 2019-05-08
Bugfix/fix data loading

xantorres:
This should fix the issues around the new contextLoader/Updater helpers and specially with clusterContextLoader/Updater.

There are also improvements with these helpers, like now having `apiClient` and `loadFromContext` in `contextLoader` args
image

And `apiClient` and `currentItems` in `contextUpdater` args
image

### DIFF
--- a/src/app/core/helpers/clusterContextLoader.js
+++ b/src/app/core/helpers/clusterContextLoader.js
@@ -1,24 +1,52 @@
-import { pathOr, chain } from 'ramda'
+import { pathOr, prop, groupBy, assocPath, flatten, omit, path } from 'ramda'
 import { loadClusters } from 'k8s/components/infrastructure/actions'
 import contextLoader from 'core/helpers/contextLoader'
+import { ensureArray } from 'utils/fp'
 
 // Returns a contextLoaded function contextualized by selected cluster (given by clusterId param)
-const clusterContextLoader = (key, loaderFn, defaultValue = []) =>
-  async ({ context, params = {}, ...rest }) => {
-    const clusters = (await loadClusters({ context, params, ...rest, reload: false }))
-      .filter(cluster => cluster.hasMasterNode)
+// Use is the same as in "contextLoader" but it also receives an additional clusterId param and
+// clusters that are passed as arguments to the loaderFn callback
+// Context will be indexed by "clusterId" and also have an "__all__" prop with all the items:
+// {
+//   "ba39856e-2580-4a72-a247-2c8e79004e2b": [
+//     { "id": "stable/capacitor", },
+//     { "id": "stable/pixel", },
+//     { "id": "stable/card", }
+//   ],
+//   "68f95f1d-875a-46a1-a1c7-e75717a49793": []
+//   "__all__": [
+//     { "id": "stable/capacitor", },
+//     { "id": "stable/pixel", },
+//     { "id": "stable/card", }
+//   ],
+// }
+const clusterContextLoader = (key, loaderFn, filterMasterNodes = true, defaultValue = []) =>
+  async ({ getContext, setContext, params = {}, ...rest }) => {
+    const keyPath = ensureArray(key)
+    const clusters = (await loadClusters({ getContext, setContext, ...rest, reload: false }))
+      .filter(cluster => !filterMasterNodes || cluster.hasMasterNode)
     const { clusterId = pathOr('__all__', [0, 'uuid'], clusters) } = params
-
-    // loaderFn will receive "clusters" and "clusterId" in "params" object
-    const data = await contextLoader([key, clusterId], loaderFn, defaultValue)(
-      { context, params: { ...params, clusterId }, clusters, ...rest },
+    const loadedData = await contextLoader([...keyPath, clusterId], loaderFn, defaultValue)(
+      { getContext, setContext, params: { ...params, clusterId }, clusters, ...rest },
     )
-    return clusterId === '__all__'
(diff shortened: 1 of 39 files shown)


src/app/core/helpers/clusterContextLoader.js
```diff
@@ -1,24 +1,22 @@
-import { pathOr, chain } from 'ramda'
+import { pathOr, prop, groupBy } from 'ramda'
 import { loadClusters } from 'k8s/components/infrastructure/actions'
 import contextLoader from 'core/helpers/contextLoader'
 
 // Returns a contextLoaded function contextualized by selected cluster (given by clusterId param)
```
brennancheung: This is a bit hard to follow.  Elaborating a little more and also a usage example in the comments would help clarify.

src/app/core/helpers/clusterContextLoader.js
```diff
@@ -1,24 +1,22 @@
...
 const clusterContextLoader = (key, loaderFn, defaultValue = []) =>
-  async ({ context, params = {}, ...rest }) => {
-    const clusters = (await loadClusters({ context, params, ...rest, reload: false }))
+  async ({ context, setContext, params = {}, ...rest }) => {
+    const clusters = (await loadClusters({ context, setContext, params, ...rest, reload: false }))
       .filter(cluster => cluster.hasMasterNode)
```
brennancheung: Hmm, I'm wondering if we always want to exclude clusters without master nodes.  Perhaps this should be an option and have it default to true.

src/app/core/helpers/contextLoader.js
```diff
@@ -1,46 +1,53 @@
...
  * Returns a function that will use context to load and cache values
- * @param pathResolver Context path on which the resolved value will be cached, it can be a function
+ * @param contextPath Context path (a string or array for deep paths) on which the resolved value will be cached
  * @param loaderFn Function returning the data to be assigned to the context
  * @param defaultValue Default value assigned to the context
  * @returns {Function}
  */
-const contextLoader = (pathResolver, loaderFn, defaultValue = []) => {
-  const moizedPathResolver = moize(ensureFunction(pathResolver))
+const contextLoader = (contextPath, loaderFn, defaultValue = []) => {
```
brennancheung: Should any mapping—converting from the API version of an object, to the version needed inside the UI happen in the `loaderFn`?  This mapping function needs to be applied when loading, and also when create/update calls return an object.

src/app/core/helpers/contextLoader.js
```diff
@@ -1,46 +1,53 @@
...
-import { ensureFunction, ensureArray } from 'utils/fp'
+import { ensureArray } from 'utils/fp'
 
 let pendingPromises = {}
+let resolvers = {}
 
 /**
  * Returns a function that will use context to load and cache values
- * @param pathResolver Context path on which the resolved value will be cached, it can be a function
+ * @param contextPath Context path (a string or array for deep paths) on which the resolved value will be cached
```
brennancheung: Good call.  This adds a lot of flexibility.

src/app/core/helpers/contextLoader.js
```diff
@@ -1,46 +1,53 @@
...
-      dissocPath(resolvedPath, pendingPromises)
+      await setContext(prevContext => assocPath(arrContextPath, output, prevContext))
+      pendingPromises = dissocPath(arrContextPath, pendingPromises)
     }
     return output
   }
+
+  resolvers = assocPath(arrContextPath, resolver, resolvers)
+
+  return resolver
```
brennancheung: Add comment on what uses this return value and why it needs it.

src/app/plugins/kubernetes/components/apps/actions.js
```diff
@@ -1,31 +1,24 @@
...
 
 export const loadApps = clusterContextLoader(
   'apps',
   async ({ context, clusters, params: { clusterId } }) => {
-    const loadClusterAppsFromContext = getClusterApps(context)
+    const { qbert } = context.apiClient
     return clusterId === '__all__'
-      ? Promise.all(clusters.map(loadClusterAppsFromContext)).then(flatten)
-      : loadClusterAppsFromContext(clusters.find(propEq('uuid', clusterId)))
+      ? asyncFlatMap(pluck('uuid', clusters), qbert.getCharts)
```
brennancheung: Very minor and optional, do we want to adopt the convention we have used elsewhere of `{ service: 'qbert', entity: 'charts' }`?

src/app/core/helpers/contextLoader.js
xantorres: For this case I would suggest having a third argument in the `contextLoader` function, this argument would be a mapper that will receive the loaded data and will do any extra transformations, also this same function would be applied when using `contextUpdater`.

brennancheung: Yes, but I think we might want to refactor this to be an object `fn({ contextPath, loaderFn, ... })` instead of individual positional parameters.  Having 4 function parameters is too many when it is positional.

xantorres: I agree but still we will have to remember that this object always requires the `contextPath` and `loaderFn` props, what do you think about leaving the first two arguments as function parameters and have a third `options` object parameter with these additional optional things?

src/app/plugins/kubernetes/components/apps/actions.js
xantorres: Good point, I will look into it in another PR

PR #311, opened 2019-05-15
Prometheus edit

brennancheung:
This adds dashboard links and editing functionality for the different Prometheus Custom Resources for the Prometheus Operator.

The `contextUpdater` function seems buggy.  When I use it to perform a delete in `deletePrometheusRule` the context is updated but the table view does not refresh.

When I use `contextUpdater` to update a rule, it creates a `context[undefined]` entry instead of placing it in `context.prometheusRules`.

### DIFF
--- a/src/api-client/ApiClient.js
+++ b/src/api-client/ApiClient.js
@@ -83,6 +83,17 @@ class ApiClient {
     }
   }
 
+  async basicPatch (url, body) {
+    try {
+      const config = this.getAuthHeaders()
+      config.headers['Content-Type'] = 'application/json-patch+json'
+      const response = await axios.patch(url, body, config)
+      return response.data
+    } catch (err) {
+      console.log(err)
+    }
+  }
+
   async basicPut (url, body) {
     try {
       const response = await axios.put(url, body, this.getAuthHeaders())
--- a/src/app/core/components/validatedForm/ValidatedFormDebug.js
+++ b/src/app/core/components/validatedForm/ValidatedFormDebug.js
@@ -1,0 +1,15 @@
+import React from 'react'
+
+import { ValidatedFormConsumer } from 'core/components/validatedForm/ValidatedForm'
+
+/* This component is primarily designed to be used as a debugging tool to
+ * show the form context during development.
+ */
+const ValidatedFormDebug = () => (
+  <ValidatedFormConsumer>
+    {props => <pre>{JSON.stringify(props, null, 4)}</pre>}
+  </ValidatedFormConsumer>
+
+)
+
+export default ValidatedFormDebug
--- a/src/app/plugins/kubernetes/components/prometheus/UpdatePrometheusInstancePage.js
+++ b/src/app/plugins/kubernetes/components/prometheus/UpdatePrometheusInstancePage.js
@@ -1,0 +1,42 @@
+import React from 'react'
+import SubmitButton from 'core/components/SubmitButton'
+import TextField from 'core/components/validatedForm/TextField'
+import ValidatedForm from 'core/components/validatedForm/ValidatedForm'
+import createUpdateComponents from 'core/helpers/createUpdateComponents'
+import { loadPrometheusInstances, updatePrometheusInstance } from './actions'
+
+class UpdatePrometheusInstanceForm extends React.Component {
+  state = this.props.initialValues
+
+  handleUpdate = data => {
+    this.props.onComplete(data)
+  }
+
+  render () {
+    const initialValues = { ...this.props.initialValues }
+    return (
+      <ValidatedForm initialValues={initialValues} onSubmit={this.handleUpdate}>
+        <TextField id="replicas" label="# of instances" info="Number of Prometheus instances" type="number" />
+        <TextField id="cpu" label="CPU" info="Expressed in millicores (1m = 1/1000th of a core)" />
+        <TextField id="memory" label="Memory" info="MiB of memory to allocate" />
+        <TextField id="retention" label="Storage Retention (days)" info="Defaults to 15 days if nothing is set" />
+        <SubmitButton>Update Prometheus Instance</SubmitButton>
+      </ValidatedForm>
+    )
+  }
+}
+
+export const options = {
+  FormComponent: UpdatePrometheusInstanceForm,
+  routeParamKey: 'id',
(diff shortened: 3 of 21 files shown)


src/app/plugins/kubernetes/components/prometheus/UpdatePrometheusInstancePage.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+import SubmitButton from 'core/components/SubmitButton'
+import TextField from 'core/components/validatedForm/TextField'
+import ValidatedForm from 'core/components/validatedForm/ValidatedForm'
+import createUpdateComponents from 'core/helpers/createUpdateComponents'
+import { loadPrometheusInstances, updatePrometheusInstance } from './actions'
+
+class UpdatePrometheusInstanceForm extends React.Component {
+  state = this.props.initialValues
```
xantorres: This state doesn't seem to be used anywhere

src/app/plugins/kubernetes/components/prometheus/UpdatePrometheusInstancePage.js
```diff
@@ -0,0 +1,42 @@
...
+
+  handleUpdate = data => {
+    this.props.onComplete(data)
+  }
+
+  render () {
+    const initialValues = { ...this.props.initialValues }
```
xantorres: `const { initialValues } = this.props`

src/app/plugins/kubernetes/components/prometheus/UpdatePrometheusRulePage.js
```diff
@@ -0,0 +1,55 @@
...
+import createUpdateComponents from 'core/helpers/createUpdateComponents'
+import uuid from 'uuid'
+import { loadPrometheusRules, updatePrometheusRules } from './actions'
+import { withStyles } from '@material-ui/styles'
+
+@withStyles(theme => ({
+  submit: { marginTop: theme.spacing.unit * 3 },
+}))
+class UpdatePrometheusRuleForm extends React.Component {
+  state = this.props.initialValues
```
xantorres: this `state` is initialized with `initialValues` but it seems to be using `rules`, is this correct?

src/app/plugins/kubernetes/components/prometheus/UpdateServiceMonitorPage.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+import KeyValuesField from 'core/components/validatedForm/KeyValuesField'
+import SubmitButton from 'core/components/SubmitButton'
+import ValidatedForm from 'core/components/validatedForm/ValidatedForm'
+import createUpdateComponents from 'core/helpers/createUpdateComponents'
+import { loadPrometheusServiceMonitors, updatePrometheusServiceMonitor } from './actions'
+import { keyValueArrToObj, objToKeyValueArr } from 'utils/fp'
+
+class UpdateServiceMonitorForm extends React.Component {
+  state = this.props.initialValues
```
xantorres: unused state

src/app/plugins/kubernetes/components/prometheus/actions.js
```diff
@@ -3,35 +3,27 @@ import { pathOr, prop, propEq } from 'ramda'
...
 
-const mapServiceMonitor = ({ clusterUuid, metadata, spec }) => ({
-  clusterUuid,
-  name: metadata.name,
-  namespace: metadata.namespace,
-  labels: metadata.labels,
-  port: spec.endpoints.map(prop('port')).join(', '),
-  selector: spec.selector.matchLabels,
-})
+const mapAsyncItems = async (values, loaderFn, mapFn) => {
```
xantorres: this looks good wouldn't just a `(await asyncFlatMap(values, loaderFn)).map(mapFn)` do the same?

src/app/core/components/validatedForm/ValidatedFormDebug.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+import { ValidatedFormConsumer } from 'core/components/validatedForm/ValidatedForm'
+
+/* This component is primarily designed to be used as a debugging tool to
+ * show the form context during development.
+ */
+const ValidatedFormDebug = () => (
+  <ValidatedFormConsumer>
```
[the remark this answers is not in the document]
brennancheung: Mostly convenience.  It's a lot more typing.  `console.log` doesn't work so well with HMR and having to expand out the objects every time in the console is more time consuming and harder to see things when is expanded and one the screen.  It's also something that needs to be added frequently when debugging Form context related issues.

src/app/plugins/kubernetes/components/prometheus/actions.js
```diff
@@ -3,35 +3,27 @@ import { pathOr, prop, propEq } from 'ramda'
...
 })
 
-const mapAlertManager = ({ clusterUuid, metadata, spec }) => ({
-  clusterUuid,
-  name: metadata.name,
-  namespace: metadata.namespace,
-  replicas: spec.replicas,
-  labels: metadata.labels,
-})
+const hasMonitoring = cluster => cluster.tags.includes('pf9-system:monitoring')
```
kevinylee1991: does this still use tags? I thought that sachin changed it to be based on some appbert packages API

src/api-client/ApiClient.js
```diff
@@ -83,6 +83,17 @@ class ApiClient {
     }
   }
 
+  async basicPatch (url, body) {
+    try {
+      const config = this.getAuthHeaders()
+      config.headers['Content-Type'] = 'application/json-patch+json'
```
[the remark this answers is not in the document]
brennancheung: Yes, it's required for patch.  I tried it with just `application/json` and it didn't work.

src/app/plugins/kubernetes/components/prometheus/actions.js
brennancheung: It does, but it's an additional API call.

src/app/plugins/kubernetes/components/prometheus/actions.js
brennancheung: I think so.  It's a bit more verbose though.

PR #313, opened 2019-05-23
Fixed issue with API client using wrong region

kevinylee1991:
Basically removed the part that set the active region in App.js from config, and instead set the active region during Session setup.

### DIFF
--- a/config.example.js
+++ b/config.example.js
@@ -48,4 +48,8 @@ const config = {
 
 const env = process.env.NODE_ENV || 'development'
 
+if (env === 'development' && !config.development.simulator) {
+  throw new Error('config.development.simulator not found')
+}
+
 module.exports = config[env]
--- a/src/app/plugins/openstack/components/SessionManager.js
+++ b/src/app/plugins/openstack/components/SessionManager.js
@@ -8,7 +8,7 @@ import { withPreferences } from 'core/providers/PreferencesProvider'
 import { getStorage, setStorage } from 'core/utils/pf9Storage'
 import LoginPage from 'openstack/components/LoginPage'
 import { loadUserTenants } from 'openstack/components/tenants/actions'
-import { pathOr, propEq } from 'ramda'
+import { path, pathOr, propEq } from 'ramda'
 
 /**
  * Sets up the Openstack session.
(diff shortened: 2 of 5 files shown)


config.example.js
```diff
@@ -48,4 +48,8 @@ const config = {
 
 const env = process.env.NODE_ENV || 'development'
 
+if (env === 'development' && !config.development.simulator) {
+  throw new Error('config.development.simulator not found')
```
brennancheung: Did this actually throw the error message?  It seemed like it was an error during babel/webpack compilation, not runtime.

Also, it is perfectly valid to not have a simulator config if you are targeting a real DU only.

src/app/plugins/openstack/components/SessionManager.js
```diff
@@ -50,11 +50,15 @@ class SessionManager extends React.Component {
...
     const userPreferences = await initUserPreferences(username)
+
     const lastTenant = pathOr('service', ['Tenants', 'lastTenant', 'name'], userPreferences)
+    const lastRegion = path(['RegionChooser', 'lastRegion', 'id'], userPreferences)
 
     const tenants = await loadUserTenants({ getContext, setContext })
     const activeTenant = tenants.find(propEq('name', lastTenant))
     const { keystone } = context.apiClient
+
+    if (lastRegion) { context.apiClient.setActiveRegion(lastRegion) }
```
brennancheung: Don't we need to set the region before querying tenants?  Or is the FQDN they visit enough?

Also verify everything still works in incognito mode when the user does not have anything saved yet.

config.example.js
kevinylee1991: Yup this threw an error message. I think this is more for my case (and to prevent any potential future cases with other people) where I wasn't able to use the webpack alias for some odd reason we were never able to make sense of.

src/app/plugins/openstack/components/SessionManager.js
kevinylee1991: It looks like all keystone requests will be going through the FQDN they visit, so I believe the order shouldn't matter.

Incognito also worked

brennancheung:
LGTM 🚢

PR #314, opened 2019-05-24
Import / Export theme.json

brennancheung:
This PR creates an export button to allow the user to save the `theme.json` and a corresponding Import button to allow the user to load the previously saved theme.

### DIFF
--- a/src/app/core/components/ExportDataButton.js
+++ b/src/app/core/components/ExportDataButton.js
@@ -1,0 +1,32 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Button } from '@material-ui/core'
+
+const ExportDataButton = ({ children, color, data, filename, variant }) => {
+  const exportData = () => {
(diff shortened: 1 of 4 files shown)


src/app/core/components/ExportDataButton.js
```diff
@@ -0,0 +1,32 @@
...
+
+const ExportDataButton = ({ children, color, data, filename, variant }) => {
+  const exportData = () => {
+    let dataStr = JSON.stringify(data, null, 4)
+    let dataUri = 'data:application/json;charset=utf-8,'+ encodeURIComponent(dataStr)
+    let exportFileDefaultName = filename
+    let linkElement = document.createElement('a')
+    linkElement.setAttribute('href', dataUri)
+    linkElement.setAttribute('download', exportFileDefaultName)
+    linkElement.click()
```
kevinylee1991: Not sure if 100% necessary but may be good to add a removeChild on the linkElement

PR #320, opened 2019-06-06
Add showToast to context and use in Prometheus

brennancheung:
Most of the toasts need to happen inside of the `actions.js`.  Since actions are not components, there is no way to use `showToast` so I added it to the `context`.  This makes it very convenient to use.

I added toasts to the `create`, `update`, and `delete` prometheus actions.

Screen Shot 2019-06-05 at 6 30 55 PM

Screen Shot 2019-06-05 at 6 30 25 PM

### DIFF
--- a/src/app/App.js
+++ b/src/app/App.js
@@ -72,8 +72,8 @@ class App extends React.Component {
       <Router>
         <ThemeProvider theme={theme}>
           <HotKeysProvider>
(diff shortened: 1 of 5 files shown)


brennancheung:
Thanks.  I'll see about changing it to the bottom right.

brennancheung:
Ok, there's in the bottom right above Intercom now.
Screen Shot 2019-06-05 at 6 55 56 PM

PR #323, opened 2019-06-08
Top extra content using a React Portal

xantorres:
image

### DIFF
--- a/src/app/core/components/Navbar.js
+++ b/src/app/core/components/Navbar.js
@@ -24,7 +24,7 @@ const styles = theme => ({
   search: {
     position: 'relative',
     borderRadius: theme.shape.borderRadius,
-    backgroundColor: '#fff',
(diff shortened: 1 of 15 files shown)


xantorres:
> This is roughly what I thought you meant when you proposed this solution before. Still not liking it. The create button has nothing to do with tabs. It grossly pollutes the logic of `Tabs` and violates the principle of "do one thing and do it well".
> 
> Also, we would have to do this for every scenario and combination of widgets that show up near the top of the page.
> 
> I still think putting the create button in the top header of the page and pushing it down to underneath the header using CSS positioning is the best and cleanest solution.

What about doing something similar but instead of creating it for the `Tabs` component, having it in the app `Toolbar` component absolutely positioning the "extra content" stuff absolutely positioned, but without the burden of using arbitrary text id's?

PR #330, opened 2019-06-19
Creating composable HoCs for data loading and mapping

xantorres:
As discussed with Brennan, this is part of the work for decoupling the data loading logic into smaller functions and HoC's

The implementation in the views looks as follows (complex example):
image

Currently still fixing a few issues and preparing some tests.

### DIFF
--- a/src/api-client/Qbert.js
+++ b/src/api-client/Qbert.js
@@ -278,9 +278,9 @@ class Qbert {
 
   /* Monocular endpoints being exposed through Qbert */
   getCharts = async (clusterId) => {
-    const output = this.client.basicGet(`${await this.clusterMonocularBaseUrl(clusterId)}/charts`)
+    const output = await this.client.basicGet(`${await this.clusterMonocularBaseUrl(clusterId)}/charts`)
     // FIXME: remove this when api is fixed (right now it is returning data.data)
-    return output.hasOwnProperty('data') ? output.data : output
+    return output && output.hasOwnProperty('data') ? output.data : output || []
   }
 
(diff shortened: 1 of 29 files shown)


src/api-client/Qbert.js
```diff
@@ -278,9 +278,9 @@ class Qbert {
 
   /* Monocular endpoints being exposed through Qbert */
   getCharts = async (clusterId) => {
-    const output = this.client.basicGet(`${await this.clusterMonocularBaseUrl(clusterId)}/charts`)
+    const output = await this.client.basicGet(`${await this.clusterMonocularBaseUrl(clusterId)}/charts`)
     // FIXME: remove this when api is fixed (right now it is returning data.data)
-    return output.hasOwnProperty('data') ? output.data : output
+    return output && output.hasOwnProperty('data') ? output.data : output || []
```
brennancheung: Might want to put this into some kind of utility.  Is this just `ensureArray`?

src/app/core/helpers/createAddComponents.js
```diff
@@ -34,6 +34,7 @@ const createAddComponents = options => {
       const { setContext, getContext, context, history } = this.props
       try {
         const existing = await (loaderFn || crudActions.list)({ setContext, getContext, context })
+        console.log('existing', existing)
```
brennancheung: remove

src/api-client/Qbert.js
xantorres: This is a workaround because the api is returning a nested data inside a data, which I suppose is a bug

PR #334, opened 2019-07-02
Feature/tenants and users

xantorres:

### DIFF
--- a/src/api-client/helpers.js
+++ b/src/api-client/helpers.js
@@ -4,7 +4,9 @@ import ApiClient from './ApiClient'
 const defaultTestTenant = 'Development Team Tenant'
 
 export const keystoneEndpoint = `${config.host}/keystone`
+export const keystoneAdminEndpoint = `${config.host}/keystone_admin`
 export const makeClient = () => new ApiClient({ keystoneEndpoint })
+export const makeAdminClient = () => new ApiClient({ keystoneEndpoint })
 
 export const getUserPass = () => {
   const username = config.username || config.simulator.username
--- a/src/app/core/helpers/clusterContextLoader.js
+++ b/src/app/core/helpers/clusterContextLoader.js
@@ -1,5 +1,4 @@
-import { pathOr, prop, groupBy, assocPath } from 'ramda'
-import { loadClusters } from 'k8s/components/infrastructure/actions'
+import { prop, groupBy, assocPath, path } from 'ramda'
 import contextLoader from 'core/helpers/contextLoader'
 import { ensureArray } from 'utils/fp'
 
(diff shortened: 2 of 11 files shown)


src/app/core/helpers/clusterContextLoader.js
```diff
@@ -20,30 +19,25 @@ import { ensureArray } from 'utils/fp'
...
-    })
-    if (clusterId === '__all__') {
-      // update all cluster indexed positions in bulk
-      await setContext(context => {
-        assocPath(keyPath, {
-          ...groupBy(prop('clusterId'), loadedData),
-          '__all__': loadedData,
-        }, context)
-      })
+const clusterContextLoader = (key, loaderFn) => {
```
brennancheung: Looks like this is still not using the data structure we talked about with `clusterId` being a param on the object.  This looks like it still has the `clusterId` as part of the path.

Also, this looks like it breaks existing code.  Is this going to break the existing Prometheus stuff?

This is going on months of constantly changing data loading over and over again and just wasting time, confusing other developers, and constantly breaking things.  We could have implemented all the data loading for all the components 3 times over by now as just simple code even if it is a little more verbose.  I'm not planning on using this abstraction anyways because with the new React hooks we can just perform the logic we need as a simple function without these huge abstractions.

Lenses (in the abstract sense) can be used for getting and setting data within a context like `clusterId`.

src/app/plugins/kubernetes/components/apps/actions.js
```diff
@@ -5,29 +5,30 @@ import { asyncFlatMap } from 'utils/fp'
 
 export const loadApps = clusterContextLoader(
   'apps',
-  async ({ apiClient, clusters, params: { clusterId } }) => {
+  async ({ apiClient, params: { clusterId }, loadFromContext }) => {
     const { qbert } = apiClient
-    return clusterId === '__all__'
-      ? asyncFlatMap(pluck('uuid', clusters), qbert.getCharts)
+    return !clusterId || clusterId === '__all__'
```
brennancheung: Move the `await loadFromContext('clusters')` into its own line.

```
if (...) { return qbert.getCharts(clusterId) }
const clusters = await loadFromContext...
return asyncFlatMap...
```

src/server/api/keystone/getTenants.js
```diff
@@ -0,0 +1,8 @@
+import context from '../../context'
```
brennancheung: We already have this in `src/server/api/keystone/getProjects.js`.

src/app/core/helpers/clusterContextLoader.js
xantorres: The change about storing the data in a single big array requires some work (is not as simple as it would seem) so here I'm still using the old indexing by `clusterId`, I simply made this change so that we could customize the getting/setting logic of the data in the context in an easy manner.

Also this is not exactly part of `DataLoader`, this is just a helper function that given a callback function stores the result in the context, if data doesn't exist already. `DataLoader` is just a component that given some data loading functions shows a spinner and calls them when mounting the component before rendering any children, once the data loading promises are resolved it renders the children.

This change is backwards compatible, so all existing prometheus code should work fine (I tested it anyways). This is basically just affecting things that belong to a cluster.

src/api-client/helpers.js
```diff
@@ -4,7 +4,9 @@ import ApiClient from './ApiClient'
 const defaultTestTenant = 'Development Team Tenant'
 
 export const keystoneEndpoint = `${config.host}/keystone`
+export const keystoneAdminEndpoint = `${config.host}/keystone_admin`
 export const makeClient = () => new ApiClient({ keystoneEndpoint })
+export const makeAdminClient = () => new ApiClient({ keystoneEndpoint })
```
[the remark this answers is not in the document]
xantorres: yes, actually this is not being used (yet), but I will remove it anyways

PR #340, opened 2019-07-10
Add Groups and Roles list pages

xantorres:


src/app/plugins/kubernetes/components/userManagement/UserRolesListPage.js
```diff
@@ -0,0 +1,24 @@
...
+  deleteFn: deleteRole,
+  name: 'Roles',
+  title: 'Roles',
+  uniqueIdentifier: 'id',
+}
+
+const { ListPage: UserRolesListPage } = createCRUDComponents(options)
+
+export default compose(
+  requiresAuthentication,
```
brennancheung: We can remove this here.  The API is already authenticated and this screen won't load unless the session is established with an authenticated user.

src/app/plugins/kubernetes/components/userManagement/actions.js
```diff
@@ -53,3 +53,60 @@ export const loadUsers = contextLoader('users', async ({ apiClient, loadFromCont
...
+  ])
+
+  return groups.map(group => {
+    const { rules = [] } = mappings.find(mapping => {
+      const rules = typeof mapping.rules === 'string' ? JSON.parse(mapping.rules) : mapping.rules
+      return !!find(
+        pipe(
+          pathOr([], ['local', 'items']),
+          find(pathEq(['group', 'id'], group.id)),
+          g => !!g, // Or: `complement(isNil)`
```
brennancheung: I think you can just use `Boolean` since it is a function and will force it to a boolean.

src/app/plugins/kubernetes/components/userManagement/actions.js
```diff
@@ -53,3 +53,60 @@ export const loadUsers = contextLoader('users', async ({ apiClient, loadFromCont
...
 }
+
+export const loadGroups = contextLoader('groups', async ({ apiClient, loadFromContext }) => {
+  const [groups, mappings] = await Promise.all([
+    apiClient.keystone.getGroups(),
+    apiClient.keystone.getGroupMappings(),
+  ])
+
+  return groups.map(group => {
+    const { rules = [] } = mappings.find(mapping => {
```
brennancheung: This is confusing.  Separate it out into separate statements.

Also, I see `const rules` twice within 2 lines so that confuses things as well.

src/app/plugins/kubernetes/components/userManagement/actions.js
```diff
@@ -53,3 +53,60 @@ export const loadUsers = contextLoader('users', async ({ apiClient, loadFromCont
...
+      const rules = typeof mapping.rules === 'string' ? JSON.parse(mapping.rules) : mapping.rules
+      return !!find(
+        pipe(
```
brennancheung: Store this pipeline function in a variable with a good name so it provides documentation on what it is doing.

src/app/plugins/kubernetes/components/userManagement/actions.js
xantorres: actually `!!` is just a shortcut for `Boolean`, but maybe it would look prettier here

PR #350, opened 2019-07-31
Small misc design tweaks

brennancheung:
* Changed the back to list view button to a close button like the current UI.
* Fixed the arrow on the next button.

### DIFF
--- a/src/app/core/components/FontAwesomeIcon.js
+++ b/src/app/core/components/FontAwesomeIcon.js
@@ -1,7 +1,17 @@
 import React from 'react'
 import clsx from 'clsx'
 
-const FontAwesomeIcon = ({ children, className, ...rest }) =>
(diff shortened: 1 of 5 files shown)


src/app/core/components/FontAwesomeIcon.js
```diff
@@ -1,7 +1,17 @@
...
+const FontAwesomeIcon = ({ children, className, size, solid, ...rest }) => {
+  const defaultClasses = [
+    'fal',
+    'fa-fw',
+    (size ? `fa-${size}` : 'fa-lg'),
+    `fa-${children}`,
+    (solid ? 'fas' : ''),
+  ]
+  return (
+    <i className={clsx(defaultClasses.join(' '), className)} {...rest} />
```
kevinylee1991: curious how fal and fas together would interact, does fas take priority?

PR #352, opened 2019-08-01
Wizard and Keyvalues design changes and minor refactoring and fixes

xantorres:
## SearchBar and Per Page
image

## Wizard (pending feedback from designer)

image

## Keys/Value field

image


src/app/core/components/FormWrapper.js
```diff
@@ -45,7 +45,7 @@ class FormWrapper extends React.Component {
               <Grid item>
                 <Button
                   className={classes.buttonBase}
-                  variant="outlined"
+                  variant="text"
```
brennancheung: This component has been replaced by `CloseComponent` in `master`.  Go with that one.

src/app/core/components/wizard/WizardStepper.js
```diff
@@ -0,0 +1,96 @@
...
+  root: {
+    color: '#EAEAF0',
+    display: 'flex',
+    height: 22,
+    alignItems: 'center',
+  },
+  active: {
+    color: '#784AF4',
+  },
+  circle: {
```
brennancheung: Screenshot didn't look like a circle, looked like a rounded square.  Most likely just need to set `borderRadius` much higher.  Set it to something big like `1000px`.  It shouldn't have any issues if it is bigger than the container.

src/app/core/components/Picklist.js
```diff
@@ -38,24 +38,24 @@ class Picklist extends React.Component {
     const { className, classes, label, name, value, options } = this.props
 
     const items = options.map(x =>
-      typeof x === 'string' ? ({ value: x, label: x }) : x
+      typeof x === 'string' ? ({ value: x, label: x }) : x,
```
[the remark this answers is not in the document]
xantorres: not sure what happened there

PR #353, opened 2019-08-04
Feature/styling adjustments

xantorres:
* Tabs on hover make it same as current UI
* Get rid of labels to the left of region and tenant 
* Add tooltip for region and tenant chooser 
* “Prometheus Instances” Tab should be renamed to “Instances”
* Clicking on “New Instance” should not deselect the “prometheus beta” menu item from left hand side menu
* For top nav bar, use same logo image as on pf9.platform9.net with same amount of spacing from left 
* Top nav bar height should be same as that on pf9.platform9.net
* Tabs should be fixed width and have a fixed right margin, text should be left aligned

#### Top selectors tooltips and Tooltip font-size and padding slightly increased: 
image

#### Tabs adjustments and animation on hover: (now they look like the clarity UI)
image

#### KeyValues component styling adjustments:
image

#### Showing a customizable message when the ListTable has no results
image

Other misc refactoring and styling fixes have been added


src/app/core/components/listTable/ListTable.js
```diff
@@ -406,13 +416,14 @@ class ListTable extends React.Component {
...
               onFiltersReset={this.handleFiltersReset}
+              onRefresh={onRefresh}
               batchActions={batchActions}
               rowsPerPage={rowsPerPage}
               onChangeRowsPerPage={this.handleChangeRowsPerPage}
               rowsPerPageOptions={[5, 10, 25, 50, 100]}
             />
             <div className={classes.tableWrapper}>
-              <Table className={classes.table}>
+              {paginatedData && paginatedData.length ? <Table className={classes.table}>
```
brennancheung: Hard to read the ternary.  Put them in variables first to make it easier to see.

PR #358, opened 2019-08-09
Bugfix/qbert server errors

xantorres:
* Fix Qbert api returning err instead of arrays
* Add `onlyMasterNodeClusters` option in `clusterizedDataLoader`

### DIFF
--- a/src/app/plugins/kubernetes/helpers/clusterizedDataLoader.js
+++ b/src/app/plugins/kubernetes/helpers/clusterizedDataLoader.js
@@ -1,4 +1,4 @@
-import { compose, pathOr } from 'ramda'
+import { compose, pathOr, pipe, identity, filter } from 'ramda'
 import withDataLoader from 'core/hocs/withDataLoader'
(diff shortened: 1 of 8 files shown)


src/app/plugins/kubernetes/helpers/clusterizedDataLoader.js
```diff
@@ -9,19 +9,33 @@ export const mapByClusterId = contextKey => ({ context = {}, params: { clusterId
...
+      ),
+    }),
+    // Load the data (the loader will receive a `params.clusterId` prop)
+    withDataLoader({ [key]: loaderFn }),
+    // Map data from context to the "data" prop to use it in the view
+    withDataMapper({
+      clusters:
+        pipe(
+          pathOr([], ['context', 'clusters']),
+          when(always(masterNodeClusters), filter(x => x.hasMasterNode)),
```
brennancheung: This is a bit confusing for people who aren't familiar with these.  Probably can be written simpler with just plain JS.

PR #359, opened 2019-08-19
Improvements on data loading, caching, updating and error handling

xantorres:
Revamped data loading/updating process by allowing cache indexization and added new helpers, hooks, and proper error handling


src/api-client/ApiClient.js
```diff
@@ -11,6 +11,19 @@ import Qbert from './Qbert'
...
 
 class ApiClient {
+  static init (options = {}) {
+    this.instance = new this(options)
+    return this.instance
+  }
+
+  static getInstance () {
+    const { instance } = this
+    if (!instance) {
```
brennancheung: What about just initializing it if it doesn't exist?  Are you trying to implement some kind of Singleton pattern here?  It's not clear why this code was added.

src/app/plugins/kubernetes/components/pods/AddDeploymentPage.js
```diff
@@ -1,71 +1,56 @@
...
+import ClusterPicklist from 'k8s/components/common/ClusterPicklist'
+import NamespacePicklist from 'k8s/components/common/NamespacePicklist'
+
+export const AddDeploymentForm = ({ onComplete }) => {
+  const [params, setParams] = useState(emptyObj)
+  const handleClusterChange = useCallback(clusterId => setParams({ clusterId }), [])
+  return (
+    <ValidatedForm onSubmit={onComplete}>
+      <PicklistField
+        DropdownComponent={ClusterPicklist}
```
brennancheung: What about using `children` instead of `DropdownComponent`?

src/app/plugins/kubernetes/components/prometheus/actions.js
```diff
@@ -1,32 +1,43 @@
...
-}
+import { asyncFlatMap, pathOrNull } from 'utils/fp'
+import { map, pathEq, find, pluck, curry, pipe, last, pathOr, prop, propEq } from 'ramda'
+import ApiClient from 'api-client/ApiClient'
+import createContextLoader from 'core/helpers/createContextLoader'
+import createContextUpdater from 'core/helpers/createContextUpdater'
+import { clustersDataKey } from '../infrastructure/actions'
+import { notFoundErr } from 'app/constants'
+
+export const clusterTagsDataKey = 'clusterTags'
```
brennancheung: Are these just to encourage the linter to be able to find things?  I've seen this style in Redux but am on the fence with it because it is very verbose, especially when you have to import it in a bunch of other spots.

src/app/plugins/kubernetes/components/prometheus/actions.js
```diff
@@ -1,32 +1,43 @@
...
 })
 
-const hasMonitoring = cluster => cluster.tags.includes('pf9-system:monitoring')
-
 /* Prometheus Instances */
 
-export const mapPrometheusInstance = ({ clusterUuid, metadata, spec }) => ({
+export const mapPrometheusInstance = curry((clusters, { clusterUuid, metadata, spec }) => ({
   clusterUuid,
+  clusterName: pipe(find(propEq('uuid', clusterUuid)), prop('name'))(clusters),
```
brennancheung: `clusters.find(propEq('uuid', clusterUuid)).name` flows a bit more linearly.

src/api-client/ApiClient.js
xantorres: I pulled off the `apiClient` from the context, now we can use it globally through the singleton instance, which is more practical and clean.

src/api-client/ApiClient.js
```diff
@@ -66,50 +79,30 @@ class ApiClient {
   }
 
   async basicGet (url) {
-    try {
-      const response = await axios.get(url, this.getAuthHeaders())
-      return response.data
-    } catch (err) {
-      console.log(err)
-    }
+    const response = await axios.get(url, this.getAuthHeaders())
```
xantorres: Errors are handled by the dataLoaders/Updaters now

src/app/core/components/cardTable/CardTable.js
```diff
@@ -91,8 +97,8 @@ class CardTable extends React.Component {
     return data.filter(
       ele =>
```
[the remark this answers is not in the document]
xantorres: Yes, there are a lot of things to improve here, but I would like to work on improvements in `CardTable`/`ListTable` components in another PR, this one is bloated enough

src/app/core/components/cardTable/CardTableToolbar.js
```diff
@@ -42,6 +51,14 @@ const styles = theme => ({
   },
 })
 
+const renderRefreshButton = moize((classes, onRefresh) =>
+  onRefresh && <Tooltip title="Refresh list">
+    <i className={clsx(classes.button, 'fas fa-fw fa-lg fa-sync')}
```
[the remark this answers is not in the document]
xantorres: It works, I changed it, thanks

src/app/plugins/kubernetes/components/pods/AddDeploymentPage.js
xantorres: I can't use children because I need to pass some props that are declared inside PicklistField (basically state for the Tooltip), so the dropdown component (whichever it is) needs to be instantiated there. 
But actually, I'm thinking about using the `withInfoTooltip` HoC in each one of the custom Picklists, with some adjustments to address the issue with the Tooltip getting stuck when opening the list of items. This way we would get rid of using the PicklistField directly.

src/app/plugins/kubernetes/components/prometheus/actions.js
xantorres: this is a way of having centralized constants instead of strings, which are harder to refactor, and also a way to force developers to import the constants from the action files, otherwise the loaders wouldn't get initialized at all.

src/app/plugins/kubernetes/components/prometheus/actions.js
xantorres: agreed, but perhaps more unsafe, as accessing `.name` directly would cause a uncaught thrown error if the prop is not found, not sure if that's ok

src/app/core/components/listTable/ListTableToolbar.js
```diff
@@ -18,7 +18,9 @@ const toolbarStyles = theme => ({
   root: {
     paddingRight: theme.spacing(1),
     color: theme.palette.grey[600],
-
+    '& .MuiOutlinedInput-root': {
```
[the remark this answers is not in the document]
xantorres: This is a specific override for inputs inside the `ListTableToolbar`, so in this case the only solution was either this or pass down a `classes` object through props to modify all the MuiOutlinedInputs in the toolbar, which I think is way more cumbersome.

src/app/core/components/validatedForm/PicklistField.js
```diff
@@ -1,36 +1,35 @@
...
-const PicklistField = React.forwardRef(({ id, info, placement, onChange, label, value, showNone, classes, hasError, errorMessage, title, options, className, ...restProps },
-  ref) => {
+const PicklistField = React.forwardRef(({
+  DropdownComponent, id, info, placement, label, value, showNone,
+  hasError, errorMessage, options, ...restProps
+}, ref) => {
   const [open, setOpen] = React.useState(false)
-  const openTooltip = () => setOpen(true)
-  const closeTooltip = () => setOpen(false)
+  const openTooltip = useCallback(() => setOpen(true), [])
```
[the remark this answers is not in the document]
xantorres: `useCallback` advantage is not just for using it in the hooks dependencies, but also to prevent unnecessary re-renders as a new function would be created on every render, the children components using this function would be re-rendered as well as the react engine would detect changes in the props.

src/app/core/helpers/createCRUDComponents.js
```diff
@@ -86,63 +77,48 @@ const createCRUDComponents = options => {
...
-      </React.Fragment>
-    )
-
-    if (loaders) {
-      if (mappers) {
-        StandardListPage = withDataMapper(mappers)(StandardListPage)
-      }
-      StandardListPage = withDataLoader(loaders)(StandardListPage)
+    return () => {
+      const [data, loading, reload] = useDataLoader(dataKey)
```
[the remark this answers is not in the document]
xantorres: I was about to use an object instead of an array, but as we really only need these 3 variables right now, it seems better to use the array syntax as it allows us to use custom variable names without the need of using aliases, this is a big advantage when using multiple dataLoaders in the same components.

src/app/core/helpers/createContextLoader.js
```diff
@@ -0,0 +1,149 @@
+import {
```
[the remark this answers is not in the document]
xantorres: I wouldn't worry too much about having to import stuff from ramda and so, as most of the IDE's nowadays handle imports automatically and even hide them when you are opening the file, even sublime text can do that with some plugin.
I prefer to have a bloated bunch of import and a clean code rather than a lot of repeated ugly `R.` all over the place, making the code harder to read

src/app/core/hocs/withDataMapper.js
```diff
@@ -1,9 +1,15 @@
...
+import { AppContext } from 'core/AppProvider'
+import { dataContextKey } from 'core/helpers/createContextLoader'
 
-const withDataMapper = mappers => Component => withAppContext(props => {
-  return <Component {...props} data={mapObjIndexed(mapper => mapper(props), mappers)} />
-})
+const withDataMapper = mappers => Component => props => {
+  const { getContext } = useContext(AppContext)
+  const mappedData = mapObjIndexed(mapper => {
+    const dataContextMapper = pipe(prop(dataContextKey), mapper)
```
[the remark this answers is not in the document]
xantorres: I'd love to use the babel plugin, but for now is still just a proposal and it's better not to rely on proposals as they are subject to change and then it would force us to refactor all our code.

src/app/core/hooks/useDataLoader.js
```diff
@@ -0,0 +1,54 @@
...
+import { emptyArr } from 'utils/fp'
+import { isEmpty } from 'ramda'
+import { ToastContext } from 'core/providers/ToastProvider'
+import { AppContext } from 'core/AppProvider'
+import { getContextLoader } from 'core/helpers/createContextLoader'
+import { memoizedDep } from 'utils/misc'
+
+const useDataLoader = (key, params, refetchOnMount = false) => {
+  const refetching = useRef(refetchOnMount)
+  const accPromises = useRef([])
```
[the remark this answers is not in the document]
xantorres: `accPromises` is a FIFO buffer of promises, this is to prevent issues in the cases two or more subsequent data loading requests are performed with different params, and the first one didn't have time to finish yet, and yes, this happens more often than it seems

PR #360, opened 2019-09-02
Params hooks, fix app catalog and other fixes

xantorres:
* Delegating data sorting logic to the loader actions via `sortWith` property when using `createCRUDActions()` or `createContextLoader()`) rathen than `ListTable`/`CardTable` components
* Added a hook for handling params (`useParams`) passed to the data loaders and a specific hook creator for loading params syncronized with the user settings (`createUsePrefParamsHook(storeKey, userPreferencesKeys)`)
* Fixed cached data retrieving and filtering issues when the server was not returning the params used for indexing the data or if forgetting to add the param manually after fetching (eg: `clusterId`)
* Fix issue with data loading memoized promises not being reused while a fetch action was in progress due to a new function reference being created on every render (specifically, the `onError` handler in the `useDataLoader` hook)
* Increased default `axios` timeout from 30 to 120 seconds
* Fixed App Catalog pages
* Other minor fixes and refactoring


src/app/core/helpers/createCRUDComponents.js
```diff
@@ -98,30 +104,40 @@ const createCRUDComponents = options => {
...
+    const usePrefParams = createUsePrefParamsHook(name, listTablePrefs)
     // ListPage
     return () => {
-      const [data, loading, reload] = useDataLoader(loaderFn)
+      const { params, getParamsUpdater } = usePrefParams(defaultParams)
+      const [data, loading, reload] = useDataLoader(loaderFn, params)
       return <>
-        <ListContainer data={data} loading={loading} reload={reload} />
+        <ListContainer
+          getParamsUpdater={getParamsUpdater}
```
brennancheung: Seems like we're doing a bit of prop drilling and we might be repeating ourselves all the time with these params passed into `ListContainer`.  I wonder if there is a cleaner way.  This seems like the kind of thing that will mostly be copy paste for each new List Table we have to create.

Then again, it's more explicit and there's less confusing abstractions.  This gives us a little more flexibility at the expense of verbosity.  I'm on the fence with this.  I could go either way.

src/app/core/helpers/createContextLoader.js
```diff
@@ -115,41 +122,51 @@ const createContextLoader = (key, dataFetchFn, options = {}) => {
...
         // Update cachedParams so that we know this query has already been resolved
-        const updateParams = pipe(arrayIfNil, contextLoaderFn._invalidatedCache
+        const updateParams = pipe(arrayIfNil, contextLoaderFn._invalidatedCache || refetch
           ? always(of(providedIndexedParams)) // Reset the params array if cache has been invalidated
           : append(providedIndexedParams))
 
         // Perfom the context update operations
         await setContext(pipe(
-          over(dataLens, pipe(cleanExistingItems, addNewItems)),
+          over(dataLens, pipe(cleanPrevItems, upsertNewItems)),
```
brennancheung: Nice, very clean.

src/app/core/hooks/useParams.js
```diff
@@ -0,0 +1,82 @@
+import { emptyObj } from 'utils/fp'
+import { useMemo, useState, useCallback } from 'react'
+import moize from 'moize'
+import { isEmpty, pick, zipObj } from 'ramda'
+import { useScopedPreferences } from 'core/providers/PreferencesProvider'
+
+/**
+ * Function that will update the params object with a new set of provided key/values,
```
brennancheung: There's not enough context to understand what this means.

Add a "why" describing the motivation for this abstraction.  What would be the alternative?  Provide an example to ground it.

Contrast and concretization are 2 great ways to make code (and writing) more understandable.

Same for the other functions defined below.

Perhaps some kind of Markdown should be added to the codebase to teach others how to use these.

src/app/plugins/kubernetes/components/infrastructure/actions.js
```diff
@@ -106,11 +108,18 @@ export const clusterActions = createCRUDActions(clustersDataKey, {
     // have references to the deleted cluster.
     loadCombinedHosts.invalidateCache()
   },
-  dataMapper: (items, { masterNodeClusters, prometheusClusters }) => pipe(
+  uniqueIdentifier: 'uuid',
+  dataMapper: (items,
+    { masterNodeClusters, appCatalogClusters, prometheusClusters }) => pipe(
     masterNodeClusters ? filter(hasMasterNode) : identity,
-    prometheusClusters ? filter(hasPrometheusEnabled) : identity
+    prometheusClusters ? filter(hasPrometheusEnabled) : identity,
```
brennancheung: What about `const filterIf = (items, fn) = items ? filter(fn) : identity`?

That would make the code shorter and quite readable and almost sound like plain English:

`filterIf(masterNodeClusters, hasMasterNode)`

Beginning programmers don't really understand the concept—or need—for an identity function so this will hide that from them.

src/app/utils/fp.js
```diff
@@ -21,6 +21,18 @@ export const noop = () => {}
 // Works for arrays and strings.  All other types return false.
 export const notEmpty = arr => !!(arr && arr.length)
 
+export const mapKeys = curry((cb, obj) => {
```
brennancheung: Give an example of inputs -> outputs a comment above this function.

What gets passed to the `cb` function?

Provide enough information so someone can see how to use this without fully understanding the implementation of this function.

src/app/utils/fp.js
```diff
@@ -231,25 +243,34 @@ export const condLiteral = (...conds) => value => {
...
 // Remove an item from an array using a predicateFn
 export const removeWith = curry((predicateFn, arr) =>
-  remove(findIndex(predicateFn, arr), 1, arr)
+  remove(findIndex(predicateFn, arr), 1, arr),
 )
 
+// Insert or update items in bulk
+export const upsertAllBy = curry((predicateFn, newItems, targetArr) => {
+  const groupedValues = values(groupBy(predicateFn, [...targetArr, ...newItems]))
+  return groupedValues.map(
```
brennancheung: It's really hard to know what the shape/type of the data are going through these functions.  Definitely add an example input/output in the comment above the function.  Possibly within the function too.

src/app/utils/fp.js
xantorres: I will remove this function as it is not being used (I used it for some stuff that then I refactored)

PR #361, opened 2019-09-06
Add aws cluster

brennancheung:
Implementing the Create AWS Cluster functionality.

We're also moving to a design that has the user first choose their type of cluster.  This greatly simplifies the UX because only the relevant fields are presented.


src/app/plugins/kubernetes/components/infrastructure/AddAwsClusterPage.js
```diff
@@ -0,0 +1,304 @@
...
+        <TextField
+          id="name"
+          label="name"
+          info="Name of the cluster"
+        />
+        <PicklistField
+          DropdownComponent={CloudProviderPicklist}
+          id="cloudProvider"
+          label="Cloud Provider"
+          options={cloudProviderOptions}
```
xantorres: You don't need to pass the `cloudProviderOptions` to the Picklist component as this is already being done inside of the component. If you want to filter the cloud providers by the ones of type `aws`, just add an additional prop to the `CloudProviderPicklist` component and do the filtering there.

src/app/constants.js
```diff
@@ -8,6 +8,7 @@ export const dashboardUrl = `${appUrlRoot}/kubernetes/`
 export const allKey = '__all__'
 export const noneKey = '__none__'
 export const listTablePrefs = ['visibleColumns', 'columnsOrder', 'rowsPerPage', 'orderBy', 'orderDirection']
+export const k8sLink = path => `${appUrlRoot}/kubernetes${path}`
```
xantorres: shouldn't this be a helper rather than a constant? Looks like `misc.js` or even `http.js` files would be a better place for this.

src/app/plugins/kubernetes/components/infrastructure/AddAwsClusterPage.js
brennancheung: I started with loading the data in the AwsClusterPage first and then added a dedicated `CloudProviderPicklist` later on.  I'll remove this code now that it is not needed here.

src/app/constants.js
brennancheung: Seems like the right place but I'll change it to be a hard coded `k8sPrefix` instead to make it more like a constant and less like a function.

src/app/core/helpers/createContextLoader.js
```diff
@@ -21,24 +21,43 @@ const arrayIfEmpty = when(isEmpty, always(emptyArr))
...
+ * @property {function|string} [fetchErrorMessage] Custom message to display after
+ * an error has been thrown
  */
 
 /**
  * Returns a function that will use context to load and cache values
  * @typedef {function} createContextLoader
  * @function createContextLoader
- * @param {string} key Key on which the resolved value will be cached
+ * @param {string} key Name of key field (ex: 'clusterId') to resolved cached value
```
xantorres: This is wrong, the key should stand for the name of the entity, and it is the property name that will be used to cache the values in the context.

src/app/utils/misc.js
```diff
@@ -52,11 +52,12 @@ export const castBoolToStr = (t = 'yes', f = 'no') => value => value ? t : f
...
 
 /**
- * Memoizes an async function so that concurrent calls to the same function (with the same params) will return the same promise
+ * Memoizes an async function to prevent so that concurrent calls to the same function
+ * (with the same params) will return the same promise.
  * @param {function} asyncFn Function that will be memoized until it gets resolved
  * @returns {function}
  */
-export const singlePromise = asyncFn => {
+export const memoizePromise = asyncFn => {
```
xantorres: Good naming, but the comment is a bit confusing, I think you should remove the "to prevent" part

src/app/core/helpers/createContextLoader.js
brennancheung: The term `key` is ambiguous.  Does it mean the `clusterId` or the actual uuid (`xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx`)?

Also, why all the complexity with key / entity stuff, trimming off the `s` at the end.  I think we can just get rid of all that.  Ruby on Rails tried to do that with all the noun inflection / pluralization and it caused a bunch of issues with not knowing what keys and indexes actually were and having to write tons of code to support it just to make it sound more English like.

src/app/utils/misc.js
brennancheung: I left out part of what I was trying to say.  I'm taking another pass at it now.

src/app/core/helpers/createContextLoader.js
xantorres: It doesn't mean either of both, is just the key used to store the values in the context, we can call it `dataKey` or `storeKey`, if you will.

The `entityName` is just "humanizing" the provided `key` value as it is going to be used in console and Toast messages, which 90% of the times is just separating the provided `key` with spaces and removing the last "s", eg `appProviders` -> `App Providers`.  
So it does automatically for you, but you can always overwrite it with your own custom value, I did that with the `repositories` for instance, as it was being generated as "Repositorie", so I manually passed the correct value. But 90% of the times is fine leaving that unset so we reduce the overall boilerplate.

PR #366, opened 2019-09-19
[WIP] MultiSelect Component

murilovmachado:
Our current `MultiSelect` didn't match our old UI `MultiSelect` component at all. So I'm rebuilding it to work and look like the previous one, and possibly add new desired features described in the `PMK Feature Parity Checklist` document.

TODO:
- Work on the visuals
  - Fixed height for the component?
  - Option styles
  - Search input styles

Screen Shot 2019-09-18 at 21 57 09
Storybook Screenshot

### DIFF
--- a/src/app/core/components/MultiSelect.js
+++ b/src/app/core/components/MultiSelect.js
@@ -26,11 +26,18 @@ const useStyles = makeStyles(theme => createStyles({
   },
 }))
 
-const MultiSelect = ({ label, options, values, onChange, maxOptions }) => {
+const MultiSelect = ({ label, options, values, onChange, maxOptions, sortSelectedFirst }) => {
   const classes = useStyles()
-  const limitOptions = (options) => maxOptions ? options.slice(0, maxOptions) : options
 
-  const [visibleOptions, setVisibleOptions] = useState(limitOptions(options))
+  const adjustOptions = (options) => {
+    const sortBySelected = (a, b) => values.includes(b.value) - values.includes(a.value)
+    const sortedOptions = sortSelectedFirst ? options.sort(sortBySelected) : options
+    const limitedOptions = maxOptions ? sortedOptions.slice(0, maxOptions) : sortedOptions
+
+    return limitedOptions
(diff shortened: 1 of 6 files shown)


src/app/core/components/MultiSelect.js
```diff
@@ -1,55 +1,82 @@
...
-}
+        <Box className={classes.container}>
+            <SearchField />
+            {optionsWithStatus.map((option) => (
+                <Option
+                    classes={classes}
+                    key={option.value}
+                    label={option.label}
+                    value={option.value}
+                    checked={option.checked}
```
xantorres: Instead of mapping all the provided options to add the `checked` status, I think it would better to just use `values.includes(option.value)` here.

xantorres:
Run `npm run lint`, there are some linting errors, it seems

src/app/core/components/MultiSelect.js
```diff
@@ -1,69 +1,90 @@
...
 
-const styles = theme => ({
-  root: {
-    display: 'flex',
-    flexWrap: 'wrap',
-  },
-  formControl: {
-    marginTop: theme.spacing(1),
-    minWidth: 200,
+const MAX_OPTIONS = 9
```
brennancheung: Don't limit by default, especially with a hard coded value.  Let that be an additional prop to the component.

src/app/core/components/MultiSelect.js
```diff
@@ -1,69 +1,131 @@
...
 })
 
 MultiSelect.propTypes = {
-  name: PropTypes.string.isRequired,
   label: PropTypes.string.isRequired,
   options: PropTypes.arrayOf(optionPropType).isRequired,
-  value: PropTypes.arrayOf(PropTypes.string),
+  values: PropTypes.arrayOf(PropTypes.string),
   onChange: PropTypes.func,
+  maxOptions: PropTypes.bool,
```
brennancheung: This is an `int` / `number` not a `bool` isn't it?

PR #370, opened 2019-09-20
[WIP] Style MultiSelect

murilovmachado:
This PR:
- Apply styling to MultiSelect to make it look the old UI
- Changes `maxOptions` to no limit the number of options, but instead, it limits the component's height, but all the options could be viewed by scrolling.

TODO:
- Move some styling to `defaultTheme.js`

Screen Shot 2019-09-20 at 14 25 27
Storybook image

### DIFF
--- a/src/app/core/components/MultiSelect.js
+++ b/src/app/core/components/MultiSelect.js
@@ -7,6 +7,8 @@ import InputAdornment from '@material-ui/core/InputAdornment'
 import OutlinedInput from '@material-ui/core/OutlinedInput'
 import FormControlLabel from '@material-ui/core/FormControlLabel'
 import Checkbox from '@material-ui/core/Checkbox'
+import CheckBoxOutlineBlankIcon from '@material-ui/icons/CheckBoxOutlineBlank'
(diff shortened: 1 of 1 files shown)


src/app/core/components/MultiSelect.js
```diff
@@ -96,24 +138,43 @@ const SearchField = ({ onSearchChange, onHitEnter }) => {
...
+        className={classes.checkbox}
+        icon={<CheckBoxOutlineBlankIcon className={classes.checkboxSize} />}
+        checkedIcon={<CheckBoxIcon className={classes.checkboxSize} />}
+        {...checkboxProps}
+      />
+    }
+    classes={{ root: classes.formControlLabelRoot, label: classes.optionLabel }}
   />
 
+const getOptionsHeight = (maxOptions) => maxOptions * 28
```
brennancheung: The hard coded 28 might not work if we change other design aspects in the future, but I'm not sure the best way to go about this without investing a bunch more time into it.  We can always change it later to be more robust but I think this is fine for now.

PR #372, opened 2019-09-23
App deployment/download + refactoring and fix warnings

xantorres:
* App deployment feature
* App download feature
* App detail back button
* Move async helpers to its own file (async.js) and refactor functions to be curried
* Minor refactoring and warnings fixes

### DIFF
--- a/src/app/plugins/kubernetes/components/apps/actions.js
+++ b/src/app/plugins/kubernetes/components/apps/actions.js
@@ -58,6 +58,10 @@ export const appVersionLoader = createContextLoader(appVersionsCacheKey, async (
     ...item,
     version: pathStr('attributes.version', item),
     appVersion: pathStr('attributes.app_version', item),
+    versionLabel: [
+      pathStr('attributes.version', item),
+      moment(pathStr('attributes.created', item)).format('MMM DD, YYYY'),
+    ].join(' - '),
     downloadLink: pathStr('attributes.urls.0', item),
   })),
   defaultOrderBy: 'version',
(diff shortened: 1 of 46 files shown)


src/app/plugins/kubernetes/components/apps/actions.js
```diff
@@ -37,9 +39,9 @@ export const singleAppLoader = createContextLoader(singleAppCacheKey, async ({ c
       return {
         ...item,
         logoUrl: icon ? pathJoin(monocularUrl, icon) : `${imageUrlRoot}/default-app-logo.png`,
-        home: item.relationships.chart.data.home,
-        sources: item.relationships.chart.data.sources,
-        maintainers: item.relationships.chart.data.maintainers,
+        home: path(['relationships', 'chart', 'data', 'home'], item),
```
brennancheung: I think there's a helper somewhere so you can specify the string `a.b.c.d` instead of using an array.  If not, maybe we should create one.

src/app/utils/async.js
```diff
@@ -0,0 +1,49 @@
...
+ * @example
+ * propsAsync({
+ *   foo: getFoo(),
+ *   boo: getBoo(),
+ * }).then(results => {
+ *   console.log(results.foo, results.boo);
+ * })
+ */
+export const propsAsync = async objPromises => {
+  const promises = Object.values(mapObjIndexed(async (promise, key) => {
```
brennancheung: What about shorthand syntax?

PR #374, opened 2019-09-24
Logging table

murilovmachado:
This PR uses stubbed data (no fetching for now), and implements:
- Table to display logging data
- Columns rendered according to the designs
- Batch action buttons according to the designs

Screen Shot 2019-09-24 at 15 57 57

### DIFF
--- a/src/app/core/components/listTable/ListTableBatchActions.js
+++ b/src/app/core/components/listTable/ListTableBatchActions.js
@@ -4,15 +4,14 @@ import { IconButton, Tooltip } from '@material-ui/core'
 import { withAppContext } from 'core/AppProvider'
 
 // FIXME this should not be accessing the context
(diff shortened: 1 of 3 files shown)


src/app/core/components/listTable/ListTableBatchActions.js
```diff
@@ -4,9 +4,9 @@ import { IconButton, Tooltip } from '@material-ui/core'
 import { withAppContext } from 'core/AppProvider'
 
 // FIXME this should not be accessing the context
-const ListTableRowActions = ({ actionClassName, context, rowActions, selected }) => {
+const ListTableBatchActions = ({ actionClassName, context, batchActions, selected }) => {
   const _selected = selected || []
```
brennancheung: _selected can be moved into the default function parameters.

PR #375, opened 2019-09-24
New Logging and Edit Logging empty pages

murilovmachado:
This PR implements:
- Clicking on `+ New Logging` => Redirects to New Logging Page
- Clicking on `Edit` => Redirects to Edit Logging Page

Screen Shot 2019-09-24 at 20 46 45
Screen Shot 2019-09-24 at 20 46 50

### DIFF
--- a/src/app/core/components/buttons/CreateButton.js
+++ b/src/app/core/components/buttons/CreateButton.js
@@ -8,7 +8,6 @@ const styles = theme => ({
     borderRadius: 2,
     textTransform: 'none',
     height: 40,
-    boxShadow: '0 3px 1px -2px rgba(0, 0, 0, 0.11), 0 2px 2px 0 rgba(0, 0, 0, 0.11), 0 1px 5px 0 rgba(0, 0, 0, 0.2)',
   }
 })
 
--- a/src/app/core/components/buttons/CreateButton.js
+++ b/src/app/core/components/buttons/CreateButton.js
@@ -7,6 +7,8 @@ const styles = theme => ({
     margin: theme.spacing(1),
     borderRadius: 2,
     textTransform: 'none',
+    height: 40,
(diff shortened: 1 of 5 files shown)


src/app/core/components/buttons/CreateButton.js
```diff
@@ -7,6 +7,8 @@ const styles = theme => ({
     margin: theme.spacing(1),
     borderRadius: 2,
     textTransform: 'none',
+    height: 40,
+    boxShadow: '0 3px 1px -2px rgba(0, 0, 0, 0.11), 0 2px 2px 0 rgba(0, 0, 0, 0.11), 0 1px 5px 0 rgba(0, 0, 0, 0.2)',
```
xantorres: They asked for removal of the shadows to match the old ui, so better don't re-add them here (we removed it in the `defaultTheme` file)
image

src/app/plugins/kubernetes/components/logging/LoggingListPage.js
```diff
@@ -163,18 +172,21 @@ const data = [
   },
 ]
 
-const LoggingPage = () => {
+const LoggingListPage = withRouter(({ history }) => {
```
xantorres: You can use `useReactRouter` hook instead of this ugly HoC here

src/app/plugins/kubernetes/components/logging/LoggingListPage.js
```diff
@@ -40,6 +46,9 @@ const useStyles = makeStyles(theme => createStyles({
   },
 }))
 
+const EDIT_URL = '/ui/kubernetes/logging/edit'
+const ADD_URL = '/ui/kubernetes/logging/add'
```
xantorres: We haven't been using uppercased constants in the code, it would be nice to follow our current convention so that we have a consistent code styling across the codebase.
Also keep in mind that we have a constants file named `constants.js`, in case you have constants that could potentially be used in more than one file.

PR #376, opened 2019-09-25
AWS Cluster Create AZs

brennancheung:
* Add field for choosing AWS Availability Zones
* Change `ValidatedForm` to also be able to accept a `renderProp`
* Added some other fields
* Added a cluster template preset picklist that to conveniently set multiple fields at once

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/AddAwsClusterPage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/AddAwsClusterPage.js
@@ -5,6 +5,7 @@ import FormWrapper from 'core/components/FormWrapper'
 // import KeyValuesField from 'core/components/validatedForm/KeyValuesField'
 // import NodesChooser from './NodesChooser'
 import AwsAvailabilityZoneChooser from './AwsAvailabilityZoneChooser'
+import AwsRegionFlavorPicklist from './AwsRegionFlavorPicklist'
 import CloudProviderPicklist from 'k8s/components/common/CloudProviderPicklist'
 import CloudProviderRegionPicklist from 'k8s/components/common/CloudProviderRegionPicklist'
 import PicklistField from 'core/components/validatedForm/PicklistField'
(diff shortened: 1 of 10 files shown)


src/app/plugins/kubernetes/components/infrastructure/actions.js
```diff
@@ -162,6 +162,18 @@ export const loadCloudProviderDetails = createContextLoader(
...
 
+export const loadCloudProviderRegionDetails = createContextLoader(
+  'cloudProviderRegionDetails',
+  async ({ cloudProviderId, cloudProviderRegionId }) => {
+    console.log(`loadCloudProviderRegionDetails(${cloudProviderId}, ${cloudProviderRegionId})`)
+    const response = await qbert.getCloudProviderRegionDetails(cloudProviderId, cloudProviderRegionId)
+    return response
+  },
+  {
+    indexBy: 'cloudProviderRegionId',
```
xantorres: By default `createContextLoader` uses  `"id"` as `uniqueIdentifier`, but in this case this won't work because no id is returned from the server.
So I think the best idea in this case would be to add a custom ID so that it can be used to prevent duplicates in the cache. 
The `id` would be a composite identifier with `cloudProviderId` and `cloudProviderRegionId`. Ie:
```
return { id:  `${cloudProviderId}-${cloudProviderRegionId}`, ...response }
```
Moreover, `indexBy` needs to contain `cloudProviderId`, otherwise we could end up retrieving multiple items when fetching the data from the cache. Ie `indexBy: ['cloudProviderId', 'cloudProviderRegionId'],`

brennancheung: I've modified the response to inject the `id` parameter for `uniqueIdentifier` to work.

brennancheung:
Decided to add a few more fields to this PR while it is still open.

Screenshot 2019-09-30 18 01 52

src/app/plugins/kubernetes/components/infrastructure/AwsRegionFlavorPicklist.js
```diff
@@ -0,0 +1,42 @@
...
+  cloudProviderId: PropTypes.string,
+  cloudProviderRegionId: PropTypes.string,
+  initialValue: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  onChange: PropTypes.func,
+  ...ValidatedFormInputPropTypes,
+}
+
+export default compose(
+  withInfoTooltip,
+  withFormContext,
```
xantorres: Same here

src/app/plugins/kubernetes/components/infrastructure/AwsAvailabilityZoneChooser.js
```diff
@@ -0,0 +1,45 @@
...
+  cloudProviderId: PropTypes.string,
+  cloudProviderRegionId: PropTypes.string,
+  initialValue: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  onChange: PropTypes.func,
+  ...ValidatedFormInputPropTypes,
+}
+
+export default compose(
+  withInfoTooltip,
+  withFormContext,
```
xantorres: If you use this field along with `PicklistField` you don't need to use `withFormContext` here, as it is already being done in `PicklistField`

src/app/plugins/kubernetes/components/infrastructure/AddAwsClusterPage.js
```diff
@@ -26,6 +34,57 @@ const AddAwsClusterPage = () => {
...
+  ]
+
+  // The template picker allows the user to fill out some useful defaults for the fields.
+  // This greatly simplifies the number of fields that need to be filled out.
+  // Presets are as follows:
+  // small (single dev) - 1 node master + worker - select instance type (default t2.small)
+  // medium (internal team) - 1 master + 3 workers - select instance (default t2.medium)
+  // large (production) - 3 master + 5 workers - no workload on masters (default t2.large)
+  const handleTemplateChoice = ({ setWizardContext, setFieldValue }) => option => {
+    const options = {
```
xantorres: Maybe you could declare `options` and other static variables outside of the render function, so that we don't recreate these objects every time, the render method would look a bit cleaner also.

PR #377, opened 2019-09-27
Use helper methods to create Logging pages: list, add and edit

murilovmachado:
The main point in this PR is to refactor `Logging` to use helper methods and be prepared for the next step, that would be fetching and caching real data. More specifically, this PR:
- Creates `LoggingStub` to be used by `Qbert` while fetching real data is not implemented
- Refactor `LoggingListPage` to use helper method `createCRUDComponents`
- Refactor `LoggingAddPage` to use helper method `createAddComponents`
- Refactor `LoggingEditPage` to use helper method `createUpdateComponents`
- Creates `LoggingForm` to be used by `LoggingAddPage` and `LoggingEditPage`
- Refactor `createCRUDComponents` to support `batchActions`
- Refactor `ListTableBatchActions` to standardize custom batchActions buttons

### DIFF
--- a/src/app/core/components/listTable/TitleAndAddButton.js
+++ b/src/app/core/components/listTable/TitleAndAddButton.js
@@ -1,0 +1,56 @@
+import React, { Fragment, useState, useCallback, useMemo } from 'react'
+import PropTypes from 'prop-types'
+import useReactRouter from 'use-react-router'
+import { makeStyles, createStyles } from '@material-ui/styles'
+import Box from '@material-ui/core/Box'
+import CreateButton from 'core/components/buttons/CreateButton'
+
+const useStyles = makeStyles(theme => createStyles({
+  container: {
+    display: 'flex',
+    justifyContent: 'space-between',
+    alignItems: 'center',
+  },
+}))
+
+const TitleAndAddButton = ({ title, addUrl, addText, renderAddDialog, reload }) => {
+  const { history } = useReactRouter()
+  const classes = useStyles()
+  const [addDialogOpen, setAddDialogOpen] = useState(false)
+
+  const handleAddDialogClose = useCallback(() => {
+    setAddDialogOpen(false)
+    reload()
+  }, [reload])
+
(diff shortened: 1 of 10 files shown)


src/app/plugins/kubernetes/components/logging/LoggingStub.js
```diff
@@ -0,0 +1,130 @@
+/* Stub data while not fetching from real API */
```
xantorres: We have our own "stub" or dummy server to put some example data. Take a look at `server` folder and add the corresponding models and routes there.

src/app/core/components/listTable/TitleAndAddButton.js
```diff
@@ -0,0 +1,56 @@
...
+  const handleAddDialogClose = useCallback(() => {
+    setAddDialogOpen(false)
+    reload()
+  }, [reload])
+
+  const addButton = useMemo(() => (
+    <CreateButton onClick={() => renderAddDialog ? setAddDialogOpen(true) : history.push(addUrl)}>
+      {addText}
+    </CreateButton>
+  ), [history])
```
xantorres: both `addText` and `addUrl` should probably be added to the list of dependencies (even though its unlikely they will ever change).

src/app/plugins/kubernetes/components/logging/LoggingEditPage.js
```diff
@@ -1,5 +1,19 @@
...
+import createUpdateComponents from 'core/helpers/createUpdateComponents'
+import LoggingForm from './LoggingForm'
+import ApiClient from 'api-client/ApiClient'
 
-const LoggingEditPage = () => <h1>Edit Logging</h1>
+const { qbert } = ApiClient.getInstance()
+
+const options = {
+  FormComponent: LoggingForm,
+  updateFn: () => {},
```
xantorres: Add TODO comments in unfinished features, just to make explicit that something is missing

PR #380, opened 2019-10-02
More AWS cluster create fields

brennancheung:
Adding more fields to the AWS cluster create wizard.

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/AddAwsClusterPage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/AddAwsClusterPage.js
@@ -8,6 +8,10 @@ import AwsAvailabilityZoneChooser from './AwsAvailabilityZoneChooser'
 import AwsRegionFlavorPicklist from './AwsRegionFlavorPicklist'
 import CloudProviderPicklist from 'k8s/components/common/CloudProviderPicklist'
 import CloudProviderRegionPicklist from 'k8s/components/common/CloudProviderRegionPicklist'
+import AwsClusterSshKeyPicklist from './AwsClusterSshKeyPicklist'
+import ClusterDomainPicklist from './ClusterDomainPicklist'
+import CheckboxField from 'core/components/validatedForm/CheckboxField'
+import KeyValuesField from 'core/components/validatedForm/KeyValuesField'
 import PicklistField from 'core/components/validatedForm/PicklistField'
 import TextField from 'core/components/validatedForm/TextField'
 import ValidatedForm from 'core/components/validatedForm/ValidatedForm'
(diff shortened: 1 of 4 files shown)


src/app/plugins/kubernetes/components/infrastructure/ClusterDomainPicklist.js
```diff
@@ -0,0 +1,41 @@
...
+  id: PropTypes.string.isRequired,
+  cloudProviderId: PropTypes.string,
+  cloudProviderRegionId: PropTypes.string,
+  initialValue: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  onChange: PropTypes.func,
+  ...ValidatedFormInputPropTypes,
+}
+
+export default compose(
+  withInfoTooltip,
```
xantorres: This is not needed, as `PicklistField` already uses `Tooltip`

src/app/plugins/kubernetes/components/infrastructure/AwsClusterSshKeyPicklist.js
```diff
@@ -0,0 +1,41 @@
...
+  id: PropTypes.string.isRequired,
+  cloudProviderId: PropTypes.string,
+  cloudProviderRegionId: PropTypes.string,
+  initialValue: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  onChange: PropTypes.func,
+  ...ValidatedFormInputPropTypes,
+}
+
+export default compose(
+  withInfoTooltip,
```
xantorres: Not needed

src/app/plugins/kubernetes/components/infrastructure/AddAwsClusterPage.js
```diff
@@ -208,18 +437,31 @@ const AddAwsClusterPage = () => {
...
-                  {/* Disable workloads on master nodes */}
-                  {/* Enable spot workers */}
-                  {/* Percent of worker nodes as spot instances */}
-                  {/* Spot price */}
+                  {/* TODO */}
+                </ValidatedForm>
+              </FormWrapper>
+            </WizardStep>
+
+            <WizardStep stepId="network" label="Network Info">
```
xantorres: This file is getting huge, wouldn't be better to refactor out the steps contents into separate files, and put everything in a `/AddAwsCluster` folder or something?

brennancheung: We should probably refactor and move a bunch of the infrastructure stuff into subfolders.  Let's save that for a future PR.  For now, it's easier to work with when it is all in one place.

src/app/plugins/kubernetes/components/infrastructure/ClusterDomainPicklist.js
brennancheung: I'll make the changes here and in the other spots.

PR #381, opened 2019-10-02
Fix top add button

murilovmachado:
This PR fixes the `Add Button` alignment issue by using `TopExtraContent`.
Also, `LoggingListPage` is a `Tab` in `LoggingPageIndex`, so there's no need to pass in a `innerTitle` parameter.


src/app/core/components/listTable/TopAddButtonAndDialog.js
```diff
@@ -0,0 +1,42 @@
+import React, { Fragment, useState, useCallback, useMemo } from 'react'
+import PropTypes from 'prop-types'
+import useReactRouter from 'use-react-router'
+import CreateButton from 'core/components/buttons/CreateButton'
+import TopExtraContent from 'core/components/TopExtraContent'
+
+const TopAddButtonAndDialog = ({ addUrl, addText, renderAddDialog, reload }) => {
```
xantorres: I don't see why this component is inside the `listTable` folder. As I understand, this button is (or should be) totally decoupled from the `ListTable` component, so probably it should be moved elsewhere.

src/app/core/components/listTable/TopAddButtonAndDialog.js
```diff
@@ -0,0 +1,42 @@
...
+  const { history } = useReactRouter()
+  const [addDialogOpen, setAddDialogOpen] = useState(false)
+
+  const handleAddDialogClose = useCallback(() => {
+    setAddDialogOpen(false)
+    reload()
+  }, [reload])
+
+  const addButton = useMemo(() => (
+    <TopExtraContent>
```
xantorres: Sometimes we want to add more content at the top right side, eg. if you look at the clarity UI:
image
We will have to implement this aswell, so we will need to have some extra flexibility to allow adding custom stuff.

murilovmachado: @xantorres we already have `TopExtraContent` that handles every case. `TopAddButtonAndDialog` was created with the intent of making the code in `ListContainer` cleaner. 

For the scenario you've shown, we can just use `TopExtraContent` directly or create another component that uses it. Worst case scenario, if we want to reuse `TopAddButtonAndDialog`, we should rename it to something more general and make it more flexible. But we can make these changes when they're really needed.

What do you think?

PR #382, opened 2019-10-03
Started guided welcome wizard

kevinylee1991:
First pass at guided welcome wizard, looking for feedback. Current placeholder data makes it look like below:

Screen Shot 2019-10-02 at 5 18 00 PM

### DIFF
--- a/src/app/core/components/expansionPanel/ExpansionPanel.js
+++ b/src/app/core/components/expansionPanel/ExpansionPanel.js
@@ -1,0 +1,83 @@
+import React from 'react'
+import { makeStyles, createStyles } from '@material-ui/styles'
+import PropTypes from 'prop-types'
+import { ExpansionPanel as MDExpansionPanel, ExpansionPanelSummary, ExpansionPanelDetails } from '@material-ui/core'
+import { ExpandMore, ChevronRight } from '@material-ui/icons'
+import FontAwesomeIcon from 'core/components/FontAwesomeIcon'
+
+const useStyles = makeStyles(theme => createStyles({
+  root: {
+    marginBottom: '4px',
+    boxShadow: '0 0 2px 0 rgba(0, 0, 0, 0.2)',
+    '&:before': {
+      background: 'none'
+    },
+    '&.Mui-expanded': {
+      margin: '4px 0px'
+    }
+  },
+  numberedStep: {
+    height: '24px',
+    width: '24px',
+    display: 'flex',
+    justifyContent: 'center',
+    alignItems: 'center',
+    background: '#e6e6e6',
+    borderRadius: '50%',
+    fontSize: '12px',
+    margin: '0px 10px'
+  },
+  completedStep: {
+    background: '#4adf74 !important',
+    color: '#ffffff'
+  },
+  summaryText: {
+    marginLeft: '10px',
+    display: 'flex',
+    fontWeight: '500',
+    alignItems: 'center'
+  },
+  panelDetails: {
+    background: '#f6fafe',
+    border: '1px solid rgba(74, 163, 223, 0.5)',
+    borderLeft: '0px',
+    borderRight: '0px',
+    padding: '40px'
+  }
+}))
+
+const ExpansionPanel = ({ expanded, onClick, children, completed, stepNumber, summary }) => {
+  const classes = useStyles()
+
+  return (
+    <MDExpansionPanel className={classes.root} expanded={expanded} onClick={onClick}>
+      <ExpansionPanelSummary>
+        { expanded
+          ? <ExpandMore />
+          : <ChevronRight />
+        }
+        <div className={classes.summaryText}>
+          { completed
+            ? <div className={`${classes.numberedStep} ${classes.completedStep}`}>
+              <FontAwesomeIcon>check</FontAwesomeIcon>
+            </div>
+            : <div className={classes.numberedStep}>{stepNumber}</div>
+          }
+          <span>{summary}</span>
+        </div>
+      </ExpansionPanelSummary>
+      <ExpansionPanelDetails className={classes.panelDetails}>
+        {children}
+      </ExpansionPanelDetails>
+    </MDExpansionPanel>
+  )
+}
+
+ExpansionPanel.propTypes = {
+  expanded: PropTypes.bool,
+  onClick: PropTypes.func,
+  summary: PropTypes.string,
+  stepNumber: PropTypes.number
+}
+
+export default ExpansionPanel
--- a/src/app/plugins/kubernetes/components/dashboard/ClusterSetupWizard.js
+++ b/src/app/plugins/kubernetes/components/dashboard/ClusterSetupWizard.js
@@ -1,0 +1,125 @@
+import React, { useCallback, useState } from 'react'
+import { except } from 'app/utils/fp'
+import { Typography } from '@material-ui/core'
+import { makeStyles, createStyles } from '@material-ui/styles'
+import ExpansionPanel from 'core/components/expansionPanel/ExpansionPanel'
+import { Link } from 'react-router-dom'
+import FontAwesomeIcon from 'core/components/FontAwesomeIcon'
+import RightAlign from 'core/components/RightAlign'
+
+const useStyles = makeStyles(theme => createStyles({
+  root: {
+    padding: '20px 40px 80px',
+    background: '#f3f3f4',
+    color: '#606060'
+  },
+  title: {
+    fontSize: '18px',
+    margin: '20px 0px',
+    fontWeight: '300'
+  },
+  description: {
+    background: '#ffffff',
+    boxShadow: '0 0 2px 0 rgba(0, 0, 0, 0.2)',
+    borderRadius: '8px',
+    padding: '30px 40px',
+    fontSize: '13px',
+    margin: '16px 0px'
+  },
+  descriptionHeader: {
+    fontSize: '18px',
+    fontWeight: '500',
+    marginBottom: '12px'
+  },
+  button: {
+    display: 'flex',
+    fontSize: '13px',
+    marginLeft: '30px',
+    '&:first-child': {
+      marginLeft: '0px'
+    }
+  },
+  icon: {
+    fontSize: '14px',
+    marginRight: '6px'
+  },
+  link: {
+    textDecoration: 'none',
+    color: '#606060'
+  }
+}))
+
+const ClusterSetupWizard = () => {
+  // Initial state will be handled by a loader later on
+  // Initialize the state with the first panel expanded
+  const [activePanels, setActivePanels] = useState([1])
+  const classes = useStyles()
+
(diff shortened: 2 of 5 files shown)


src/app/core/components/expansionPanel/ExpansionPanel.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react'
+import { withStyles } from '@material-ui/styles'
+import PropTypes from 'prop-types'
+import { ExpansionPanel as MDExpansionPanel, ExpansionPanelSummary, ExpansionPanelDetails, Typography } from '@material-ui/core'
+import { ExpandMore, ChevronRight } from '@material-ui/icons'
+
+const styles = theme => ({
```
brennancheung: Look into using `makeStyles` / `useStyles`.  Should be able to grep for `useStyles` in the codebase to see an example.  It cleans up the component hierarchy in the DOM.

src/app/core/components/icons/TopRightIcons.js
```diff
@@ -0,0 +1,70 @@
...
+})
+
+class TopRightIcons extends React.PureComponent {
+  renderButton = (icon) => {
+    const { classes } = this.props
+
+    return (
+      <div key={icon.key} className={classes.button}>
+        <Link className={classes.link} to={icon.link}>
+          <i className={`${icon.style} ${icon.icon} ${classes.icon}`}></i>
```
brennancheung: There's a `FontAwesomeIcon` component somewhere in our codebase already.

src/app/core/components/icons/TopRightIcons.js
```diff
@@ -0,0 +1,70 @@
...
+      </div>
+    )
+  }
+
+  render () {
+    const { classes, icons } = this.props
+
+    return (
+      <div className={classes.buttonContainer}>
+        { icons.map(icon => this.renderButton(icon)) }
```
brennancheung: Probably easier and cleaner to just use `children` instead of accepting a data structure describing the icons.  Just have `TopRightIcons`.

Ex:

```<TopRightIcons><FontAwesomeIcon ... /><FontAwesomeIcon .../></TopRightIcons>```

File: src/app/plugins/kubernetes/components/dashboard/DashboardPage.js [hunk 4]
```diff
@@ -0,0 +1,31 @@
...
+import ClusterSetupWizard from './ClusterSetupWizard'
+
+const styles = theme => ({
+  welcome: {
+    fontSize: '24px',
+    marginBottom: '20px'
+  },
+})
+
+class Dashboard extends React.Component {
```
xantorres: Not a priority, but we are trying to embrace functional components and Hooks instead of class components and HoC's, so it would be nice to see new implementations created that way.

src/app/core/components/expansionPanel/ExpansionPanel.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+})
+
+const ExpansionPanel = ({ classes, expanded, onClick, children, completed, stepNumber, summary }) => {
+  return (
+    <MDExpansionPanel className={classes.root} expanded={expanded} onClick={onClick}>
+      <ExpansionPanelSummary>
+        {expanded ? <ExpandMore /> : <ChevronRight />}
+        <div className={classes.summaryText}>
+          { completed ? <div className={`${classes.numberedStep} ${classes.completedStep}`}><i className='fal fa-check' /></div> : <div className={classes.numberedStep}>{stepNumber}</div> }
```
xantorres: This line is too long and hard to read, please split it into lines or/and set it in a variable and then assign it here

src/app/core/components/icons/TopRightIcons.js
```diff
@@ -0,0 +1,70 @@
...
+    fontSize: '14px',
+    marginRight: '6px'
+  },
+  link: {
+    textDecoration: 'none',
+    color: '#606060'
+  }
+})
+
+class TopRightIcons extends React.PureComponent {
```
xantorres: We already have a component called `TopExtraContent` that will probably fulfill your needs, you should probably use/improve that to avoid code repetition.

src/app/plugins/kubernetes/components/dashboard/ClusterSetupWizard.js
```diff
@@ -0,0 +1,111 @@
...
+    icon: 'fa-file-alt',
+    label: 'Setup Documentation',
+    link: '/ui/kubernetes/dashboard'
+  }
+]
+
+class ClusterSetupWizard extends React.Component {
+  // Initial state will be handled by a loader later on
+  state = {
+    panelCtrl: {
```
xantorres: A simple array with index numbers would be simpler and in case of refactoring (adding or removing panels) we wouldn't need to remember to update this state object.

src/app/core/components/icons/TopRightIcons.js
kevinylee1991: I tried thinking of a way to take children, but i'm not sure how to keep the styling (since each icon representation is split into nested html). Any suggestions?

src/app/core/components/icons/TopRightIcons.js
kevinylee1991: I tried taking a look at TopExtraContent -- it seems to be top content for the whole page. Currently my component is really just right-aligning items with flexbox, maybe that would be a more useful component?

src/app/plugins/kubernetes/components/dashboard/ClusterSetupWizard.js
kevinylee1991: I couldn't think of a good way to implement this with without losing the ability to keep several of the panels open. Could you give me an overview of the solution you were thinking of?

src/app/plugins/kubernetes/components/dashboard/ClusterSetupWizard.js
```diff
@@ -68,9 +69,35 @@ class ClusterSetupWizard extends React.Component {
...
     return (
       <div className={classes.root}>
-        <TopRightIcons icons={icons} />
+        <RightAlign>
+          <div className={classes.button}>
+            <Link className={classes.link} to='/ui/kubernetes/dashboard'>
+              <FontAwesomeIcon className={classes.icon}>user-plus</FontAwesomeIcon>
+              <span>Invite Team Member</span>
+            </Link>
+          </div>
```
brennancheung: Make this into a small component.  You can define it as a nested functional component inside of the parent so that it has access to the same closure for the `classes`.

src/app/plugins/kubernetes/components/dashboard/ClusterSetupWizard.js
xantorres: Example using hooks:
```
// Initialize the state with the first panel activated (index 0)
const [activePanels, setActivePanels] = useState([0])

const togglePanel = useCallback(panelIdx => {
	setActivePanels(
		activePanels.includes(panelIdx)
			? except(panelIdx, activePanels) // Remove the panel idx if expanded
			: [panelIdx, ...activePanels]    // Add the panel idx if not expanded
	)
}, [activePanels])
```

src/app/plugins/kubernetes/components/dashboard/ClusterSetupWizard.js
xantorres: In case of defining the functional component inside the parent as @brennancheung suggested, I would also suggest using `useMemo` to prevent creating new instances of that component on every render.

src/app/plugins/kubernetes/components/dashboard/ClusterSetupWizard.js
```diff
@@ -0,0 +1,125 @@
...
+        expanded={activePanels.includes(1)}
+        onClick={() => togglePanel(1)}
+        stepNumber="1"
+        summary="Choose a destination for your cluster"
+      >
+        <div>Stuff goes here blah blah blah</div>
+      </ExpansionPanel>
+      <ExpansionPanel
+        expanded={activePanels.includes(2)}
+        onClick={() => togglePanel(2)}
```
brennancheung: Make `togglePanel` curry the panel number and return a partial function with it bound.

Ex:

`const togglePanel = panelIdx => () => { ... }`
`onClick={togglePanel(2)}`

This cleans up the markup.

src/app/plugins/kubernetes/components/dashboard/ClusterSetupWizard.js
```diff
@@ -0,0 +1,125 @@
...
+      </RightAlign>
+      <Typography className={classes.title}>Set up your first Kubernetes cluster</Typography>
+      <div className={classes.description}>
+        <Typography className={classes.descriptionHeader}>Create your first Kubernetes Cluster</Typography>
+        <Typography>Create your first Kubernetes cluster so you can start running some appications on it.</Typography>
+      </div>
+      <ExpansionPanel
+        expanded={activePanels.includes(1)}
+        onClick={() => togglePanel(1)}
+        stepNumber="1"
```
brennancheung: Is `stepNumber` a number or a string?  Looks like it is supposed to be the index?  I don't think it matters much but it should be consistent.  Either change to `stepNumber={1}` or `togglePanel('1')` so it is the same.  Also, add a prop documenting the type so that it enforces the consistency.

src/app/core/components/expansionPanel/ExpansionPanel.js
```diff
@@ -0,0 +1,78 @@
...
+
+const ExpansionPanel = ({ expanded, onClick, children, completed, stepNumber, summary }) => {
+  const classes = useStyles()
+
+  return (
+    <MDExpansionPanel className={classes.root} expanded={expanded} onClick={onClick}>
+      <ExpansionPanelSummary>
+        {expanded ? <ExpandMore /> : <ChevronRight />}
+        <div className={classes.summaryText}>
+          { completed ? <div className={`${classes.numberedStep} ${classes.completedStep}`}>
```
brennancheung: `?` and `:` should be aligned on the same column unless it's short enough to fit on a single line.  It's hard to see the JSX hierarchy with this code formatting style.

src/app/plugins/kubernetes/components/dashboard/ClusterSetupWizard.js
```diff
@@ -0,0 +1,125 @@
...
+  return (
+    <div className={classes.root}>
+      <RightAlign>
+        <ButtonLink
+          url='/ui/kubernetes/dashboard'
+          icon='user-plus'
+          label='Invite Team Member'
+        />
+        <ButtonLink
+          url='/ui/kubernetes/dashboard'
```
xantorres: Duplicated url here

src/app/plugins/kubernetes/components/dashboard/ClusterSetupWizard.js
xantorres: Actually, it doesn't necessary require to be a number, it can be a string as long as there is consistency as you pointed out.

PR #389, opened 2019-10-10
Refactor code to work with real Logging data

murilovmachado:
This PR implements:
- Refactor routes to reflect real endpoints
- Fetches logging data for each cluster
- Refactor `LoggingStub` to provide data similar to the real one


src/app/plugins/kubernetes/components/logging/LoggingStub.js
```diff
@@ -80,51 +43,124 @@ const getLoggings = () => [
...
-      'http://mybucket.s3.johnsmith',
-    ],
+  apiVersion: 'logging.pf9.io/v1alpha1',
+  metadata: {
+    name: 'es-object',
+    generation: 1,
+    resourceVersion: '3054',
+    creationTimestamp: '2019-10-08T20:07:28Z',
+    selfLink: '/apis/logging.pf9.io/v1alpha1/outputs/es-object',
+    uid: '2836f085-39cf-4b15-ae56-a653a46c8d44'
```
brennancheung: You can use `uuid.v4()` for these.  Also, we use the `faker` library to stub out a lot of data.  Be very careful about copy and pasting data from API's into stubs that get checked into github because they may contain sensitive information.  That's not the case here but it's generally better to fake the data rather than copy and paste actual responses.  The code is more generic and we can simulate different amounts of data when you use a mock that auto-generates random data.

src/app/plugins/kubernetes/components/logging/LoggingStub.js
```diff
@@ -80,51 +43,124 @@ const getLoggings = () => [
...
+
+const createLoggingsJSON = (loggingsOfOneCluster) => {
+  const items = []
+
+  loggingsOfOneCluster.logStorage.forEach((storage, index) => {
+    const destination = loggingsOfOneCluster.logDestination[index]
+
+    if (storage === storageType.S3) {
+      items.push(s3Item(destination))
+    } else if (storage === storageType.ELASTIC_SEARCH) {
```
brennancheung: No need for the `else` here.

If you have longer lists of variants you can also do:

```javascript
const mapStorage = {
  [storageType.S3]: s3Item,
  [storageType.ELASTIC_SEARCH]: elasticSearchItem,
}
items.push(mapItems[storage](destination)
```
```

PR #390, opened 2019-10-11
Added segment page tracking

kevinylee1991:
Page change tracking for Segment. Seemed to make sense to place into that component.

May be necessary to add a feature flag for this, like if we have e2e testing for example we don't want to be sending those events to Segment.

### DIFF
--- a/src/app/core/components/AppContainer.js
+++ b/src/app/core/components/AppContainer.js
@@ -3,6 +3,7 @@ import axios from 'axios'
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import { withStyles } from '@material-ui/styles'
+import { withRouter } from 'react-router-dom'
 import clsx from 'clsx'
 import Intercom from 'core/components/integrations/Intercom'
 import Navbar, { drawerWidth } from 'core/components/Navbar'
@@ -50,6 +51,7 @@ const styles = theme => ({
   },
 })
 
+@withRouter
 @withStyles(styles, { withTheme: true })
 class AppContainer extends PureComponent {
   async componentDidMount () {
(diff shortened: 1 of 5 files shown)


src/app/core/components/AppContainer.js
```diff
@@ -72,6 +74,23 @@ class AppContainer extends PureComponent {
     }
   }
 
+  componentWillMount() {
+    const { history } = this.props
+
+    this.unlisten = this.props.history.listen((location, action) => {
+      if (!analytics) { return }
```
brennancheung: * You should only do 1 check for `if (!analytics)` at the top.
* Move analytics functionality into it's own method and call it.
* I'm assuming `analytics` is exposed in `window.analytics` from some global script added to the page?  Let's not call Segment directly.  Go through an intermediary function that we can customize later on if we need to.
* Come up with a pure JSON format that is passed to it.  That will be parsed and then sent to segment.  The redux standard actions are probably a good starting point.  (`{ type: 'page_load', payload: { route: '...' } }`).

src/app/static/index.html
```diff
@@ -9,5 +9,10 @@
   </head>
   <body>
     <div id="root"></div>
+    <script type="text/javascript">
+      !function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t,e){var n=document.createElement("script");n.type="text/javascript";n.async=!0;n.src="https://cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var a=document.getElementsByTagName("script")[0];a.parentNode.insertBefore(n,a);analytics._loadOptions=e};analytics.SNIPPET_VERSION="4.1.0";
```
brennancheung: Can we rename `analytics` to something like `segment_tracker` or will this break their functionality?  I don't like the fact that they take up such a common and universal token.

PR #392, opened 2019-10-15
Azure cluster creation

brennancheung:
The Azure Cluster Create is very similar to AWS.  I'm just copying it and will make the tweaks in future commits.


src/app/plugins/kubernetes/components/infrastructure/AddAzureClusterPage.js
```diff
@@ -0,0 +1,465 @@
...
+  { label: 'Create new network', value: 'newNetwork' },
+]
+
+const AddAzureClusterPage = () => {
+  const { params, getParamsUpdater } = useParams()
+  const { history } = useReactRouter()
+  const onComplete = () => {
+    history.push('/ui/kubernetes/infrastructure#clusters')
+  }
+  const [create] = useDataUpdater(clusterActions.create, onComplete)
```
xantorres: blocking the form when the update is being performed would be a good idea, you can use the second variable (`loading`) returned by `useDataUpdater` along with `Progress` component to do so

PR #394, opened 2019-10-16
Fix fetch loggings for invalid clusters

murilovmachado:
This PR fixes the issue of trying to fetch loggings from an invalid cluster. It also adjusts the stubbed `getLoggings(clusterId)` to return an object instead of a list.

### DIFF
--- a/src/app/plugins/kubernetes/components/logging/actions.js
+++ b/src/app/plugins/kubernetes/components/logging/actions.js
@@ -1,6 +1,7 @@
 import ApiClient from 'api-client/ApiClient'
 import createCRUDActions from 'core/helpers/createCRUDActions'
 import { clustersCacheKey } from 'k8s/components/infrastructure/actions'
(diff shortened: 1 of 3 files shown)


src/app/plugins/kubernetes/components/logging/actions.js
```diff
@@ -33,13 +33,18 @@ const mapLoggings = (cluster, loggings) => {
...
-
-    const loggingsPromises = clusters.map(async cluster => {
-      const response = await qbert.getLoggings(cluster.uuid)
-      return mapLoggings(cluster, response[0])
-    })
-
-    const loggings = await Promise.all(loggingsPromises)
+    const loggings = []
+
+    for (const cluster of clusters) {
```
brennancheung: Use `utils/async/mapAsync`.  Push the error handling into the API client as well so it keeps the application code cleaner.

PR #395, opened 2019-10-17
Create Bare OS Cluster

brennancheung:
This PR adds the ability to create clusters from nodes with just a bare OS (manual).

The backend team is still working on the CLI tool.  I have added placeholder for now until we have that available.


src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
```diff
@@ -0,0 +1,272 @@
...
+  { label: 'Default API groups and versions', value: 'default' },
+  { label: 'All API groups and versions', value: 'all' },
+  { label: 'Custom', value: 'custom' },
+]
+
+const AddBareOsClusterPage = () => {
+  const { params } = useParams()
+  const { history } = useReactRouter()
+  const onComplete = () => {
+    history.push('/ui/kubernetes/infrastructure#clusters')
```
xantorres: `useDataUpdater` will pass a `success` parameter to this function, you can use that to decide whether to redirect the user or not depending on the result of the cluster creation

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
```diff
@@ -0,0 +1,272 @@
...
+    await create(body)
+
+    // 3. Attach the nodes
+    const nodes = [
+      ...data.masterNodes.map(uuid => ({ isMaster: true, uuid })),
+      ...data.workerNodes.map(uuid => ({ isMaster: false, uuid })),
+    ]
+    qbert.attachNodes(body.clusterUuid, nodes)
+
+    return body
```
xantorres: shouldn't all this be in `clusterActions.createFn` rather than here? It seems to me that the view is doing far too much.

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
```diff
@@ -0,0 +1,272 @@
...
+
+            <WizardStep stepId="network" label="Network Info">
+              <FormWrapper title="Network Info">
+                <ValidatedForm initialValues={wizardContext} onSubmit={setWizardContext} triggerSubmit={onNext}>
+                  {({ setFieldValue, values }) => (
+                    <>
+                      <TextField
+                        id="masterVipIpv4"
+                        label="Virtual IP address for cluster"
+                        info=""
```
xantorres: Empty info?

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/ClusterHostChooser.js
```diff
@@ -0,0 +1,80 @@
...
+  const [selected, setSelected] = useState(value)
+
+  // TODO: need to figure out a way to do validtion with our system to select 1, 3, or 5 nodes only for masters
+  // const isValid = () => !isMaster || [1, 3, 5].includes(selected.length)
+
+  const allSelected = () => selected.length === hosts.length
+  const toggleAll = () => setSelected(allSelected() ? [] : hosts.map(x => x.uuid))
+  const isSelected = uuid => selected.includes(uuid)
+
+  const [nodes] = useDataLoader(loadNodes)
```
xantorres: `const [nodes, loading] = useDataLoader(loadNodes)`
and then wrap everything in a `<Progress loading={loading}>`

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/ClusterHostChooser.js
```diff
@@ -0,0 +1,80 @@
...
+  const notInExcludeList = node => isMaster || !excludeList.includes(node.uuid)
+
+  const hosts = nodes.filter(notAssignedToCluster).filter(notInExcludeList)
+
+  const toggleHost = uuid => () => {
+    const newHosts = isSelected(uuid) ? selected.filter(x => x !== uuid) : [ ...selected, uuid ]
+    setSelected(newHosts)
+  }
+
+  useEffect(() => {
```
xantorres: why not create a function (using `useCallback` if you will) that both call `setSelected` and `onChange` instead of waiting for changes on the state with a `useEffect` here?

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
brennancheung: Possibly.  I could go either way.  It's actually doing several different actions though.  Looking info up, creating a cluster, and attaching nodes.  So it's more like 3 actions that are completely different.

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/ClusterHostChooser.js
brennancheung: Actually, I think I'm going to try and see if I can completely get rid of state in the component and just make it a fully controlled component.

PR #397, opened 2019-10-18
StorageClasses empty Add Page

murilovmachado:
This PR creates an empty StorageClasses AddPage and link it to the route `/storage_classes/add`


xantorres:
> This LGTM. @xantorres Are we still doing `createAddComponents` as the best way to create new CRUD pages? I ask because a lot of stuff that doesn't quite fit the mold requires dropping down and creating things more manually.

We are still doing so, yes, usually `createCRUDComponents` for creating pages with lists, and `createAddComponents` for pages with add forms.
But I would like to refactor this as soon as possible to get rid of these helpers and instead create wrapper React Components that will be easier to understand and customize.

PR #399, opened 2019-10-21
Creates Add Storage Class Wizard

murilovmachado:
This PR creates Add Storage Class Wizard with 2 wizard steps: Basic and Custom. It also implements the forms in both steps. `handleSubmit` will be implemented in a future PR.


src/app/plugins/kubernetes/components/storage/StorageClassesAddPage.js
```diff
@@ -1,9 +1,91 @@
...
+            info="Name for this storage class."
+            required
+          />
+          <PicklistField
+            DropdownComponent={ClusterPicklist}
+            id="clusterId"
+            label="Cluster"
+            onChange={getParamsUpdater('clusterId')}
+            info="The cluster to deploy this storage class on."
+            value={params.clusterId}
```
brennancheung: The information from this field will be passed through along with all the other form data in the `onSubmit`.  If you don't have any intermediary need (other fields that need to use this value) then you can safely omit the `value` and `onChange` props.

src/app/plugins/kubernetes/components/storage/StorageClassesAddPage.js
```diff
@@ -1,9 +1,91 @@
...
+            required
+          />
+          <CheckboxField
+            id="isDefault"
+            label="Use as Default Storage Class"
+            onChange={getParamsUpdater('isDefault')}
```
brennancheung: Doesn't look like you are using `isDefault` on any of the other fields so you can omit the `onChange`.

src/app/plugins/kubernetes/components/storage/StorageClassesAddPage.js
```diff
@@ -1,9 +1,82 @@
...
+    <p>
+      Optionally edit the storage class YAML for advanced configuration. See this <a href='https://kubernetes.io/docs/concepts/storage/persistent-volumes/#storageclasses'>article</a> for more information.
+    </p>
+    <p>
+      <b>NOTE:</b> In case of a conflict with options selected on the previous page, changes you make here will override them.
+    </p>
+    <br />
+    <FormWrapper title="Customize">
+      <ValidatedForm onSubmit={onSubmit} triggerSubmit={triggerSubmit}>
+        <CodeMirror
```
brennancheung: Verify `CodeMirror` works with `ValidatedForm`.  It might be a standalone controlled component.

murilovmachado: Looking at other pieces of our codebase `CodeMirror` is being used with `ValidatedForm`.

PR #401, opened 2019-10-22
Pods improvements and fixes

xantorres:
- Created common `AddResourcePage` component to deal with the creation of pods, deployments and services
- Add documentation links
- Resource type picklist
- Allow the insertion of YAML example templates
- Add help info in `AddResourcePage` form fields
- Fix Tooltip in `CodeMirror` field
- Allow selection of all clusters in pods, deployments and services list pages
- Add additional columns in pods, deployments and services list pages
- Allow filtering by Namespace
- Disable delete deployments function


src/app/plugins/kubernetes/components/pods/renderLabels.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { Typography } from '@material-ui/core'
+import { toPairs } from 'ramda'
+
+const stylesByType = {
+  label: { color: '#00C036', fontWeight: 400 },
```
brennancheung: We should ideally be pulling all styling values from the theme.

src/app/utils/misc.js
```diff
@@ -14,6 +14,36 @@ export const parseJSON = str => {
...
+/**
+ * Given a number of seconds returns the number of
+ * years, months, days, hours and minutes in a human readable format
+ * @param seconds
+ * @returns {string}
+ */
+export const secondsToString = seconds => {
+  let remainingSeconds = seconds
+  const units = {
+    'year': 24 * 60 * 365,
```
brennancheung: No need for the single quotes.

Also, I don't think `year` is accurate.

Might be cleaner as:

```
const hour = 60 * 60
const day = 24 * hour
const month = 30 * day
...
const units = { hour, day, month }
```

Also, check it the `moment` library which we already have installed has any of this functionality.

src/app/utils/misc.js
```diff
@@ -14,6 +14,36 @@ export const parseJSON = str => {
...
+    'month': 24 * 60 * 60 * 30,
+    'day': 24 * 60 * 60,
+    'hour': 60 * 60,
+    'min': 60,
+    // 'second': 1,
+  }
+  const results = []
+  for (const unitName in units) {
```
brennancheung: This seems like it can be written dramatically simpler with `map`.  Also, the `amount === 1` and `>= 2` has duplicate overlapping code paths.  You can add the first part unconditionally and then check to see if you need to add the `s`.

src/app/utils/misc.js
xantorres: I already checked, it has something called `fromNow()` but unfortunately only displays the larger unit, so I had to create this

src/app/plugins/kubernetes/components/pods/renderLabels.js
xantorres: This is a very specific thing that will only be applied here, not sure if it's worth it to have it in the theme.

src/app/utils/misc.js
xantorres: I would still need to check if the `amount` is >= than 1 so not much extra gain here.

PR #403, opened 2019-10-23
Implement StorageClasses createFn

murilovmachado:
This PR implements `createFn` and fix some pieces in order for the creation of StorageClasses to work.


src/api-client/Qbert.js
```diff
@@ -214,8 +214,8 @@ class Qbert {
     return data.items.map(this.convertResource(clusterId)).map(mapStorageClass)
   }
 
-  createStorageClass = async (clusterId, params) => {
-    return this.client.basicPost(`${await this.baseUrl()}/clusters/${clusterId}/k8sapi/apis/storage.k8s.io/v1/storageclasses`)
+  createStorageClass = async (clusterId, body) => {
+    return this.client.basicPost(`${await this.baseUrl()}/clusters/${clusterId}/k8sapi/apis/storage.k8s.io/v1/storageclasses`, body)
```
xantorres: you can use `clusterBaseUrl()` rather than `baseUrl()`

src/app/plugins/kubernetes/components/storage/StorageClassesAddPage.js
```diff
@@ -57,27 +62,67 @@ const BasicStep = ({ onSubmit, triggerSubmit }) =>
     </FormWrapper>
   </WizardStep>
 
-const CustomizeStep = ({ onSubmit, triggerSubmit }) =>
+const CustomizeStep = ({ ...rest }) =>
```
xantorres: `const CustomizeStep = props =>`

src/api-client/Qbert.js
murilovmachado: actually the URL is slightly different, so I can't use `clusterBaseUrl`. Note the `/storage.k8s.io/v1` part.

PR #407, opened 2019-10-23
Return strings instead of booleans

murilovmachado:
The backend expects strings for `true` and `false` instead of booleans in the Storage Class YAML. This PR fixes the front-end to meet this requirement.


brennancheung:
This seems like a bug in the backend.  We can change this for now to get it working but I'm going to follow up to see if there is anything we can do to make this more consistent with normal JSON API conventions using normal booleans.

PR #408, opened 2019-10-23
[WIP] Reload storageClass list after adding

murilovmachado:
This PR fixes the code so that after adding a `StorageClass`, it loads the `ListPage`.

TODO: Right now, when it loads `ListPage` it doesn't show the newly added `StorageClass` as in the image below. There's an item in the table, but without its info.

Screen Shot 2019-10-23 at 19 44 20

Any idea why this is happening? @xantorres @brennancheung 
After refreshing, it shows everything just fine


xantorres:
The mapping of the data should be done *always* using `dataMapper` option from the `contextLoader` or `createCRUDActions` (whichever you are using) this way the data added to the cache using `createFn` will be remapped and you will have the custom fields available in the view.

xantorres:
Pushed a commit with the necessary changes to fix the list and the refreshing after adding a storage class.

PR #409, opened 2019-10-24
Small tweaks to Bare OS Cluster creation

brennancheung:
* Make Virtual IP conditionally required when number of master nodes > 1
* Allow user to specify multiple Metal LB IP pool ranges
* Change network interface picklist to only use interface name (not IP)

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
@@ -8,6 +8,7 @@ import KeyValuesField from 'core/components/validatedForm/KeyValuesField'
 import PicklistField from 'core/components/validatedForm/PicklistField'
 import TextField from 'core/components/validatedForm/TextField'
 import ValidatedForm from 'core/components/validatedForm/ValidatedForm'
+import VipInterfaceChooser from './VipInterfaceChooser'
 import Wizard from 'core/components/wizard/Wizard'
 import WizardStep from 'core/components/wizard/WizardStep'
 import useDataUpdater from 'core/hooks/useDataUpdater'
@@ -31,7 +32,7 @@ const runtimeConfigOptions = [
 ]
 
 const AddBareOsClusterPage = () => {
-  const { params } = useParams()
+  const { params, getParamsUpdater } = useParams()
   const { history } = useReactRouter()
   const onComplete = success => {
     if (!success) { return console.error('Unable to create cluster') }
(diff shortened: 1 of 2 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/VipInterfaceChooser.js
```diff
@@ -0,0 +1,40 @@
...
+import { pathStrOr } from 'utils/fp'
+import { uniq } from 'ramda'
+
+const VipInterfaceChooser = forwardRef(({
+  masterNodes, hasError, errorMessage, ...rest
+}, ref) => {
+  const [nodes, loading] = useDataLoader(loadNodes)
+  const masters = nodes.filter(node => masterNodes.includes(node.uuid))
+
+  const options = uniq(masters.map(node => {
```
xantorres: This looks a bit expensive, `useMemo` perhaps?

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
```diff
@@ -31,7 +32,7 @@ const runtimeConfigOptions = [
 ]
 
 const AddBareOsClusterPage = () => {
-  const { params } = useParams()
+  const { params, getParamsUpdater } = useParams()
   const { history } = useReactRouter()
   const onComplete = success => {
     if (!success) { return console.error('Unable to create cluster') }
```
xantorres: You don't need to do this, errors are already shown at console and a Toast notification will pop up too

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/VipInterfaceChooser.js
brennancheung: It's not expensive at all.  Typical scenario is between 2-5 nodes.  Not worth the added code complexity.

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
brennancheung: I'm confused, your last comment on this PR suggested I use the `success` parameter.

xantorres: It makes sense in some cases, eg to decide if you redirect the user to the list screen, or stay in there has been an error, so the user can fix any mistakes and try again

PR #410, opened 2019-10-24
Fix prometheus rules page

smanpathak:
This change fixes the prometheus rule edit page by fetching right columns, checking for nulls, etc.
Screen Shot 2019-10-23 at 6 29 14 PM


src/app/plugins/kubernetes/components/prometheus/PrometheusRulesTable.js
```diff
@@ -1,30 +1,46 @@
...
+
+const renderKeyValues = obj => Object.entries(obj)
+  .map(([key, value]) => `${key}: ${value}`)
+  .join(', ')
 
 const PrometheusRulesTable = ({ rules, onDelete }) => {
+  var allRules = []
+  rules.forEach(function (group) {
+    group.rules.forEach(function (rule) {
+      rule['group'] = group.name
```
xantorres: you are mutating rule on each iteration, and also you are overwriting the `group` prop on each iteration of each `rule`, in a way that only the last group will be assigned to the rule.

smanpathak: Hi @xantorres 
I have rewritten this with a map and flatten. Could not figure out the reduce.

Thanks

xantorres: Well I just copied you the code, but it's ok

src/app/plugins/kubernetes/components/prometheus/PrometheusRulesTable.js
```diff
@@ -1,30 +1,44 @@
...
+import { isNilOrEmpty } from 'utils/fp'
+
+const renderKeyValues = obj => Object.entries(obj)
+  .map(([key, value]) => `${key}: ${value}`)
+  .join(', ')
 
 const PrometheusRulesTable = ({ rules, onDelete }) => {
+  var groups = rules
+  var allRules = groups.map((group) => {
+    return [...group.rules.map(rule => ({
```
xantorres: You don't need to do the `[... ` is useless in this case

brennancheung: Use `const` instead of `var`.

Also when using short anonymous functions that return objects there is a shorthand:

`arr.map(x => ({ foo: x.abc, bar: x.xyz }))`

Basically just wrapping the object with parentheses.

smanpathak: removed.

PR #415, opened 2019-10-25
Storage classes isDefault behaviour

murilovmachado:
If there is already a default storage class for the chosen cluster, `isDefault` checkbox will be disabled.

Screen Shot 2019-10-25 at 16 19 23

### DIFF
--- a/src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
+++ b/src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useCallback } from 'react'
 import yaml from 'js-yaml'
 import createAddComponents from 'core/helpers/createAddComponents'
 import Wizard from 'core/components/wizard/Wizard'
@@ -9,10 +9,11 @@ import TextField from 'core/components/validatedForm/TextField'
 import PicklistField from 'core/components/validatedForm/PicklistField'
 import CheckboxField from 'core/components/validatedForm/CheckboxField'
 import CodeMirror from 'core/components/validatedForm/CodeMirror'
-import { codeMirrorOptions } from 'app/constants'
+import { codeMirrorOptions, allKey } from 'app/constants'
 import ClusterPicklist from 'k8s/components/common/ClusterPicklist'
-import { storageClassesCacheKey } from './actions'
+import storageClassesActions, { storageClassesCacheKey } from './actions'
 import useParams from 'core/hooks/useParams'
+import useDataLoader from 'core/hooks/useDataLoader'
 
 const initialContext = {
   isDefault: false,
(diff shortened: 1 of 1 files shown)


src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
```diff
@@ -32,53 +33,62 @@ export const AddStorageClassForm = ({ onComplete }) =>
...
-    return null
+const BasicStep = ({ onSubmit, triggerSubmit }) => {
+  const listStorageClassesParams = {
+    clusterId: allKey,
+    healthyClusters: true,
   }
+  const [storageClasses] = useDataLoader(storageClassesActions.list, listStorageClassesParams)
+  const { params, updateParams } = useParams()
+  const updateClusterId = useCallback(clusterId => updateParams({ clusterId }), [params.clusterId])
+  const defaultExists = !!storageClasses.find(storageClass =>
```
brennancheung: Move the conditional into its own function.

brennancheung: The name `defaultExists` sounds like the name for a predicate function, not a list of storageClasses.

src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
```diff
@@ -32,53 +33,62 @@ export const AddStorageClassForm = ({ onComplete }) =>
...
-  if (isInvalidContext) {
-    return null
+const BasicStep = ({ onSubmit, triggerSubmit }) => {
+  const listStorageClassesParams = {
+    clusterId: allKey,
+    healthyClusters: true,
   }
+  const [storageClasses] = useDataLoader(storageClassesActions.list, listStorageClassesParams)
+  const { params, updateParams } = useParams()
+  const updateClusterId = useCallback(clusterId => updateParams({ clusterId }), [params.clusterId])
```
brennancheung: `getParamsUpdater` can be used to create these functions for you.

src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
```diff
@@ -32,53 +33,62 @@ export const AddStorageClassForm = ({ onComplete }) =>
...
+            label="Name"
+            info="Name for this storage class."
+            required
+          />
+          <PicklistField
+            DropdownComponent={ClusterPicklist}
+            id="clusterId"
+            label="Cluster"
+            info="The cluster to deploy this storage class on."
+            onChange={updateClusterId}
```
brennancheung: `onChange={getParamsUpdater('clusterId')}`

src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
murilovmachado: `defaultExists` is actually a `boolean`.

src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
```diff
@@ -32,53 +33,61 @@ export const AddStorageClassForm = ({ onComplete }) =>
...
+            info="Name for this storage class."
+            required
+          />
+          <PicklistField
+            DropdownComponent={ClusterPicklist}
+            id="clusterId"
+            label="Cluster"
+            info="The cluster to deploy this storage class on."
+            onChange={getParamsUpdater('clusterId')}
+            value={params.clusterId}
```
xantorres: Pass a `showAll={false}` prop here, as I think it has to be mandatory to assign a cluster when creating a Storage class, so the "All" option makes no sense here

src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
```diff
@@ -32,53 +33,61 @@ export const AddStorageClassForm = ({ onComplete }) =>
...
-    <CustomizeStepContent {...props} />
-  </WizardStep>
-
-const CustomizeStepContent = ({ isActive, wizardContext, onSubmit, triggerSubmit }) => {
-  const isInvalidContext = wizardContext.name == null || wizardContext.isDefault == null
-  if (isInvalidContext) {
-    return null
+const BasicStep = ({ onSubmit, triggerSubmit }) => {
+  const listStorageClassesParams = {
+    clusterId: allKey,
```
xantorres: Probably you don't want to show the "All" option so it will be better to leave this to null

src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
murilovmachado: The `All` option is actually not being shown.

PR #416, opened 2019-10-25
Add grafana dashboard link

smanpathak:
This change adds link to access grafana dashboard service running on PML clusters which have monitoring feature enabled. It appears like below --

Screen Shot 2019-10-25 at 4 09 53 PM


src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -33,7 +31,6 @@ const getKubernetesVersion = async clusterId => {
 }
 
 export const hasMasterNode = propSatisfies(isTruthy, 'hasMasterNode')
-export const hasHealthyMasterNodes = propSatisfies(healthyMasterNodes => healthyMasterNodes.length > 0, 'healthyMasterNodes')
```
xantorres: This filter was used in other pages, such as the Storage class pages, why is it being removed here?

smanpathak: Good catch, it was removed during code reshuffling I guess. Posting updated

PR #417, opened 2019-10-25
Implement storageType picklist

murilovmachado:
Implements Storage Type Picklist.

Screen Shot 2019-10-25 at 20 15 32
Screen Shot 2019-10-25 at 20 15 40


src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
```diff
@@ -69,6 +70,15 @@ const BasicStep = ({ onSubmit, triggerSubmit }) => {
             onlyHealthyClusters
             required
           />
+          <PicklistField
+            DropdownComponent={StorageTypePicklist}
+            id="storageType"
+            label="Storage Type"
+            info="Select the storage type for this storage class. The list of available storage types is specific to the cloud provider that the cluster belongs to."
+            onChange={getParamsUpdater('storageType')}
+            value={params.storageType}
```
brennancheung: You can probably omit the value.  Any of the `...Field` components have default implementations of `onChange` and `value` that use the `ValidatedForm`'s context to store and retrieve its state.  The supplied `onChange` event that you pass is called *in addition to* setting the state in `ValidatedForm`'s context.

It's only when something deviants from the normal default that these fields need to set.  In your case it looks like you only need to customize the setting of the param.

Also, if the dependent field is within the same `ValidatedForm` you can use the renderProps version of `ValidatedForm`.  It provides access to a `values` object that you can use.  You can see a good example of fields that depend on other field values here: https://github.com/platform9/pf9-ui-plugin/blob/master/src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js#L175

Note that in the example, you do not need to use `onChange` or `value` in the props.  If the dependent field is inside a different `ValidatedForm` (as in the case of multiple `WizardStep`s) then you will have to fall back to using `params.xyz` and `getParamsUpdater('xyz')`.

src/app/plugins/kubernetes/components/storage/StorageTypePicklist.js
```diff
@@ -0,0 +1,40 @@
+import React, { forwardRef } from 'react'
+import PropTypes from 'prop-types'
+import Picklist from 'core/components/Picklist'
+
+// We need to use `forwardRef` as a workaround of an issue with material-ui Tooltip https://github.com/gregnb/mui-datatables/issues/595
+const StorageTypePicklist = forwardRef(({
+  loading,
```
brennancheung: The `loading` value is supposed to be used in conjunction with the `Progress` component.  `grep` for `Progress` for examples in the codebase.

src/app/plugins/kubernetes/components/storage/StorageTypePicklist.js
```diff
@@ -0,0 +1,40 @@
...
+  onChange,
```
brennancheung: Looks like you are just passing `onChange` through and so you can probably just delete this line and the prop in the Picklist and let it pass through in `...rest`.

src/app/plugins/kubernetes/components/storage/StorageTypePicklist.js
brennancheung: Or sometimes `loading` can be used to disable a component as well instead of using a `progress`.

Usually something like `disabled={loading}` if the component supports it and it makes sense for the given scenario.

murilovmachado: Yep, I've forgotten to pass `loading` to `Picklist` component. In this case, I'll pass it through `props` variable, since I'm not modifying it (not doing something like `loading={loading || clustersLoading}`)

PR #419, opened 2019-10-26
decco container for next gen UI

bcle:
- app name is pf9-serenity
- complies with decco/axon standard by supply an init-region script (which does nothing presently) and an app_metadata image annotation
- uses confd to set /etc/nginx/conf.d/pf9/cors.conf dynamically from Consul
- in order to build and push image, first run "make rpm" from support/ directory, then cd to decco-container/ and run "make push"
- required change in pf9-qbert repo: must add "admin_url" endpoint to Keystone at init-region time, the new UI needs it, whereas the old one uses "internal_url"
- testing: teamcity build: http://teamcity.platform9.horse:8111/viewLog.html?buildId=1034415&buildTypeId=Pf9project_ContainerBuilds_SerenityContainer&tab=buildResultsDiv, live UI: see https://du-test-kplane-leb-2568.platform9.horse/ui/

### DIFF
--- a/decco-container/Makefile
+++ b/decco-container/Makefile
@@ -1,0 +1,76 @@
+# Copyright (C) 2019 Platform 9 Systems, Inc.
+
+.SUFFIXES:
+.PHONY: clean push image stage
+
+SRCROOT = $(abspath $(dir $(CURDIR)/../$(word $(words $(MAKEFILE_LIST)),$(MAKEFILE_LIST))))
+CONTAINER_ROOT=$(SRCROOT)/decco-container
+BUILD_DIR = $(SRCROOT)/build
+CONTAINER_BUILD_DIR = $(SRCROOT)/build-container
+STAGE = $(CONTAINER_BUILD_DIR)/stage
+CONTAINER_TAG = $(CONTAINER_BUILD_DIR)/container-tag
+EXTRACTED_RPM_DIR = $(CONTAINER_BUILD_DIR)/rpm
+PUBLIC_FILES =  $(STAGE)/publicfiles
+METADATA_FILE =  $(STAGE)/app_metadata.json
+UI_DIR =  $(PUBLIC_FILES)/ui
+PF9_VERSION ?= 1.0.0
+BUILD_NUMBER ?= 0
+BUILD_ID := $(BUILD_NUMBER)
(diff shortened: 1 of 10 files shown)


decco-container/conf/confd/templates/cors.conf
```diff
@@ -0,0 +1,24 @@
+# CORS config for nginx. Add this configuration to a location spec on a
+#    'slave' DU in nginx to provide an OPTIONS response that allows X-Auth-Tokens
+#    from a 'master' origin
+# NOTE - this works, but it's a little dicey - see
+#     https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/
+
+set $master_origin '{{getv "/fqdn"}}';
```
brennancheung: Does this also include `http://localhost:3000` for when we want to test a local development build and point it at an arbitrary backend?

decco-container/Makefile
```diff
@@ -0,0 +1,76 @@
...
+#	cp -a $(BUILD_DIR)/* $@/
+
+$(METADATA_FILE): $(UI_DIR)
+	echo "Downloading consul"
+	cd $(STAGE) && wget -q $(CONSUL_URL) && unzip consul_*.zip
+	cd $(STAGE) && wget -q -O confd $(CONFD_URL) && chmod +x confd
+	cp -f $(CONTAINER_ROOT)/Dockerfile $(STAGE)
+	cp -r $(CONTAINER_ROOT)/conf $(STAGE)
+	cp -f $(CONTAINER_ROOT)/{init-region,start.sh} $(STAGE)
+	python $(CONTAINER_ROOT)/y2j.py $(CONTAINER_ROOT)/app_metadata.yaml > $@
```
brennancheung: Looks like `app_metadata.yaml` is just a static file.  Why not just check in a JSON file instead of a YAML file and skip this step as well as the need for python and the `y2j` utility?

decco-container/conf/confd/templates/cors.conf
bcle: No it doesn't. But this is a container. I'd like to understand the desired workflow for a UI developer testing a local change. I'll connect with you offline.

decco-container/Makefile
bcle: Yeah, it's just a convention copied over from other repos. YAML is slightly simpler to read. I agree that JSON would probably be simpler from a build perspective.

PR #420, opened 2019-10-28
RBAC UI save progress

kevinylee1991:
Unfortunately having trouble getting the wizard to work well. With the time crunch, I think I'm going to need assistance getting this to work properly. I need to learn how all the helper components work and how / when to use them, but I don't think I'll be able to learn it fast enough for the upcoming kubernetes release deadline. Had some of these following issues:

1. In RbacChecklist component, I want the checklist to reload each time the clusterId is updated. I tried following the example in NamespacePicklist with useDataLoader, but I wasn't able to get it working with my loadApiGroups function in my rbac actions.js. I tried just loading it with no hook, but it caused an infinite loop. Ended up just putting it in useEffect for now so that I had something to work with.
2. Checkboxes have some lag when toggling. Not sure if this could be because checkboxes are expensive and keep re-rendering.
3. In the HTML there are nested maps -- I tried to separate them out in different functions, but somehow they did not load when I had them in separate functions. Maybe I was missing some specific syntax -- decided to revisit later.
4. After creating the role, it does not show up on the table correctly, it seems that whatever mechanism updates the cache after a create operation does not work with my form. Need to refresh the table so that the correct data is retrieved. I'm guessing I have some mismatch somewhere where I'm supplying something wrong to one of those helper components.
5. Trouble finding out how to get caching to work with the loadApiGroups function in rbac actions.js. Need it to work per cluster. Wasn't sure where to find a good example, so I figured I would try to figure it out later.


src/app/plugins/kubernetes/components/rbac/RbacChecklist.js
```diff
@@ -0,0 +1,140 @@
...
+  // Tried to separate out each map into its own function, but it
+  // does not work for some reason... had to put it all into
+  // this one function for it to work.
+  const renderApiGroup = (group) => {
+    return (
+      <div key={group.groupName} className={classes.group}>
+        <div className={classes.groupColumn}>{group.groupName}</div>
+        <div className={classes.resources}>
+          {group.resources.map(resource => {
+            return (
```
brennancheung: Use the shorthand and skip the extra `return` and `{`.

src/app/plugins/kubernetes/components/rbac/RbacChecklist.js
```diff
@@ -0,0 +1,140 @@
...
+              <div key={resource.name} className={classes.resource}>
+                <div className={classes.resourceName}>{resource.name}</div>
+                <FormGroup row>
+                  {resource.verbs.map(verb => {
```
brennancheung: Use shorthand

src/app/plugins/kubernetes/components/rbac/actions.js
```diff
@@ -0,0 +1,151 @@
...
+// import { map, pathEq, find, pluck, curry, pipe, last, pathOr, prop, propEq } from 'ramda'
+import { ascend, assoc, prop, propEq, sort } from 'ramda'
+import { flatMapAsync } from 'utils/async'
+import { parseClusterParams } from 'k8s/components/infrastructure/clusters/actions'
+import { clustersCacheKey } from '../infrastructure/common/actions'
+
+const { qbert } = ApiClient.getInstance()
+const uniqueIdentifier = 'metadata.uid'
+
+// Ask brennan / xan how to cache these API requests
```
brennancheung: Take a look at the actions under Apps for some examples on how to do this.  We basically have `contextUpdater` helpers for this.

src/api-client/Qbert.js
```diff
@@ -7,6 +7,30 @@ const normalizeClusterizedResponse = (clusterId, response) => propOr([], 'items'
   x => ({ ...x, clusterId }))
 const normalizeClusterizedUpdate = (clusterId, response) => ({ ...response, clusterId })
 
+const rbacApiGroupsToRules = (apiGroups) => {
+  const rules = []
+  Object.keys(apiGroups).forEach((apiGroupName) => {
```
brennancheung: These look like they should be nested maps.

src/api-client/Qbert.js
```diff
@@ -7,6 +7,30 @@ const normalizeClusterizedResponse = (clusterId, response) => propOr([], 'items'
...
+    Object.keys(apiGroups[apiGroupName]).forEach((resourceName) => {
+      const rule = {
+        apiGroups: [],
```
brennancheung: `apiGroups: apiGroupName === 'core' ? [''] : [apiGroupName]`

src/api-client/Qbert.js
```diff
@@ -7,6 +7,30 @@ const normalizeClusterizedResponse = (clusterId, response) => propOr([], 'items'
...
+        resources: [],
+        verbs: []
```
brennancheung: No need to set something to empty array and then unconditionally set it down below.  Just set it when you are initializing it.

`verbs: Object.keys(apiGroups[apiGroupName][resourceName])`

src/server/api/qbert/roles/roleActions.js
```diff
@@ -4,6 +4,7 @@ import Role from '../../../models/qbert/Role'
 export const getRoles = (req, res) => {
   const { namespace, clusterId } = req.params
   const roles = Role.list({ context, config: { clusterId, namespace } })
+  console.log(roles)
```
brennancheung: remove

src/app/plugins/kubernetes/components/rbac/RbacChecklist.js
```diff
@@ -0,0 +1,113 @@
...
+  // Todo: how to reset checkedItems upon clusterId update?
+  const handleCheck = useCallback((group, resource, verb) => {
+    const path = [group, resource, verb]
+    if (hasPath(path, checkedItems)) {
+      setCheckedItems(dissocPath(path, checkedItems))
+      onChange(dissocPath(path, checkedItems))
+    } else {
+      setCheckedItems(assocPath(path, true, checkedItems))
+      onChange(assocPath(path, true, checkedItems))
+    }
```
xantorres: All this `if`-`else` statement could be replaced by:
```
const newCheckedItems = ifElse(
	hasPath(path), 
	dissocPath(path), 
	assocPath(path, true)
)(checkedItems)
setCheckedItems(newCheckedItems)
onChange(newCheckedItems)
```
This way you don't call the same functions twice.

src/app/plugins/kubernetes/components/rbac/RbacChecklist.js
```diff
@@ -0,0 +1,113 @@
...
+  }, [checkedItems])
+
+  const renderVerb = useCallback(curry((groupName, resourceName, verb) =>
+    <FormControlLabel
+      key={verb}
+      control={
+        <Checkbox
+          checked={hasPath([groupName, resourceName, verb], checkedItems)}
+          onChange={(e) => handleCheck(groupName, resourceName, verb, e.target.checked)}
```
xantorres: you don't need to pass `e.target.checked` here

src/app/plugins/kubernetes/components/rbac/actions.js
```diff
@@ -0,0 +1,215 @@
...
+  return apiResources
+}, {
+  indexBy: 'clusterId',
+  uniqueIdentifier: 'id',
+})
+
+createContextLoader('apiResources', async ({ clusterId, apiGroup }) => {
+  const response = await qbert.getApiResourcesList({ clusterId, apiGroup })
+  const apiResources = response.resources.map((item) => {
+    item.id = `${item.name}-${clusterId}`
```
xantorres: Please avoid mutating data, use the following instead:
```
return {
	...item,
	id: `${item.name}-${clusterId}`
}
```

src/app/plugins/kubernetes/components/rbac/actions.js
```diff
@@ -0,0 +1,215 @@
...
+  uniqueIdentifier: 'id',
+})
+
+const clusterApiGroupsCacheKey = 'apiGroups'
+export const apiGroupsLoader = createContextLoader(clusterApiGroupsCacheKey, async ({ clusterId }) => {
+  const response = await qbert.getApiGroupList(clusterId)
+  const apiGroups = response.groups.map((item) => {
+    item.id = `${item.name}-${clusterId}`
+    return item
+  })
```
xantorres: Same here

src/app/plugins/kubernetes/components/rbac/actions.js
```diff
@@ -0,0 +1,215 @@
...
+  return apiGroups
+}, {
+  dataMapper: async (apiGroups, { clusterId }, loadFromContext) => {
+    const coreResourcesResponse = await loadFromContext('coreApiResources', { clusterId })
+    const coreResources = {
+      name: 'core',
+      groupVersion: 'core',
+      resources: coreResourcesResponse,
+    }
+    const otherResources = await mapAsync(async apiGroup => {
```
xantorres: The naming here is a bit off, why is this called `otherResources` when is clearly `apiGroups` what's being mapped here?
`apiGroupsWithResources` would be better

src/app/plugins/kubernetes/components/rbac/actions.js
```diff
@@ -0,0 +1,215 @@
...
+    return items.map(item => ({
+      ...item,
+      id: pathStr('metadata.uid', item),
+      name: pathStr('metadata.name', item),
+      namespace: pathStr('metadata.namespace', item),
+      clusterName: pipe(find(propEq('uuid', item.clusterId)), prop('name'))(clusters),
+      created: pathStr('metadata.creationTimestamp', item),
+    }))
+  },
+  uniqueIdentifier,
```
xantorres: It seems `indexBy` option is missing here

src/app/plugins/kubernetes/components/rbac/actions.js
```diff
@@ -0,0 +1,215 @@
...
+    const clusters = await loadFromContext(clustersCacheKey, { healthyClusters: true })
+    return items.map(item => ({
+      ...item,
+      id: pathStr('metadata.uid', item),
+      name: pathStr('metadata.name', item),
+      clusterName: pipe(find(propEq('uuid', item.clusterId)), prop('name'))(clusters),
+      created: pathStr('metadata.creationTimestamp', item),
+    }))
+  },
+  uniqueIdentifier,
```
xantorres: Missing `indexBy`

src/app/plugins/kubernetes/components/rbac/actions.js
```diff
@@ -0,0 +1,215 @@
...
+    const { qbert } = ApiClient.getInstance()
+    const item = currentItems.find(propEq('id', id))
+    if (!item) {
+      throw new Error(`Unable to find role with id: ${id}`)
+    }
+    const { clusterId, name } = item
+    await qbert.deleteClusterRole(clusterId, name)
+  },
+  dataMapper: async (items, params, loadFromContext) => {
+    const clusters = await loadFromContext(clustersCacheKey, { healthyClusters: true })
```
xantorres: Here you are passing a `healthyClusters: true` option, but in the `listFn` you are missing it, so potentially you will have issues when data mapped here belongs to a cluster that hasn't healthy clusters.

src/app/plugins/kubernetes/components/common/UserMultiSelect.js
```diff
@@ -0,0 +1,41 @@
+import React, { forwardRef } from 'react'
+import PropTypes from 'prop-types'
+import { projectAs, sortByProperty } from 'utils/fp'
+import { ValidatedFormInputPropTypes } from 'core/components/validatedForm/withFormContext'
+import useDataLoader from 'core/hooks/useDataLoader'
+import MultiSelect from 'core/components/MultiSelect'
+import { mngmUserActions } from 'k8s/components/userManagement/users/actions'
+
+const UserPicker = forwardRef(({ onChange, ...rest }, ref) => {
+  const [users, loadingUsers] = useDataLoader(mngmUserActions.list)
```
xantorres: use `defaultOrderBy: 'username'` option when using `createCRUDActions`, or you can just pass a `{ orderBy: 'username }` param such as: `useDataLoader(mngmUserActions.list, { orderBy: 'username' })`

src/app/plugins/kubernetes/components/prometheus/actions.js
```diff
@@ -62,6 +62,7 @@ export const prometheusInstanceActions = createCRUDActions(prometheusInstancesCa
     await qbert.deletePrometheusInstance(instance.clusterUuid, instance.namespace, instance.name)
   },
   dataMapper: async (items, params, loadFromContext) => {
+    console.log(items)
```
xantorres: remove this

src/app/plugins/kubernetes/components/rbac/AddClusterRoleBindingPage.js
```diff
@@ -0,0 +1,72 @@
...
+import { clusterRoleBindingsCacheKey, clusterRoleBindingActions } from './actions'
+import useParams from 'core/hooks/useParams'
+import useReactRouter from 'use-react-router'
+import useDataUpdater from 'core/hooks/useDataUpdater'
+
+const defaultParams = {}
+export const AddClusterRoleBindingForm = () => {
+  const { params, getParamsUpdater } = useParams(defaultParams)
+  const { history } = useReactRouter()
+  const onComplete = () => history.push('/ui/kubernetes/rbac#clusterRoleBindings')
```
xantorres: you might want to not redirect if there has been an error when submitting, you can do so by using the first parameter passed to the `onComplete` function:
```
const onComplete = success => success && history.push('/ui/kubernetes/rbac#clusterRoleBindings')
```

src/app/plugins/kubernetes/components/rbac/AddClusterRoleBindingPage.js
```diff
@@ -0,0 +1,72 @@
...
+  const [createClusterRoleBindingAction] = useDataUpdater(clusterRoleBindingActions.create, onComplete)
```
xantorres:
```
const [createClusterRoleBindingAction, creatingClusterRoleBinding] = ...
```
and then wrap everything in a `FormWrapper`
```
<FormWrapper
    title="New Cluster Role Binding"
    loading={creatingClusterRoleBinding}
    backUrl={listUrl}>
```
This way you have 2 new features:
* The user will be able to go back at any time (FormWrapper includes a close button to go back)
* The form will be locked while the create operation is in progress

src/app/plugins/kubernetes/components/rbac/RoleBindingsListPage.js
```diff
@@ -0,0 +1,51 @@
...
+import createCRUDComponents from 'core/helpers/createCRUDComponents'
+import ClusterPicklist from 'k8s/components/common/ClusterPicklist'
+import useDataLoader from 'core/hooks/useDataLoader'
+import { roleBindingsCacheKey, roleBindingActions } from './actions'
+import { createUsePrefParamsHook } from 'core/hooks/useParams'
+import { listTablePrefs } from 'app/constants'
+import { pick } from 'ramda'
+
+const defaultParams = {
+  masterNodeClusters: true,
```
xantorres: do we really want to pass `masterNodeClusters`  param to `roleBindingActions`, or is just a copy & paste leftover?

src/app/plugins/kubernetes/components/rbac/RolePicklist.js
```diff
@@ -0,0 +1,62 @@
...
+import useDataLoader from 'core/hooks/useDataLoader'
+import PropTypes from 'prop-types'
+import { isEmpty, propOr, head } from 'ramda'
+import { allKey } from 'app/constants'
+import { roleActions, clusterRoleActions } from './actions'
+
+// We need to use `forwardRef` as a workaround of an issue with material-ui Tooltip https://github.com/gregnb/mui-datatables/issues/595
+const RolePicklist = forwardRef(
+  ({ clusterId, loading, onChange, selectFirst, showAllRoleTypes, ...rest }, ref) => {
+    const [roles, rolesLoading] = useDataLoader(roleActions.list, { clusterId })
```
xantorres: why are you passing `clusterId` to this data loader? Is this filtering by `clusterId` whatsoever?
It seems to me that `roleActions` and `clusterRoleActions` could be merged and just use `params` to filter by `clusterId` (if needed)

src/app/plugins/kubernetes/components/rbac/RolePicklist.js
```diff
@@ -0,0 +1,62 @@
...
+    const [clusterRoles, clusterRolesLoading] = useDataLoader(clusterRoleActions.list, { clusterId })
+
+    const options = useMemo(() => {
+      // Pass in Role/ClusterRole in value to know which type of role
+      // was selected when forming the request body
+      const mappedRoles = roles.map(role => ({
+        label: `Role: ${role.name}`,
+        value: `Role:${role.name}`
+      }))
+      const mappedClusterRoles = clusterRoles.map(role => ({
```
xantorres: as I said, you should have just `roleActions` and do all of this logic there, using the provided `params`, views should not be responsible of any data manipulation/parsing

src/app/plugins/kubernetes/components/rbac/RoleBindingsListPage.js
kevinylee1991: i think it's a copy/paste

src/app/plugins/kubernetes/components/rbac/RolePicklist.js
kevinylee1991: I only want to display the roles that are present on the selected cluster

PR #423, opened 2019-10-29
Add Tenant page and delete Tenant action

xantorres:
... and ListTable performance improvements + added `Component` option in the columns definition to allow rendering a custom component inside of each cell.

### DIFF
--- a/src/api-client/Keystone.js
+++ b/src/api-client/Keystone.js
@@ -1,17 +1,18 @@
 import axios from 'axios'
 import { pluck } from 'ramda'
 import { getHighestRole } from './helpers'
+import { pathJoin } from 'utils/misc'
 
 const authConstructors = {
   password: (username, password) => ({
     user: {
       name: username,
       domain: { id: 'default' },
-      password
-    }
+      password,
+    },
   }),
 
-  token: token => ({ id: token })
+  token: token => ({ id: token }),
 }
 
 const constructAuthBody = (method, ...args) => {
@@ -19,9 +20,9 @@ const constructAuthBody = (method, ...args) => {
     auth: {
       identity: {
         methods: [method],
-        [method]: authConstructors[method](...args)
+        [method]: authConstructors[method](...args),
       },
-    }
+    },
   }
 
   return body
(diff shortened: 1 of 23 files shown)


src/app/core/components/validatedForm/ValidatedForm.js
```diff
@@ -55,9 +55,13 @@ class ValidatedForm extends PureComponent {
    * Note: many components use event.target.value but we only need value here.
    * Note: values can be passed up to parent component by supplying a setContext function prop
    */
-  setFieldValue = moize(field => (value, validateAll) => {
-    this.setState(setStateLens(value, ['values', field]),
-      () => {
+  setFieldValue = moize(field => {
+    const fieldLens = lensPath(['values', field])
+    return (updaterFn, validateAll) => {
+      const stateUpdater = typeof updaterFn === 'function'
```
brennancheung: This style of functional polymorphism might not work well when we want to add Typescript typing information.  I know Typescript has the `|` (union types) but it is inconsistent.  It's called `updaterFn` but may not be a function at all so that is also confusing.

What about passing in an object that has `updaterFn` or `updaterSpec`?

It's also not clear what the format is for `updaterFn` in either form.  If know I need to give it a function I still don't know what is it being passed and what am I supposed to return.  If it's not a function, I don't know the type / format I'm supposed to pass in either.

src/app/core/components/validatedForm/ValidatedForm.js
```diff
@@ -71,8 +75,12 @@ class ValidatedForm extends PureComponent {
         if (this.props.setContext) {
           this.props.setContext(this.state.values)
         }
-      },
-    )
+      })
+    }
+  })
+
+  getFieldValue = moize(field => (getterFn = identity) => {
```
brennancheung: This doesn't look expensive enough to warrant memoization.

src/app/plugins/kubernetes/components/userManagement/RolesPicklist.js
```diff
@@ -8,21 +8,27 @@ import { allKey } from 'app/constants'
...
+}, ref) => {
+  const [roles, rolesLoading] = useDataLoader(mngmRoleActions.list, { allRoles })
   const options = useMemo(() => projectAs(
     { label: 'name', value: 'id' }, roles,
   ), [roles])
 
   // Select the first role as soon as roles are loaded
   useEffect(() => {
-    if (!isEmpty(options) && selectFirst) {
+    if (!isEmpty(options) && selectFirst && !disabled && !value) {
```
brennancheung: You can use de Morgan's Law to simplify.  Also, move the function to the end so it is only called if it really needs to be.

```javascript
if (selectFirst && !(disabled || value || isEmpty(options)))
```

src/app/core/components/listTable/ListTable.js
```diff
@@ -121,7 +122,7 @@ class ListTable extends PureComponent {
     }
   }
 
-  handleClick = row => event => {
+  handleClick = moize(row => event => {
```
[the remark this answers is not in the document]
xantorres: Memoization is important in this case, to prevent any issues with the extra memory space as you mention I passed a `maxSize` of 100 to just memoize the last 100 handlers.

src/api-client/Keystone.js
```diff
@@ -85,10 +100,31 @@ class Keystone {
   }
 
   getAllTenantsAllUsers = async () => {
-    const response = await axios.get(this.tenantsUrl, this.client.getAuthHeaders())
+    const response = await axios.get(this.allTenantsAllUsersUrl, this.client.getAuthHeaders())
     return response.data.tenants
   }
 
+  addUserRole = async ({ tenantId, userId, roleId }) => {
+    const response = await axios.put(pathJoin(
```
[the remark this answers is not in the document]
xantorres: `pathJoin` just joins and array of parameters with `/`, and removing duplicated slashes. It adds some readability over string interpolation in long urls. In this case the string was quite long so I used it but probably it would be better to use it for all the url parts and not just the `projectUrl`

src/app/core/components/validatedForm/ValidatedForm.js
xantorres: Is not about expensiveness, is always about preventing re-renderings when passing down the function as a prop

src/app/core/components/validatedForm/ValidatedForm.js
xantorres: 99% of the times we will just pass a value here, but in some complex cases we might want to pass an updater function instead (eg, when the value is an object and we just want to update an specific part of that object), so being able to pass an updater function (that will receive the current value) instead of a plain value gives a lot of flexibility. 
But I will split this into two functions so there won't be any chance of confusion.

src/api-client/Keystone.js
```diff
@@ -52,21 +53,35 @@ class Keystone {
   }
 
   get endpoint () { return this.client.options.keystoneEndpoint }
+
```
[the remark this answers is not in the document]
xantorres: I almost fully tamed Webstorm to not do things like this but it still misbehaves sometimes. In this case, thought, I kind of liked it so I decided to keep it.

PR #424, opened 2019-10-30
Refactor user management and edit tenant

xantorres:


src/app/plugins/kubernetes/components/userManagement/users/actions.js
```diff
@@ -0,0 +1,86 @@
...
+      id: user.id,
+      username: user.name,
+      displayname: user.displayname,
+      email: user.email,
+      defaultProject: user.default_project_id,
+      twoFactor: pipe(
+        find(propEq('user_id', user.id)),
+        when(isNil, always(emptyObj)),
+        propEq('type', 'totp'),
+        mfaEnabled => mfaEnabled ? 'enabled' : 'disabled',
```
brennancheung: `util/misc.js` has `castBoolToStr` for this.

`castBoolToStr('enabled', 'disabled')`

src/app/plugins/kubernetes/components/userManagement/groups/actions.js
```diff
@@ -0,0 +1,54 @@
...
+      const groupRules = mappingRules.reduce((groupRules, rule) => {
+        if (any(pathEq(['group', 'id'], group.id), rule.local)) {
+          // Remove FirsName & LastName mapping from remote attribute array.
+          return groupRules.concat(rule.remote.slice(2))
+        }
+        return groupRules
+      }, emptyArr)
+      // Stringify the results
+      const samlAttributesString = groupRules.reduce((samlAttributes, rule) => {
+        if (rule.hasOwnProperty('any_one_of')) {
```
brennancheung: Might be able to clean this up more.  Looks like there is a lot of repetition between these.

Also, do we know that `any_one_of` and `not_any_of` are mutually exclusive?  The way the code is written right now it is either or.  Just wondering if that is a valid assumption?

xantorres: this convoluted logic is basically translated from the old UI, it was still harder to understand, but I think its 100% accurate with what they were doing there.

PR #425, opened 2019-10-31
Misc fixes for AWS cluster creation submit

brennancheung:
There are numerous errors when trying to submit the POST call.  This fixes some of them.

There are a few more that are still needed.  I will remove the WIP once those are fixed as well.

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/AddAwsClusterPage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/AddAwsClusterPage.js
@@ -34,12 +34,14 @@ const initialContext = {
   numWorkers: 1,
   enableCAS: false,
   usePf9Domain: true,
-  network: 'newVpc',
+  network: 'newPublic',
   containersCidr: '10.20.0.0/16',
   servicesCidr: '10.21.0.0/16',
   networkPlugin: 'flannel',
   mtuSize: 1440,
   runtimeConfigOption: 'default',
+  isPrivate: false,
+  internalElb: false,
 }
 
 const templateOptions = [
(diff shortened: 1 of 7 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/AddAwsClusterPage.js
```diff
@@ -243,6 +246,11 @@ const renderCustomNetworkingFields = ({ params, getParamsUpdater, values, setFie
   )
 }
 
+const alphaNumericAndHyphensOnly = s => {
```
xantorres: Move this to `misc.js` as we already have some string manipulation functions there.

src/app/plugins/kubernetes/components/infrastructure/clusters/AddAwsClusterPage.js
```diff
@@ -251,10 +259,15 @@ const AddAwsClusterPage = () => {
   }
   const [create, loading] = useDataUpdater(clusterActions.create, onComplete)
 
+  // It's a little weird to load cloud providers here but React won't let me use a hook inside of handleSubmit.
+  // They can't get to the submit button without first setting a cloud provider so in theory
+  // this should already exist before handleSubmit is called.
+  const [cloudProviders] = useDataLoader(cloudProviderActions.list)
```
xantorres: you should probably load the `cloudProviders` inside of the `clustersActions.create` function
In any case in my PR (https://github.com/platform9/pf9-ui-plugin/pull/426) you can check some of the possibilities you have.

brennancheung: I'll see if I can cleanly refactor it to do this.

PR #428, opened 2019-11-01
Misc fixes and edit Tenant dialog

xantorres:
* Styling adjustments
* Fix warning about "compact" prop
* Fix misc ListTable issues
* Edit Tenant feature

### DIFF
--- a/src/app/plugins/kubernetes/components/userManagement/tenants/actions.js
+++ b/src/app/plugins/kubernetes/components/userManagement/tenants/actions.js
@@ -1,8 +1,11 @@
 import ApiClient from 'api-client/ApiClient'
-import { pluck, pipe, map, prop, find, propEq, filter } from 'ramda'
+import { keys, pluck, pipe, prop, find, propEq, filter, always, isNil, reject, omit } from 'ramda'
 import { namespacesCacheKey } from 'k8s/components/namespaces/actions'
 import createCRUDActions from 'core/helpers/createCRUDActions'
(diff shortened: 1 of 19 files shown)


src/app/plugins/kubernetes/components/userManagement/tenants/actions.js
```diff
@@ -13,33 +15,84 @@ export const filterValidTenants = tenant => !reservedTenantNames.includes(tenant
...
+    const updateTenantPromise = keystone.updateProject(tenantId, {
+      ...currentTenant,
+      name,
+      description,
+    })
+    const prevRoleAssignments = prevRoleAssignmentsArr.reduce((acc, roleAssignment) => ({
+      ...acc,
+      [pathStr('user.id', roleAssignment)]: pathStr('role.id', roleAssignment),
+    }), {})
+    const updateUserRolesPromises = keys({ ...prevRoleAssignments, ...roleAssignments }).reduce(
```
brennancheung: This logic is a bit hard to follow what is going on here.  Can this be broken down more into smaller functions or approached a different way?

Also, if you are just appending to `operations` each time, would a `map` make more sense here?

xantorres: Not all times, in case there is nothing to update I won't add any promise to the accumulator.
This logic is convoluted because is really a bit complex, there's no way around this, I have to add/remove users and roles individual depending on what's selected. I will try to improve variable naming and split it into small functions though.

PR #434, opened 2019-11-05
Add/Edit User pages (Tenants & Users section)

xantorres:
Also fixes some issues and errors in ValidatedForm and validations

TODO: dev server api resources and action create/update functions

image

### DIFF
--- a/src/app/core/utils/fieldValidators.js
+++ b/src/app/core/utils/fieldValidators.js
@@ -20,7 +20,7 @@ export const customValidator = (validator, errorMessage) =>
   new FieldValidator(validator, errorMessage)
 
 const fieldIsUnset = value => isNil(value) || isEmpty(value) || value === false
-export const hasMinLength = moize(minLen => both(is(String), val => val.length >= minLen))
+export const hasMinLength = minLen => both(is(String), val => val.length >= minLen)
 export const hasOneLowerChar = both(is(String), test(/[a-z]/))
 export const hasOneUpperChar = both(is(String), test(/[A-Z]/))
 export const hasOneNumber = both(is(String), test(/[0-9]/))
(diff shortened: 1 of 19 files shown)


src/app/core/utils/fieldValidators.js
```diff
@@ -49,32 +54,50 @@ export const lengthValidator = (minLength, maxLength) =>
...
+    hasOneUpperChar,
+    hasOneNumber,
+    hasOneSpecialChar,
+  ])(value),
+  cond([
+    [complement(hasMinLength(8)), always('Password must be at least 8 characters long')],
+    [complement(hasOneLowerChar), always('Password must contain at least one lowercase letter')],
+    [complement(hasOneUpperChar), always('Password must contain at least one uppercase letter')],
+    [complement(hasOneNumber), always('Password must contain at least one number')],
+    [complement(hasOneSpecialChar), always('Password must contain at least one special character')],
```
brennancheung: Might be able to write a small function to simplify some of this repetition and remove the `complement` and `always` that add a little clutter.  Also, the list is being repeated twice so probably only need to define it once.

src/app/core/utils/fieldValidators.js
```diff
@@ -20,27 +20,32 @@ export const customValidator = (validator, errorMessage) =>
   new FieldValidator(validator, errorMessage)
 
 const fieldIsUnset = value => isNil(value) || isEmpty(value) || value === false
+export const hasMinLength = moize(minLen => both(is(String), val => val.length >= minLen))
```
brennancheung: Do we **really** need to use `moize` here:interrobang:

src/app/core/utils/fieldValidators.js
xantorres: The second time is a switch to show a different message depending on the error

src/app/core/utils/fieldValidators.js
xantorres: It doesn't hurt

brennancheung: Creating this function is cheap.  Probably faster than the time it takes to do the extra memory lookup from memoization.  Modern CPU's can execute anywhere from 50-200 instructions in the time it takes to do a single memory fetch operation depending on if it is in L1/L2/main memory/TLB miss.  Even if it was slower to create than perform the lookup, we don't need to optimize things that are going to have less than a nanosecond of difference to the end user.

http://wiki.c2.com/?PrematureOptimization

xantorres: as I always say, this is not about creating the function cost, is about passing new props to children components that causes React engine to re-render, and that always add some overhead, specially in large forms with a lot of inputs and sub-components, so micro optimizations and memoization techniques like this are the way to go

xantorres: anyway, I have checked, and it seems I'm not passing down the result of the memoized function as a prop (I'm using it to check if the password had a minimal length), so it is not be needed at all in this case.

PR #438, opened 2019-11-07
Remove service monitor selector field from table view

smanpathak:
This change removes the service monitor selector field from table view.
The selector field is of kubernetes API type: metav1.LabelSelector (https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#servicemonitorspec)
which in itself can be of type MatchLabels or MatchExpressions (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#labelselector-v1-meta)

MatchLabels is simple key value array (which was handled)
MatchExpressions has operator format, which is more complex to display.

I believe for simplicity, we can omit this field from table view altogether.

### DIFF
--- a/src/app/plugins/kubernetes/components/prometheus/PrometheusServiceMonitors.js
+++ b/src/app/plugins/kubernetes/components/prometheus/PrometheusServiceMonitors.js
@@ -1,3 +1,4 @@
+import React from 'react'
 import createCRUDComponents from 'core/helpers/createCRUDComponents'
 import { prometheusServiceMonitorsCacheKey } from './actions'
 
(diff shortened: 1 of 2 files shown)


src/app/plugins/kubernetes/components/prometheus/PrometheusServiceMonitors.js
```diff
@@ -10,8 +10,7 @@ export const columns = [
   { id: 'clusterName', label: 'Cluster' },
   { id: 'namespace', label: 'Namespace' },
   { id: 'labels', label: 'Labels', render: renderKeyValues },
-  { id: 'port', label: 'Port' },
-  { id: 'selector', label: 'Selector', render: renderKeyValues },
```
xantorres: What about setting the ``display` prop to `false` instead of removing the column altogether? That way the column will be hidden by default but we still let the user choose to display the column if he wants to do so (via the gear button)

brennancheung: I'm in agreement with Xan here.  If that field even provides a small amount of value to someone, we can add `display: false` to the column definition.  That way advanced users can add the column to the table if they want.

smanpathak: As I explained in change description, the selector value can be complex. Not just key value pairs.
e.g. it can be in formats like
(KV example) ----
app: frontend
team: dev
---

(MatchExpression example)---
team in [dev, test, production]
app not in [db, middleware, ops]
---

brennancheung: In that case I think using `JSON.stringify(value)` or `JSON.stringify(value, null, 4)` if you want it to be pretty printed would be a good option.

smanpathak: @brennancheung I have added processing for both type of values in latest commit.

PR #439, opened 2019-11-07
Feature/forgot password functionality

rahulkurumkar:
Added functionality to reset the user password.


src/app/plugins/openstack/components/ForgotPasswordPage.js
```diff
@@ -0,0 +1,120 @@
...
+
+    setLoading(true)
+    const body = { username: emailId }
+    try {
+      const response = await axios.post(forgotPasswordRequestUrl, body)
+
+      if (response.status === 200) {
+        setLoading(false)
+        setIsResetSuccessful(true)
+      }
```
xantorres: if for any reason `response.status` is not 200, we will show an infinite loading spinner because you never call `setLoading(false)` nor `setIsError()`

src/app/constants.js
```diff
@@ -3,6 +3,8 @@ export const defaultUniqueIdentifier = 'id'
 export const appUrlRoot = '/ui'
 export const imageUrlRoot = `${appUrlRoot}/images`
 export const loginUrl = `${appUrlRoot}/openstack/login`
+export const forgotPasswordUrl = `${appUrlRoot}/openstack/forgot_password`
+export const forgotPasswordRequestUrl = '/clemency/request'
```
xantorres: no base url here?

src/app/plugins/openstack/components/ForgotPasswordPage.js
```diff
@@ -0,0 +1,120 @@
...
+  },
+})
+
+const ForgotPasswordPage = props => {
+  const { classes } = props
+  const { history } = useReactRouter()
+  const [emailId, setEmailId] = useState('')
+  const [loading, setLoading] = useState(false)
+  const [isError, setIsError] = useState(false)
+  const [isResetSuccessful, setIsResetSuccessful] = useState(false)
```
xantorres: you should combine these `useState` into one ie `useState({ loading: false, isError: null, isResetSuccessful: false, emailId: ''}` (even better if you initialize the state object outside of the render body) to prevent excessive re-renders and unexpected issues that might arise because of it

src/app/constants.js
rahulkurumkar: The base URL is not required in this case because "Clemency" is altogether a different service.

src/app/plugins/openstack/components/ForgotPasswordPage.js
rahulkurumkar: Will handle this case

src/app/plugins/openstack/components/ForgotPasswordPage.js
rahulkurumkar: Will look into it

src/app/plugins/openstack/components/ForgotPasswordPage.js
```diff
@@ -0,0 +1,143 @@
...
+      updateParams({
+        loading: false,
+        isResetSuccessful: false,
+        isError: true
+      })
+    } finally {
+      updateParams({
+        loading: false,
+        isResetSuccessful: true
+      })
```
xantorres: I think we should get rid of this `finally`, it's better not to call 2 state updates in the same line of execution, also this is quite confusing

src/app/plugins/openstack/components/ForgotPasswordPage.js
```diff
@@ -0,0 +1,143 @@
...
+      <div className="forgot-password-page">
+        <Grid container justify="center" className={classes.root}>
+          <Grid item md={4} lg={3}>
+            <Paper className={classes.paper}>
+              <img src="/ui/images/logo-color.png" className={classes.img} />
+              <form className={classes.form} onSubmit={handleFormSubmit}>
+                <Typography variant="subtitle1" align="center">
+                    Password Reset
+                </Typography>
+                {!params.isResetSuccessful ? (<><TextField
```
brennancheung: Too many lines between the `?` and `:` symbols of the ternary expression.  Might be useful to make a small render function for some of this.  Also, multi-line tags should be indented such that the opening and closing tag are on the same column.

src/app/plugins/openstack/components/SessionManager.js
```diff
@@ -32,6 +34,10 @@ const SessionManager = withRouter(props => {
     const username = user && user.username
     let unscopedToken = tokens && tokens.unscopedToken
 
+    if (location.pathname === forgotPasswordUrl) {
```
brennancheung: Can the forgot password page just be an ordinary route inside of `openstack/index.js`?  It seems like you are manually implementing route handling here instead of just using `react-router`.

src/app/plugins/openstack/components/ForgotPasswordPage.js
rahulkurumkar: Incorporated review changes

src/app/plugins/openstack/components/SessionManager.js
rahulkurumkar: I think this code is executed before routes are created

src/app/plugins/openstack/components/ForgotPasswordPage.js
rahulkurumkar: Fixed

src/app/plugins/openstack/components/ForgotPasswordPage.js
```diff
@@ -0,0 +1,160 @@
...
+      newState =
+        response.status === 200
+          ? { isResetSuccessful: true }
+          : { isError: true }
+    } catch (err) {
+      newState = {
+        isError: true
+      }
+    } finally {
+      updateParams({ ...newState, loading: false })
```
xantorres: this `finally` makes the code a bit more confusing and harder to follow that it should be. Just having `try` and `catch` with its own `updateParams` (even if you duplicated `loading: false` would make things easier to read

src/app/plugins/openstack/components/ForgotPasswordPage.js
```diff
@@ -62,75 +63,91 @@ const ForgotPasswordPage = props => {
...
+                      type="email"
+                      className={classes.textField}
+                      onChange={handleEmailChange()}
+                    />
+                    {params.isError && (
+                      <div className={classes.errorContainer}>
+                        <Alert variant="error" message="Something went wrong" />
+                      </div>
+                    )}
+                    <SubmitButton label="RESET MY PASSWORD" />
```
brennancheung: Don't use all caps here.  This can be done in CSS and if we every want to change the styling it's better if it isn't hard-coded to caps.

src/app/plugins/openstack/components/ForgotPasswordPage.js
brennancheung: Agreed with Xan here.  Any time you can avoid branching it's almost always a win.  https://medium.com/better-programming/why-you-need-to-stop-using-else-statements-5b1fd09dea9e

rahulkurumkar: Incorporated review changes

src/app/plugins/openstack/components/ForgotPasswordPage.js
rahulkurumkar: Handled through CSS

src/app/plugins/openstack/components/ForgotPasswordPage.js
```diff
@@ -0,0 +1,163 @@
...
+
+    try {
+      const response = await axios.post(forgotPasswordApiUrl, body)
+
+      newState =
+        response.status === 200
+          ? { isResetSuccessful: true }
+          : { isError: true }
+
+      updateParams({ ...newState, loading: false })
```
xantorres: I guess now you can get rid of `newState` variable and set the values directly.

src/app/plugins/openstack/components/ForgotPasswordPage.js
```diff
@@ -0,0 +1,163 @@
...
+    } catch (err) {
+      newState = {
+        isError: true
+      }
+
+      updateParams({ ...newState, loading: false })
```
xantorres: same here, just set the `isError: true` directly

src/app/plugins/openstack/components/ForgotPasswordPage.js
```diff
@@ -0,0 +1,163 @@
...
+              </form>
+            </Paper>
+          </Grid>
+        </Grid>
+      </div>
+    </Progress>
+  )
+}
+
+export default compose(withStyles(styles))(ForgotPasswordPage)
```
xantorres: please use `useStyles` hook instead of this HoC, look at our existing views to see an example

src/app/plugins/openstack/components/ForgotPasswordPage.js
rahulkurumkar: Fixed

src/app/plugins/openstack/components/ForgotPasswordPage.js
rahulkurumkar: Fixed it

src/app/plugins/openstack/components/ForgotPasswordPage.js
rahulkurumkar: Done

xantorres:
Looks good so far. 
Could you please rebase and squash this PR into 2-3 commits? 13 commits looks like an overkill for just a bunch of files and lines changed.

xantorres:
now it seems there are 22 commits instead of 3

PR #443, opened 2019-11-08
Duplicated cache items on update

xantorres:
Thanks @kevinylee1991 for spotting this issue

### DIFF
--- a/src/app/core/helpers/createContextUpdater.js
+++ b/src/app/core/helpers/createContextUpdater.js
@@ -1,6 +1,6 @@
 import {
   hasPath, path, assocPath, pathEq, always, over, append, lensPath, identity, isNil, pipe, pickAll,
-  reject, either, equals, mergeLeft, allPass, map, head, split, __,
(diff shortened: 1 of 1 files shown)


src/app/core/helpers/createContextUpdater.js
```diff
@@ -129,7 +129,7 @@ function createContextUpdater (cacheKey, dataUpdaterFn, options = {}) {
       pickAll(allIndexKeys),
       reject(either(isNil, equals(allKey))),
     )(params)
-    const eqIds = allPass(map(idPath => pathEq(idPath, __, params), uniqueIdentifierPaths))
+    const eqIds = allPass(map(idPath => pathEq(idPath, path(idPath, params)), uniqueIdentifierPaths))
```
brennancheung: Can you move everything inside the `map` into its own short 1-line function and give it a good name?  I think that will help make this easier to parse / understand.

Maybe add a comment on what is going on in this algorithm as well and what it is for.

PR #445, opened 2019-11-08
Kubeconfig Download Dialog UI

murilovmachado:
This PR implements the UI for kubeconfig download dialog.

Screen Shot 2019-11-08 at 20 25 59
Screen Shot 2019-11-08 at 20 26 13

### DIFF
--- a/src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
+++ b/src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useMemo } from 'react'
 import { makeStyles } from '@material-ui/core/styles'
 import createCRUDComponents from 'core/helpers/createCRUDComponents'
 import kubeConfigActions from './actions'
@@ -22,7 +22,7 @@ const KubeConfigListPage = () => {
   const [isDialogOpen, toggleDialog] = useToggler()
   const classes = useStyles()
 
-  const columns = getColumns(toggleDialog)
+  const columns = useMemo(() => getColumns(toggleDialog), [toggleDialog])
 
   const options = {
     cacheKey: 'kubeconfig',
--- a/src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
+++ b/src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
@@ -5,15 +5,12 @@ import kubeConfigActions from './actions'
 import ValidatedForm from 'core/components/validatedForm/ValidatedForm'
 import TextField from 'core/components/validatedForm/TextField'
 import DownloadDialog from './DownloadDialog'
+import SimpleLink from 'core/components/SimpleLink'
 
 const useStyles = makeStyles(theme => ({
   link: {
     display: 'block',
     width: 'fit-content',
-    textDecoration: 'none',
-    '&:hover': {
-      textDecoration: 'underline',
-    }
   },
   clusterConfig: {
     marginTop: theme.spacing(5),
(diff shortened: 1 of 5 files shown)


src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
```diff
@@ -44,18 +57,25 @@ const KubeConfigListPage = () => {
...
+    render: (contents, row) => kubeConfigLink(row, handleClickDownload),
+  },
   { id: 'url', label: 'URL' },
 ]
 
-// TODO: implement kubebonfig link
-const kubeConfigLink = (value) =>
-  <a href='#'>Download kubeconfig</a>
+const kubeConfigLink = (row, handleClickDownload) =>
+  <a onClick={() => handleClickDownload(row)}>Download kubeconfig</a>
```
xantorres: we have a component for links called `SimpleLink`

src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
```diff
@@ -44,18 +57,25 @@ const KubeConfigListPage = () => {
       </a>
```
xantorres: you could use our `SimpleLink`, so you won't need to use specific styles here

src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
```diff
@@ -28,6 +39,8 @@ const KubeConfigListPage = () => {
     name: 'Kubeconfig',
     compactTable: true,
     blankFirstColumn: true,
+    multiSelection: false,
+    onSelect: handleClickDownload,
```
xantorres: the Kubeconfig will be downloaded when the user clicks the row, is that intended? Also what is the point of allowing row selection if there are no actions defined at table level? You should probably disable row selection at all using `showCheckboxes: false`

src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
```diff
@@ -28,6 +39,8 @@ const KubeConfigListPage = () => {
     name: 'Kubeconfig',
     compactTable: true,
     blankFirstColumn: true,
+    multiSelection: false,
```
xantorres: even if `multiSelection: false`, it will still display radio buttons to allow individual row selection, which I think is pointless here as we don't have row actions at table level

src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
```diff
@@ -18,8 +21,16 @@ const useStyles = makeStyles(theme => ({
 }))
 
 const KubeConfigListPage = () => {
+  const [isDialogOpen, setDialogOpen] = useState(false)
```
xantorres: you can use `const [isDialogOpen, toggleDialog] = useToggler()` as a simplification of this, as it will allow you to get rid of both `handleClickDownload` and `handleDialogClose`

src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
```diff
@@ -18,8 +21,16 @@ const useStyles = makeStyles(theme => ({
...
   const classes = useStyles()
 
+  const handleClickDownload = (row) => {
+    setDialogOpen(true)
+  }
+  const handleDialogClose = () => setDialogOpen(false)
+
+  const columns = getColumns(handleClickDownload)
+
   const options = {
```
xantorres: as a minor optimization, these options could be inside of a `useMemo` to prevent reassigning a new instance of this object to the table if `handleClickDownload` (or `toggleDialog` if you use `useToggle`) hasn't changed, which should prevent the re-rendering of the whole table every time this component renders.

src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
murilovmachado: There are no actions defined for a row, but a download dialog will appear when the user clicks a row. The old UI behaves like this (also showing the checkbox)

src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
murilovmachado: The old UI shows radio buttons for each row (even though there aren't actions defined)

xantorres: maybe we should fix things that we see are pointless, I don't think there's need to copy unused features.

PR #446, opened 2019-11-09
Fix context updating by using returned output data rather than provided params

xantorres:


xantorres:
closing this as this approach will cause `delete` operation to stop working

PR #447, opened 2019-11-10
Added rbac edit wizards (still need to implement simulator)

kevinylee1991:
Plan to add simulator in a different PR


src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,130 @@
...
+                // Add verbs that belong to rule
+                rule.verbs.map(verb => {
+                  rbac = assocPath([apiGroup.name, resource.name, verb], true, rbac)
+                })
+              }
+            })
+          }
+        })
+      } else {
+        rule.apiGroups.map(apiGroupName => {
```
xantorres: Same thing applies here, don't use `map` if you are not using the return values.

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,130 @@
...
+                })
+              } else {
+                // Add verbs that belong to rule
+                rule.verbs.map(verb => {
+                  rbac = assocPath([apiGroup.name, resource.name, verb], true, rbac)
+                })
+              }
+            })
+          } else {
+            rule.resources.map(resourceName => {
```
xantorres: this is a bit confusing and redundant. You should have the `if (resourcesWildCard)` inside of the iteration to prevent the repetition of the `rule.resources.map(resourceName => {` and other nested loops and logic.

src/app/plugins/kubernetes/components/rbac/UpdateClusterRoleBindingPage.js
```diff
@@ -0,0 +1,77 @@
...
+  groups: [],
+}
+
+const UpdateClusterRoleBindingPage = () => {
+  const { match, history } = useReactRouter()
+  const clusterRoleBindingId = match.params.id
+  const onComplete = useCallback(
+    success => success && history.push('/ui/kubernetes/rbac#clusterRoleBindings'),
+    [history])
+  const [clusterRoleBindings] = useDataLoader(clusterRoleBindingActions.list, { clusterId: allKey })
```
xantorres: if you pass the `clusterId` as a url param, we won't need to load the whole `clusterRoleBindingActions` list for each one of the clusters, which will be adding a huge impact on the loading time of this page. 
Also, please make use of the second returned parameter by the `useDataLoader` (ie `const [clusterRoleBindings, loadingClusterRoleBindings] = useDataLoader(...` to show a `Progress` component that will show a loading spinner for the user to know that something is being loaded. To do so you can simply add a loading prop to `FormWrapper` such as `<FormWrapper loading={loadingClusterRoleBindings} ...`

src/app/plugins/kubernetes/components/rbac/UpdateClusterRolePage.js
```diff
@@ -0,0 +1,63 @@
...
+import { allKey } from 'app/constants'
+import FormWrapper from 'core/components/FormWrapper'
+import PresetField from 'core/components/PresetField'
+
+// TODO: get default params for rbac property
+const defaultParams = {
+  rbac: {}
+}
+
+const UpdateClusterRolePage = () => {
```
xantorres: Comments on `UpdateClusterRoleBindingPage` applies to this page as well

src/app/plugins/kubernetes/components/rbac/UpdateRoleBindingPage.js
```diff
@@ -0,0 +1,80 @@
...
+import GroupMultiSelect from 'k8s/components/common/GroupMultiSelect'
+import FormWrapper from 'core/components/FormWrapper'
+import PresetField from 'core/components/PresetField'
+
+const defaultParams = {
+  users: [],
+  groups: [],
+}
+
+const UpdateRoleBindingPage = () => {
```
xantorres: Same here

src/app/plugins/kubernetes/components/userManagement/actions.js
```diff
@@ -0,0 +1,188 @@
+import ApiClient from 'api-client/ApiClient'
```
xantorres: This looks like a bad rebase/merge, I refactored this file to 4 files into its related subfolders, so it should not exist anymore.

src/app/utils/fp.js
```diff
@@ -185,9 +185,11 @@ export const updateInArray = curry((predicateFn, updateFn, arr) =>
 
 // Like `updateInArray` but stops after finding the element to update
 // Also like ramda `adjust` but using a predicateFn
-export const adjustWith = curry((predicateFn, updateFn, arr) =>
-  adjust(findIndex(predicateFn, arr), updateFn, arr),
-)
+export const adjustWith = curry((predicateFn, updateFn, arr) => {
+  console.log('adjustWith', arr)
+  console.log(findIndex(predicateFn, arr))
```
xantorres: Please remove this console.logs

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,132 @@
+import { assocPath } from 'ramda'
+
+// This function needs to be modernized, but can be
+// done perhaps after kubecon release
+const setRbacObject = (rules, apiGroups) => {
+  let rbac = {}
+  const wildCardRules = []
+
+  rules.map(rule => {
```
brennancheung: What is it you are trying to do here with all this code?  It's a bit confusing.  Document the before and after of the data in comments to make it clearer what is going on.

brennancheung: Perhaps split things up into separate smaller functions as well.  These whole thing is a giant confusing state machine.

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,132 @@
...
+    // UI does not have support for nonResourceURLs
+    if (rule.nonResourceURLs) {
+      return
+    }
+    rule.apiGroups.map(apiGroup => {
+      rule.resources.map(resource => {
+        rule.verbs.map(verb => {
+          if (apiGroup === '*' || resource === '*' || verb === '*') {
+            wildCardRules.push(rule)
```
brennancheung: Adding a `return` for early termination will save you the `else` branching down below.

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,132 @@
...
+          } else if (apiGroup === '') {
```
brennancheung: Branch at a lower point where the difference actually is.  You are repeating code unnecessarily here.

`rbac = assocPath([ apiGroup || 'core, resource, verb...`

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,132 @@
...
+        rule.apiGroups.map(apiGroupName => {
+          // Need to get full list of resources that belong to
+          // apiGroup in case of resources wildcard
+          const apiGroup = apiGroups.find(_apiGroup => (
+            _apiGroup.name === apiGroupName
+          ))
+          if (resourcesWildCard) {
+            // Add all resources that belong to apiGroup
+            apiGroup.resources.map(resource => {
+              if (verbsWildCard) {
```
brennancheung: This is very confusing here.  It looks like the only difference is whether you are using `resource` or `rule`.  If that is the case set a single variable to the appropriate one then collapse the duplicate code to just use that variable.

Also, you are using `map` here but you are using an assignment to a single variable `rbac` here so is this really an array or is there a better way to handle it.  I'm not really following the logic here so I'm not sure what alternatives to suggest.

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,132 @@
...
+                _resource.name === resourceName
+              ))
+              // If resource not in apiGroup, skip resource
+              if (!resource) {
+                return
+              }
+              if (verbsWildCard) {
+                // Add all verbs that belong to resource
+                resource.verbs.map(verb => {
+                  rbac = assocPath([apiGroup.name, resource.name, verb], true, rbac)
```
brennancheung: Same.  You are not branching at the appropriate place.  The key branch point is `resource` and `rule`.  Collapse that into a single point and then you don't need to duplicate the code paths from there.

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+}
+
+const processApiGroupResources = (apiGroup, rule, apiGroupsList) => {
+  const apiGroupFromList = apiGroupsList.find(_apiGroup => (
+    _apiGroup.name === apiGroup
+  ))
+  const resourceNames = apiGroupFromList.resources.map(resource => resource.name)
+  const resources = rule.resources.map(resource => {
+    if (resource === '*') {
```
brennancheung: This can be much shorter as a ternary operator.  I see you are returning either a `resource` or `resourceNames`.  Does this mean it is either a `string` or an `array`?  Caveat, assuming I'm correct, it might be less confusing and more consistent if you always returned an array.  That way you collapse branching and prevent code that consumes this function from needing to differentiate the 2 possible return values.  In general, a function should return just 1 type.

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,99 @@
+import { flatten, mergeDeepLeft, uniq } from 'ramda'
+
+const processApiGroup = (apiGroup, apiGroupsList) => {
+  if (apiGroup === '*') {
+    return apiGroupsList.map(group => group.name)
+  } else if (apiGroup === '') {
```
brennancheung: Don't need the `else` here because you already are doing a `return` above.

```javascript
if (apiGroup === '*') { return apiGroupsList.map(prop('name')) }
return apiGroup === '' ? 'core' : apiGroup

```

I do have some concern though that this function returns either a `string` or an `array` this makes using this function very difficult to use because the return type is inconsistent.  It forces that complexity onto the consumer of this function.

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,99 @@
...
+    if (resource === '*') {
+      return resourceNames
+    } else {
+      return resource
+    }
+  })
+  const flattenedResources = flatten(resources)
+  // Account for potential duplicates with wildcard + a specified resource
+  // Remove resource if it does not exist in the api group
+  const validatedResources = uniq(flattenedResources.filter(resource => {
```
brennancheung: Get rid of the braces and the `return`.  Just use the JS shorthand.

`uniq(flattenedResources.filter(resource => resourceName.includes(resource))`

You can also use the Ramda variant that uses partial application to make this cleaner:
`uniq(flattenedResources.filter(includes(resource))`

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,99 @@
...
+
+const processApiGroupVerbs = (item, rule, apiGroupsList) => {
+  const apiGroupFromList = apiGroupsList.find(_apiGroup => (
+    _apiGroup.name === item.apiGroup
+  ))
+  const resourcesWithVerbs = item.resources.map(resource => {
+    const resourceFromList = apiGroupFromList.resources.find(_resource => (
+      _resource.name === resource
+    ))
+    const verbs = rule.verbs.map(verb => {
```
brennancheung: Use ternary and JS shorthand syntax.  Same for a bunch of other places.

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,99 @@
...
+  // Account for potential duplicates with wildcard + a specified resource
+  // Remove resource if it does not exist in the api group
+  const validatedResources = uniq(flattenedResources.filter(resource => {
+    return resourceNames.includes(resource)
+  }))
+  return uniq(validatedResources)
+}
+
+const processApiGroupVerbs = (item, rule, apiGroupsList) => {
+  const apiGroupFromList = apiGroupsList.find(_apiGroup => (
```
brennancheung: `const apiGroupFromList = apiGroupsList.find(propEq('name', item.apiGroup))`

src/app/plugins/kubernetes/components/rbac/actions.js
```diff
@@ -235,6 +262,35 @@ export const roleBindingActions = createCRUDActions(roleBindingsCacheKey, {
 
     return qbert.createClusterRoleBinding(data.clusterId, data.namespace, body)
   },
+  updateFn: async data => {
+    const users = data.users.map((user) => {
```
brennancheung: Get rid of the `return`.  Use JS shorthand for these kinds of things.

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,99 @@
...
+    const resources = processApiGroupResources(apiGroup, rule, apiGroups)
+    return { apiGroup, resources }
+  })
+
+  const withVerbs = withResources.map(item => {
+    const apiGroupVerbs = processApiGroupVerbs(item, rule, apiGroups)
+    return { [item.apiGroup]: apiGroupVerbs }
+  })
+
+  const translatedRule = withVerbs.reduce((accum, current) => {
```
brennancheung: Ramda has a function called `mergeAll` that does this.

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
xantorres: or this just as 
```
pipe(
    flatten,
    filter(includes(resource)),
    uniq,
)(resources)
```

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,99 @@
...
+    // Account for potential duplicates with wildcard + a specified verb
+    // Remove verb if it does not exist in the resource
+    const validatedVerbs = uniq(flattenedVerbs.filter(verb => {
+      return resourceFromList.verbs.includes(verb)
+    }))
+    return { resource, verbs: validatedVerbs }
+  })
+  const verbsByResource = resourcesWithVerbs.reduce((accum, current) => {
+    const verbsObject = current.verbs.reduce((accum, verb) => {
+      return { ...accum, ...{ [verb]: true } }
```
xantorres: This can be simplified as just 
`const verbsObject = current.verbs.reduce((accum, verb) => ({ ...accum, [verb]: true }))`
or even
`const verbsObject = pipe(map(verb => [verb, true]), fromPairs)(current.verbs)`

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,99 @@
...
+    }, {})
+    const resourceVerbs = { [current.resource]: verbsObject }
+    return { ...accum, ...resourceVerbs }
+  }, {})
```
xantorres: actually, it seems like this whole verbsByResource could be processed by using ramda `groupBy` (https://ramdajs.com/docs/#groupBy)

src/app/plugins/kubernetes/components/rbac/SetRbacObject.js
```diff
@@ -0,0 +1,99 @@
...
+    if (current.nonResourceURLs) {
+      return accum
+    }
+    const newRules = processRule(current, apiGroups)
+    return mergeDeepLeft(accum, newRules)
+  }, {})
+  return rbac
+}
+
+export default setRbacObject
```
xantorres: Please `camelCase` the filename, as this is a function, not a Class or a react Component

src/app/core/helpers/createCRUDComponents.js
```diff
@@ -48,6 +48,7 @@ const createCRUDComponents = options => {
     editUrl,
     editCond,
     editDisabledInfo,
+    editWithClusterId,
```
xantorres: this `option` just feels wrong, this helper should be totally agnostic about clusters or any other business particularity.

kevinylee1991: Can't think of a good place to put it at the moment -- later change perhaps?

src/app/plugins/kubernetes/components/rbac/UpdateRoleBindingPage.js
```diff
@@ -0,0 +1,80 @@
...
+    success => success && history.push('/ui/kubernetes/rbac#roleBindings'),
+    [history])
+  const [roleBindings] = useDataLoader(roleBindingActions.list, { clusterId })
+  const roleBinding = useMemo(
+    () => roleBindings.find(propEq('id', roleBindingId)) || emptyObj,
+    [roleBindings, roleBindingId])
+  // const [update, updating] = useDataUpdater(roleBindingActions.update, onComplete)
+  const { params, updateParams, getParamsUpdater } = useParams(defaultParams)
+
+  const [updateRoleBindingAction] = useDataUpdater(roleBindingActions.update, onComplete)
```
xantorres: Please use the second received variable (loading) to block the UI showing a spinner when the update operation is in progress

src/app/core/helpers/createCRUDComponents.js
xantorres: We should really find an alternative way whenever possible before adding business code in helpers that are used in many places, because later it will be harder to refactor.

brennancheung: Agreed.  This really violates separation of concerns and this component already has too much business domain logic in it.

brennancheung: Worst case scenario you expose a function for generating the URL's that gets passed the relevant data.

src/app/plugins/kubernetes/components/rbac/UpdateClusterRolePage.js
```diff
@@ -0,0 +1,63 @@
...
+  const clusterId = match.params.clusterId
+  const onComplete = useCallback(
+    success => success && history.push('/ui/kubernetes/rbac#clusterRoles'),
+    [history])
+  const [clusterRoles] = useDataLoader(clusterRoleActions.list, { clusterId })
+  const clusterRole = useMemo(
+    () => clusterRoles.find(propEq('id', clusterRoleId)) || emptyObj,
+    [clusterRoles, clusterRoleId])
+  const [updateClusterRoleAction] = useDataUpdater(clusterRoleActions.update, onComplete)
+  const handleSubmit = params => data => updateClusterRoleAction({ ...data, ...params })
```
xantorres: no need to create a function with `params`, as this variable is already accessible from the outer context.

src/app/plugins/kubernetes/components/rbac/UpdateClusterRoleBindingPage.js
```diff
@@ -0,0 +1,77 @@
...
+    [history])
+  const [clusterRoleBindings] = useDataLoader(clusterRoleBindingActions.list, { clusterId })
+  const clusterRoleBinding = useMemo(
+    () => clusterRoleBindings.find(propEq('id', clusterRoleBindingId)) || emptyObj,
+    [clusterRoleBindings, clusterRoleBindingId])
+  const { params, updateParams, getParamsUpdater } = useParams(defaultParams)
+
+  const [updateClusterRoleBindingAction] = useDataUpdater(clusterRoleBindingActions.update, onComplete)
+  const handleSubmit = data => updateClusterRoleBindingAction({ ...data, ...params })
+  useEffect(() => {
```
xantorres: Please don't use `params`, `ValidatedForm` already passes the form values to `handleSubmit` upon submission of the form.

src/app/plugins/kubernetes/components/common/GroupMultiSelect.js
```diff
@@ -1,21 +1,24 @@
...
 
   const handleValuesChange = values => {
     setValues(values)
     onChange && onChange(values)
   }
 
   const groupsList = projectAs({ label: 'name', value: 'name' }, groups)
+  useEffect(() => {
+    onChange && onChange(values)
+  }, [initialValue])
```
brennancheung: Did you mean `values`?  `initialValue` by definition should never change.

PR #448, opened 2019-11-10
[PMK-2186] Created smaller alert component for login failure alert

kevinylee1991:
Debating whether or not to keep a separate component for smaller alerts? For this first pass, copied the entire Alert component and changed the styling. Could possibly add a prop that determines the size, but it could be possible these large and small alerts will diverge more in the future. Thoughts?

Screen Shot 2019-11-11 at 1 53 09 AM

### DIFF
--- a/src/app/core/components/Alert.js
+++ b/src/app/core/components/Alert.js
@@ -27,6 +27,14 @@ const useStyles = makeStyles(theme => ({
     width: '100%',
     border: 0,
   },
+  rootSmall: {
+    maxWidth: 800,
+    padding: theme.spacing(1, 1),
+    margin: theme.spacing(2, 0),
+    width: '100%',
+    border: 0,
+    display: 'flex',
+  },
   success: { color: green[600] },
   error: { color: theme.palette.error.dark },
   info: { color: theme.palette.primary.dark },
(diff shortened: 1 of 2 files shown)


src/app/plugins/openstack/components/LoginPage.js
```diff
@@ -150,7 +150,7 @@ export class LoginPage extends React.PureComponent {
         Service</SimpleLink>.
       </Typography>
       <Typography className={classes.paragraph} variant="caption" color="textSecondary">
-        © 2014-2018 Platform9 Systems, Inc.
+        © 2014-2019 Platform9 Systems, Inc.
```
xantorres: we can make this future proof by doing:
```
© 2014-{moment().format('YYYY')} Platform9 Systems, Inc.
```

src/app/core/components/AlertSmall.js
```diff
@@ -0,0 +1,89 @@
...
+    opacity: 0.9,
+    marginRight: theme.spacing(1),
+  },
+  message: {
+    alignSelf: 'center',
+    flexGrow: 1
+  },
+}))
+
+const AlertSmall = ({ children, message, variant }) => {
```
xantorres: I'm not a fan of this approach, It would be better to refactor our already existing `Alert` and perhaps add a `small` property which then we could reuse thorough the entire codebase

brennancheung: Agreed.  We should favor re-using components as much as possible.

src/app/core/components/Alert.js
```diff
@@ -27,6 +27,14 @@ const useStyles = makeStyles(theme => ({
     width: '100%',
     border: 0,
   },
+  rootSmall: {
```
xantorres: I suggest you using style props, so that you don't need to do change the classes depending on the prop, Eg:
```
root: {
	  maxWidth: 800,
    position: 'relative',	
    padding:  ({ small }) => small ? theme.spacing(1, 1) : theme.spacing(1, 8),
    ....
}
```

src/app/core/components/Alert.js
```diff
@@ -52,24 +71,28 @@ const useStyles = makeStyles(theme => ({
...
   },
+  messageSmall: {
+    alignSelf: 'center',
+    flexGrow: 1
+  },
 }))
 
-const Alert = ({ children, message, variant }) => {
+const Alert = ({ children, message, variant, small }) => {
   const classes = useStyles()
```
xantorres: here you can pass the `const classes = useStyles({ small })` prop to use it within the styles.

PR #451, opened 2019-11-11
Auto select first region if stored region doesn't exist

xantorres:
This fixes the confusing issue that happens when switching between servers, which tried to use the previously selected Region even if that region no longer existed.
However, it is not yet a perfect solution, as it is now, errors are still shown if a region is chosen without some required services (such as `qbert`)
I will address this other issue in another PR, after further discussing the possible solutions.

### DIFF
--- a/src/api-client/Keystone.js
+++ b/src/api-client/Keystone.js
@@ -1,5 +1,5 @@
 import axios from 'axios'
-import { pluck } from 'ramda'
+import { pluck, pipe, values, head } from 'ramda'
 import { getHighestRole } from './helpers'
 import { pathJoin } from 'utils/misc'
 
@@ -111,7 +111,7 @@ class Keystone {
       ...this.client.getAuthHeaders(),
       params: {
         'scope.project.id': tenantId,
-        'include_names': true
+        'include_names': true,
       },
     })
     return response.data.role_assignments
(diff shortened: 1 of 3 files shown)


src/api-client/Keystone.js
```diff
@@ -281,16 +281,15 @@ class Keystone {
   }
 
   getServicesForActiveRegion = async () => {
-    if (!this.client.serviceCatalog) {
-      await this.getServiceCatalog()
-    }
+    const { serviceCatalog = await this.getServiceCatalog(), activeRegion } = this.client
```
brennancheung: Nice way to clean it up.  Might be cleaner with `activeRegion` first to make the separation cleaner.  Or just use multiple lines.

src/app/plugins/openstack/components/regions/RegionChooser.js
```diff
@@ -1,38 +1,39 @@
...
+  const [selectedRegion, setRegion] = useState()
+  const curRegion = useMemo(() => {
+    if (selectedRegion) {
+      return selectedRegion
+    }
+    if (lastRegion && find(propEq('id', lastRegion.id), regions)) {
+      return lastRegion.id
+    }
+    return pipe(head, prop('id'))(regions) || 'Current Region'
+    // return 'Current Region'
```
brennancheung: This line can be removed

PR #452, opened 2019-11-11
[PMK-2185] Fixed login page links

kevinylee1991:


src/app/plugins/openstack/components/LoginPage.js
```diff
@@ -122,7 +122,8 @@ export class LoginPage extends React.PureComponent {
         label={
           <div>
             <span>I have a Multi-Factor Authentication (MFA) token. (</span>
-            <SimpleLink src="http://www.platform9.com">more info</SimpleLink>
+            <SimpleLink src='https://platform9.com/support/setup-multi-factor-authentication-with-platform9/'
```
brennancheung: We have `ExternalLink` which removes the boilerplate of the `target='_blank'`.  Add the `rel` to it if it doesn't exist.

PR #453, opened 2019-11-12
Onboarding

jruddell:
### WIP:
- Need to figure out where to pull / listen for data.
- Need updated details about where to install the CLI.
- Need to figure out how to deep link into a wizard flow


src/app/plugins/kubernetes/components/onboarding/DownloadCliWizard.tsx
```diff
@@ -0,0 +1,101 @@
...
+    () => [
+      'Open the Terminal App',
+      <div className={classes.flex}>
+        <Typography>Type</Typography>
+        <Code>
+          <span>curl -# -o ftp://ftp.example.com/file.zip</span>
+        </Code>
+      </div>,
+    ],
+    [perminentKey],
```
xantorres: This dependency here will never be updated,  moreover you are not using it inside of the `useMemo` so you can just get rid of it.

src/app/plugins/kubernetes/components/onboarding/DownloadCliWizard.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+const perminentKey = 'never_update'
+
+const DownloadCliWizard = () => {
+  const classes = useStyles({})
+
+  const {params} = useParams()
+
+  const handleSubmit = parameters => async data => {
+    console.log(parameters, data)
```
xantorres: `params` is available in the scope, so there's no point of passing it as a parameter here.

src/app/plugins/kubernetes/components/onboarding/DownloadCliWizard.tsx
```diff
@@ -0,0 +1,101 @@
...
+    alignItems: 'center'
+  }
+}))
+
+const perminentKey = 'never_update'
+
+const DownloadCliWizard = () => {
+  const classes = useStyles({})
+
+  const {params} = useParams()
```
xantorres: `params` is never updated? I guess this is a WIP?

src/app/plugins/kubernetes/components/onboarding/DownloadCliWizard.tsx
```diff
@@ -0,0 +1,101 @@
...
+import BulletList from 'core/components/BulletList'
+import Code from 'core/components/CodeBlock'
+
+const initialContext = {
+
+}
+
+const useStyles = makeStyles<any, {spacing?: number}>((theme: Theme) => ({
+  detailSection: {
+    marginTop: ({spacing}) => theme.spacing(spacing)
```
xantorres: you allow passing a `spacing` prop but I don't see this  being used anywhere.

src/app/core/components/FormWrapper.js
```diff
@@ -27,7 +27,7 @@ const FormWrapper = ({
   backUrl = undefined,
   children,
   title,
-  className,
+  className = undefined,
```
xantorres: why? Everything is `undefined` by default already.

src/app/plugins/kubernetes/components/onboarding/OnboardingPage.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const OnboardingPage = () => {
+  const classes = useStyles({})
+  const [activePlatform, setActivePlatform] = useState<SupportedPlatforms>(SupportedPlatforms.Other)
+  const [activeView, setActiveView] = useState<null | 'other'>(null)
+  const handleClick = useCallback(
+    (type: SupportedPlatforms) => {
+      setActivePlatform(type)
+    },
+    [activePlatform],
```
xantorres: You can probably get rid of this dependency here. Just `[]` should be fine

src/app/plugins/kubernetes/components/onboarding/DownloadCliWizard.tsx
jruddell: @xantorres  If I dont want to re-create this list of items, but need to define in the render because it uses styling. What is the more appropriate way of doing that?

src/app/plugins/kubernetes/components/onboarding/DownloadCliWizard.tsx
jruddell: yes still WIP, although I'm not positive I even need params yet

src/app/plugins/kubernetes/components/onboarding/DownloadCliWizard.tsx
jruddell: It is being used in the `WizardStepDetails`

src/app/core/components/FormWrapper.js
jruddell: Typescript infers from javascript. If there is no typing that specifies an optional param, then giving a default inline assignment value helps it infer optional parameters

src/app/plugins/kubernetes/components/onboarding/DownloadCliWizard.tsx
xantorres: Just declare it outside of the render function

jruddell: I can't declare it outside of the render body because the elements use one of the classes from `useStyles`. Does `useMemo` with an empty array of dependencies provide value? meaning if there is no list of dependencies then doesn't it just update every render regardless? feels like that would be more overhead

xantorres: I know you need the clasname, that's why you can create a memoized function that receives the clasname as param. But `useMemo` also is good enough

xantorres: Another option is to create a small component, so you can use `useStyles` there

src/app/plugins/kubernetes/components/dashboard/DashboardPage.js
```diff
@@ -43,6 +43,7 @@ const useStyles = makeStyles(theme => ({
 
 const serviceReports = [
   {
+    key: 'pods-service',
```
xantorres: why not just use the `route` as the key? They are unique anyway

PR #455, opened 2019-11-13
Fix logging view

smanpathak:
This change fixes some issues with logging outputs table view.
The logic maps multiple output objects to a single cluster and displays by cluster in case of multiple clusters.
However, following cases needed addressing
* The logging feature may not be enabled on a cluster (don't show)
* Logging feature is enabled, but no outputs are created (don't show)
Also fixed the field so it shows cluster name and not uuid.


src/app/plugins/kubernetes/components/logging/actions.js
```diff
@@ -24,10 +24,10 @@ const mapLoggings = (cluster, loggings) => {
   })
 
   return {
-    cluster: cluster.uuid,
+    cluster: cluster.name,
```
brennancheung: I think you will need to create a new field for the name.  The caching layer requires this `uuid` to scope by cluster.  Removing it will effectively cause each cluster to overwrite each other cluster's logging data.

smanpathak: Thanks, will do

src/app/plugins/kubernetes/components/logging/actions.js
```diff
@@ -24,10 +24,11 @@ const mapLoggings = (cluster, loggings) => {
   })
 
   return {
-    cluster: cluster.uuid,
+    cluster: cluster.id,
+    clusterName: cluster.name,
     status: cluster.status,
-    logStorage,
-    logDestination,
+    logStorage: logStorage,
```
xantorres: In Javascript you don't need to specify the variable name twice if the object prop matches the variable, so this is redundant

src/app/plugins/kubernetes/components/logging/actions.js
```diff
@@ -40,7 +41,7 @@ const loggingActions = createCRUDActions(loggingsCacheKey, {
       return response ? mapLoggings(cluster, response) : null
     }, clusters)
 
-    return loggings.filter(cluster => cluster !== null)
+    return loggings.filter((logStorage) => logStorage != null)
```
xantorres: Parenthesis are not required when there's only on parameter

smanpathak: Done

PR #456, opened 2019-11-13
Kubeconfig backend

murilovmachado:
This PR Implements download kubeconfig Yaml file.


src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
```diff
@@ -33,9 +41,14 @@ const KubeConfigListPage = () => {
     compactTable: true,
     blankFirstColumn: true,
     multiSelection: false,
-    onSelect: toggleDialog,
+    onSelect,
   }), [toggleDialog])
 
+  const downloadKubeconfig = async (clusterId) => {
+    const kubeconfig = await getKubeconfig({ clusterId })
+    console.log({ kubeconfig })
```
murilovmachado: Why this returns undefined? @xantorres @brennancheung 
In the `actions` file, it returns the correct yaml string, not sure why I'm getting undefined here.

xantorres: This is not how `customOperations` work, these operations are meant to be used to perform some kind of update and they should return the update (full) cache array. If you want to retrieve something from the server api, either you use a contextLoader (or `listFn` when using `createCRUDActions`) or you just skip using a context loader (which is intended to cache stuff) and do the api call directly.

src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
```diff
@@ -20,9 +23,48 @@ const useStyles = makeStyles(theme => ({
 
 const KubeConfigListPage = () => {
   const [isDialogOpen, toggleDialog] = useToggler()
+  const [selectedCluster, setSelectedCluster] = useState()
+  const [downloadedKubeconfigs, setDownloadedKubeconfigs] = useState({})
+  const [currentKubeconfig, setCurrentKubeconfig] = useState('')
+  const [generateYaml, setGenerateYaml] = useState(false)
```
brennancheung: Maybe collapse these into a single state.  It's a tradeoff either way so your call.

src/app/plugins/kubernetes/components/apiAccess/kubeConfig/DownloadDialog.js
```diff
@@ -22,7 +22,7 @@ const useStyles = makeStyles(theme => ({
   },
 }))
 
-const DownloadDialog = ({ onClose, isDialogOpen }) => {
+const DownloadDialog = ({ onDownload, onClose, isDialogOpen }) => {
```
jruddell: small naming thing, `onDownload` reads as the callback after a download is complete. However this is more about triggering a download if I understand correctly. If thats the case probably best to rename this callback to something more like `onSubmit`

src/app/plugins/kubernetes/components/apiAccess/kubeConfig/KubeConfigListPage.js
```diff
@@ -51,24 +112,28 @@ const KubeConfigListPage = () => {
...
+        onDownload={() => downloadKubeconfig(selectedCluster, generateYaml)}
+        onClose={handleCloseDialog}
+        isDialogOpen={isDialogOpen}
+      />
     </>
   )
 }
 
-const getColumns = (handleClickDownload) => [
+const getColumns = (clickDownloadYamlFile) => [
```
jruddell: another naming thing. This method is for handling a download action / click. We should probably follow the convention onMethod -> handleMethod. reading `clickDownloadYamlFile` doesn't immediately pop out as a click handler callback compared to `handleDownloadYamlFile`

PR #460, opened 2019-11-13
Sending session cookie with keystone token using token2cookie endpoint

xantorres:
Also fixed some warnings and routing issues

### DIFF
--- a/src/api-client/Keystone.js
+++ b/src/api-client/Keystone.js
@@ -290,7 +290,6 @@ class Keystone {
     if (!activeRegion || !servicesByRegion.hasOwnProperty(activeRegion)) {
       // Just assume the first region we come across if there isn't one set.
       return pipe(values, head)(servicesByRegion)
-      // return null
     }
     return servicesByRegion[activeRegion]
   }
--- a/src/api-client/Keystone.js
+++ b/src/api-client/Keystone.js
@@ -1,5 +1,5 @@
 import axios from 'axios'
-import { pluck, pipe, values, head } from 'ramda'
+import { pluck, pipe, values, head, path } from 'ramda'
 import { getHighestRole } from './helpers'
 import { pathJoin } from 'utils/misc'
 
@@ -212,6 +212,20 @@ class Keystone {
       // authentication failed
       console.error(err)
       return null
+    }
+  }
+
+  resetCookie = async () => {
+    const services = await this.getServicesForActiveRegion()
+    const linksUrl = path(['regioninfo', 'public', 'url'], services)
+
+    // set cookie for accessing hostagent rpms
+    try {
+      const { data: { links = {} } } = await axios.get(linksUrl)
+      const token2cookieUrl = links.token2cookie
+      await axios.get(token2cookieUrl)
+    } catch (err) {
+      console.warn('Setting session cookie for accessing hostagent rpms failed')
     }
   }
 
--- a/src/app/App.js
+++ b/src/app/App.js
@@ -5,7 +5,6 @@ import AppContainer from 'core/components/AppContainer'
 import HotKeysProvider from 'core/providers/HotKeysProvider'
 import PreferencesProvider from 'core/providers/PreferencesProvider'
 import { ToastProvider } from 'core/providers/ToastProvider'
-import DeveloperToolsEmbed from 'developer/components/DeveloperToolsEmbed'
 import LogoutPage from 'openstack/components/LogoutPage'
 import SessionManager from 'openstack/components/SessionManager'
 import ThemeManager from './ThemeManager'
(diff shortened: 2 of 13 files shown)


src/app/App.js
```diff
@@ -17,46 +16,33 @@ import plugins from 'app/plugins'
...
-        <Route path={pathJoin(plugin.basePath, 'reset_password')} exact component={ResetPasswordPage} />
-        <Route path={pathJoin(plugin.basePath, 'forgot_password')} exact component={ForgotPasswordPage} />
-        <Route path={pathJoin(plugin.basePath, 'logout')} exact component={LogoutPage} />
-        {defaultRoute && <Redirect to={defaultRoute || '/ui/404'} />}
-        {showFooter && renderFooter()}
-      </Switch>
-    }} />
+  const genericRoutes = [
+    // TODO implement generic login page?
+    { link: { path: pathJoin(plugin.basePath, 'login') }, component: null },
```
brennancheung: Is `pathJoin` doing something other than just adding the `/`?  Seems like it might be easier just to use string interpolation.

src/api-client/Keystone.js
```diff
@@ -215,6 +215,20 @@ class Keystone {
...
   }
 
+  resetCookie = async () => {
+    const services = await this.getServicesForActiveRegion()
+    const linksUrl = path(['regioninfo', 'public', 'url'], services)
+
+    // set cookie for accessing hostagent rpms
+    try {
+      const { data: { links = {} } } = await axios.get(linksUrl)
+      const token2cookieUrl = links.token2cookie
```
jruddell: Defaulting `links` to an empty object looks like it can be undefined. if thats the case then we should probably check if the `token2cookie` param is there before doing the next request.

src/app/App.js
```diff
@@ -17,46 +16,33 @@ import plugins from 'app/plugins'
...
+    { link: { path: pathJoin(plugin.basePath, 'reset_password') }, exact: true, component: ResetPasswordPage },
+    { link: { path: pathJoin(plugin.basePath, 'forgot_password') }, exact: true, component: ForgotPasswordPage },
+    { link: { path: pathJoin(plugin.basePath, 'logout') }, exact: true, component: LogoutPage },
+    {
+      link: { path: pathJoin(plugin.basePath, '') },
+      // TODO: Implement 404 page
+      render: () => <Redirect to={defaultRoute || '/ui/404'} />,
+    },
+  ]
+  return [...plugin.getRoutes(), ...genericRoutes].map(route => {
```
jruddell: Why are we defining multiple login, reset_password, logout... etc routes? It looks like were creating these for each set of plugin routes. Shouldn't this just be a one time operation / definition?

src/app/App.js
xantorres: Is a bit cleaner and shorter than using interpolation, specially with multiple variables, in this case is not much of a difference

src/app/plugins/openstack/components/LogoutPage.js
```diff
@@ -1,26 +1,34 @@
...
+    await setContext(pipe(
+      assoc(dataCacheKey, emptyArr),
+      assoc(paramsCacheKey, emptyArr),
+    ))
     history.push(loginUrl)
-  }
+  }, [])
 
-  render = () => null
+  return <div />
```
jruddell: Do we need to return a empty div here or is null ok?

src/app/plugins/openstack/components/regions/RegionChooser.js
```diff
@@ -21,28 +31,28 @@ const RegionChooser = props => {
...
     await updatePrefs({ lastRegion })
+    setActiveRegion(region)
+    await keystone.resetCookie()
+    invalidateLoadersCache()
+
+    // Changing the Region will cause all the current active `useDataLoader`
+    // hooks to reload its data
+    setContext(assoc('currentRegion', region))
+
+    // Redirect to the root of the current section (there's no need to reload all the app)
```
jruddell: 👍

src/app/core/helpers/createContextLoader.js
```diff
@@ -134,7 +137,7 @@ const createContextLoader = (cacheKey, dataFetchFn, options = {}) => {
    * @returns {Promise<array>} Fetched or cached items
    */
   const contextLoaderFn = memoizePromise(
-    async ({ getContext, setContext, params = emptyObj, refetch = false, dumpCache = false, additionalOptions = emptyObj }) => {
+    async ({ getContext, setContext, params = emptyObj, refetch = contextLoaderFn._invalidatedCache, dumpCache = false, additionalOptions = emptyObj }) => {
```
jruddell: looking at the code here more, probably better to rename `_invalidatedCache` to something like `shouldInvalidateCache` so it reads better. Not a huge deal though

src/app/App.js
xantorres: These routes were already existing, I just refactored them so that we could use a single `<Switch>` component for all the plugin, as it was before there were problems when redirecting to the default route because `<Redirect>` was not working inside of the nested switch statements.
Regarding whether we should or not have shared routes for the common auth operations, it makes sense to me, the only problem I see is that there will be no way to identify the plugin the user was in if he decides to logout or change password and then come back and login again. 
We can probably figure this out later though. It's out of the scope of this PR

src/app/core/helpers/createContextLoader.js
xantorres: The initial underscore is a way of marking the prop as "private", not something to be used outside of the `createContextLoader`, about naming it differently I'm ok with it, I will rename it in another PR.

src/app/plugins/openstack/components/LogoutPage.js
xantorres: I return a div so that eslint doesn't complain when importing `React` as it doesn't detect jsx code. It doesn't matter anyway, we do a redirect to the login page when the component mounts.

src/api-client/Keystone.js
xantorres: This was just about obout preventing a fatal reference error when calling `token2cookie`. I'm a bit paranoid with that, but in this case I refactored the code later to use a `try` `catch` statement so it doesn't matter now. In any way it doesn't bother me much, but we can get rid of it as it is useless right now.

jruddell: Yea, I more meant we shouldn't be calling the api call next line if its undefined. but yea its kinda a useless point anyways

src/app/core/helpers/createContextLoader.js
jruddell: yea but if you really want private then you should just use a symbol as a key. I generally dislike underscores to mean private because a lot of people dont adhere to that nuance.

xantorres: I like the idea of using a symbol, will take a look later.

src/app/App.js
jruddell: Yea, I was noticing the extra routes, so wanted to bring it up. Do we need to know where they are logging out of? and wouldn't we already know that since we know where they currently are when clicking logout?

PR #467, opened 2019-11-14
PMK-2184: Added "Network Backend" options for CNI.

rahulkurumkar:
Added "Network Backend" options for CNI.

AddedNetworkBackend

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
@@ -18,7 +18,7 @@ import { clusterActions } from '../actions'
 import { pick } from 'ramda'
 import { pathJoin } from 'utils/misc'
 import { k8sPrefix } from 'app/constants'
-import { RadioGroup, Radio, FormControlLabel, FormLabel, FormControl } from '@material-ui/core';
+import { RadioGroup, Radio, FormControlLabel, FormLabel } from '@material-ui/core'
 
 const { qbert } = ApiClient.getInstance()
 
@@ -38,7 +38,7 @@ const runtimeConfigOptions = [
 ]
 
 const AddBareOsClusterPage = () => {
-  const { params, getParamsUpdater, updateParams } = useParams({networkPlugin: 'flannel'})
+  const { params, getParamsUpdater, updateParams } = useParams({ networkPlugin: 'flannel' })
   const { history } = useReactRouter()
   const onComplete = () => {
     history.push('/ui/kubernetes/infrastructure#clusters')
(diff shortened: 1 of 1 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
```diff
@@ -217,6 +217,24 @@ const AddBareOsClusterPage = () => {
...
+                        />
+                        <FormControlLabel
+                          value="calico"
+                          control={<Radio color="primary" />}
+                          label="Calico"
+                        />
+                        <FormControlLabel
+                          value="canal"
+                          control={<Radio color="primary" />}
+                          label="Calico with Flannel Newtworking"
```
kevinylee1991: typo "Networking"

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
```diff
@@ -59,7 +60,7 @@ const AddBareOsClusterPage = () => {
       ...pick('privileged appCatalogEnabled customAmi tags'.split(' '), data),
     }
     if (data.httpProxy) { body.httpProxy = data.httpProxy }
-    if (data.networkPlugin === 'calico') { body.mtuSize = data.mtuSize }
+    if (params.networkPlugin === 'calico') { body.mtuSize = data.mtuSize }
```
brennancheung: Do we have the same bug on the other cluster types?  If so, can you make this change there as well.  John refactored the common functionality between clusters into 1 spot so you might have to do this if you pull the latest anyway.

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
```diff
@@ -217,6 +217,24 @@ const AddBareOsClusterPage = () => {
                         label="HTTP Proxy"
                         info="Specify the HTTP proxy for this cluster.  Leave blank for none.  Uses format of <scheme>://<username>:<password>@<host>:<port> where <username>:<password>@ is optional."
                       />
+                      <FormLabel component="legend">Network Backend</FormLabel>
```
brennancheung: We're actually using a Picklist in the other cluster types.  We should probably make it consistent.
You can see how it is implemented in AWS here:
https://github.com/platform9/pf9-ui-plugin/blob/master/src/app/plugins/kubernetes/components/infrastructure/clusters/AddAwsClusterPage.js#L493

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
brennancheung: Why is this a param instead of just an ordinary form field.  You can copy the AWS implementation that has the same logic.

PR #468, opened 2019-11-14
[PMK-2178] Added region/tenant label to region and tenant choosers

kevinylee1991:
Screen Shot 2019-11-15 at 12 10 36 AM


kevinylee1991:
btw that blue thing jutting out is an add button -- maybe we need a border on our header lol

PR #476, opened 2019-11-15
PMK-2184: Added "Network Backend" options for CNI.

rahulkurumkar:


src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
```diff
@@ -217,6 +230,22 @@ const AddBareOsClusterPage = () => {
...
+                        info=""
+                        onChange={handleNetworkPluginChange({ setWizardContext, setFieldValue })}
+                        required
+                      />
+                      {values.networkPlugin === 'calico' &&
+                      <TextField
+                        id="mtuSize"
+                        label="MTU Size"
+                        info="Maximum Transmission Unit (MTU) for the interface (in bytes)"
+                        required={values.networkPlugin === 'calico'}
```
xantorres: this is unnecessary, as the TextField will not even be rendered if `networkPlugin` is not "calico", so just `required` is enough

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
```diff
@@ -36,6 +37,19 @@ const runtimeConfigOptions = [
...
 ]
 
+const networkPluginOptions = [
+  { label: 'Flannel', value: 'flannel' },
+  { label: 'Calico', value: 'calico' },
+  { label: 'Canal (experimental)', value: 'canal' },
+]
+
+const handleNetworkPluginChange = ({ setWizardContext, setFieldValue }) => option => {
+  if (['calico', 'canal', 'weave'].includes(option)) {
```
xantorres: this condition will always be met, as far as I know, since there are no other options available to select, so we can get rid of this

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
```diff
@@ -36,6 +37,19 @@ const runtimeConfigOptions = [
...
+    setWizardContext({ privileged: true, custom: option })
```
xantorres: I don't see the need of updating the `wizardContext`

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
```diff
@@ -36,6 +37,19 @@ const runtimeConfigOptions = [
...
+    setFieldValue('privileged')(true)
```
xantorres: This whole logic is a bit confusing and this whole `handleNetworkPluginChange` is probably unnecessary, just check if any networkPlugin has been selected upon submission of the form, and if it's so, just assign the `privileged` variable manually.

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
rahulkurumkar: We have `flannel` as other option to be selected in that case it won't be disabled checked privileged

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
rahulkurumkar: you are right, will just remove it

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
rahulkurumkar: Remove the change hander

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
rahulkurumkar: Remove the handler

PR #479, opened 2019-11-15
Back to origin from the first Wizard step

jruddell:
Added a back option to each "Add Cluster" flow after selecting the platform. Fixes PMK-2213

image

image

image

### DIFF
--- a/src/app/core/components/wizard/Wizard.js
+++ b/src/app/core/components/wizard/Wizard.js
@@ -1,4 +1,5 @@
 import React, { PureComponent } from 'react'
+import { withRouter } from 'react-router-dom'
 import PropTypes from 'prop-types'
 import WizardButtons from 'core/components/wizard/WizardButtons'
 import NextButton from 'core/components/buttons/NextButton'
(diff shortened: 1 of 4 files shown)


src/app/core/components/wizard/Wizard.js
```diff
@@ -35,6 +37,10 @@ class Wizard extends PureComponent {
   }
 
   handleBack = () => {
+    if (this.canBackAtFirstStep()) {
```
brennancheung: Separate going back and going to the origin as 2 distinct methods.

src/app/core/components/wizard/Wizard.js
```diff
@@ -92,7 +98,7 @@ class Wizard extends PureComponent {
         {renderStepsContent({ wizardContext, setWizardContext, onNext: this.onNext })}
         <WizardButtons>
           {onCancel && <CancelButton onClick={onCancel} />}
-          {this.hasBack() &&
+          {(this.hasBack() || this.canBackAtFirstStep()) &&
```
brennancheung: Separate prev step and going to origin as 2 different lines / elements.

PR #484, opened 2019-11-18
Implement username/password auth method

murilovmachado:
This PR implements username/password auth method for downloading a kubeconfig.

Screen Shot 2019-11-18 at 19 03 59

### DIFF
--- a/src/app/plugins/kubernetes/components/apiAccess/kubeConfig/DownloadDialog.js
+++ b/src/app/plugins/kubernetes/components/apiAccess/kubeConfig/DownloadDialog.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react'
+import React, { useState } from 'react'
 import { makeStyles } from '@material-ui/core/styles'
 import ValidatedForm from 'core/components/validatedForm/ValidatedForm'
 import TextField from 'core/components/validatedForm/TextField'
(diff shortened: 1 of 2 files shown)


src/app/plugins/kubernetes/components/apiAccess/kubeConfig/DownloadDialog.js
```diff
@@ -25,74 +28,105 @@ const useStyles = makeStyles(theme => ({
...
+    if (authMethod === 'token') {
+      setNote('Token authentication is the preferred method for downloading kubeconfig. The kubeconfig will remain valid for the next 24 hours.')
+      setConfirmLabel('Download Config')
+      setType('button')
+    } else {
+      setNote('Password authentication is less secure than token authentication, but the kubeconfig will remain functional for as long as the username and password are valid.')
+      setConfirmLabel('Validate Credentials')
+      setType('submit')
+    }
+  }, [authMethod])
```
brennancheung: This state is unnecessary.  All these values are 1-to-1 mapping with `authMethod`.  You can conditionally render this based on the `authMethod`.

Worse case scenario you can use `const note = authMethod === 'token' ? 'abc' : 'xyz'`.

But I would favor something like:

`{authMethod === 'token' && <div>...</div>}`

jruddell: Be careful rendering with `&&`. sometimes you'll get unexpected elements rendered in the browser. `{ cond && <div>...</div> }` when this evaluates to false I've seen a `0` get rendered in the browser to represent `false`, since that is what it evaluates to. Best to render null

murilovmachado: we can use `!!cond && <div>...</div>` to avoid a `0` value being rendered. I'll make the fix.

PR #485, opened 2019-11-19
node detail page, fixes PMK-2265

jruddell:
image

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/nodes/NodeDetailsPage.tsx
+++ b/src/app/plugins/kubernetes/components/infrastructure/nodes/NodeDetailsPage.tsx
@@ -56,8 +56,10 @@ const ClusterDetailsPage = () => {
         </>
       }
     >
-      {loading ? <Progress message="Loading Nodes..." /> : null}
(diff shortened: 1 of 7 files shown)


src/app/plugins/kubernetes/components/infrastructure/nodes/NodeDetailsPage.tsx
```diff
@@ -0,0 +1,118 @@
...
+  )
+}
+
+const NodeDetail = ({ uuid, name, combined, ...rest }: ICombinedNode) => {
+  const { flex, card } = useStyles({})
+  const hostId = uuid
+  const roles = combined?.roles
+  const hostname = name
+  const os = combined?.osInfo
+  const hypervisorType = combined?.resmgr?.hypervisor_info?.hypervisor_type
```
xantorres: ❤️

jruddell: upgraded our version of typescript so we can use it! no more helper functions for path lookups needed! 😍

PR #487, opened 2019-11-19
PMK-2261: Hide the spot instance node column and spot instance fields for a cluster

rahulkurumkar:
Hide the spot instance node column and spot instance fields for a cluster


rahulkurumkar:
Instead of 'display: false` I have added `excluded: true` so that it won't be available in column selector as well.

src/app/plugins/kubernetes/components/infrastructure/nodes/NodesListPage.js
```diff
@@ -61,7 +61,7 @@ export const columns = [
   { id: 'storage', label: 'Storage', render: renderStats('disk') },
   { id: 'clusterName', label: 'Cluster', render: renderClusterLink },
   { id: 'isMaster', label: 'Is Master?', render: isMaster },
-  { id: 'isSpotInstance', label: 'Spot Instance?', render: getSpotInstance },
+  { id: 'isSpotInstance', label: 'Spot Instance?', excluded: true, render: getSpotInstance },
```
brennancheung: Use `display: false` not `excluded`.  We still want them to be able to allow advanced users to add this column.

PR #490, opened 2019-11-20
PMK-2264: Added nodes information in cluster grid.

rahulkurumkar:


src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
```diff
@@ -113,6 +113,40 @@ const renderLinks = links => {
...
+  if (!nodes || !nodes.length) {
+    return <div>0</div>
+  }
+  const [expanded, setExpanded] = useState(false)
+  return (
+    <div>
+      {expanded ? (
+        <div>
+          {nodes.map(renderNodeLink)}
+          <SimpleLink onClick={() => setExpanded(!expanded)}>
```
brennancheung: `<SimpleLink>` is really more for actual navigation.  I think you can just use `<Typography>` with an `onClick`.  In fact, all text should ideally be going through `<Typography>` so that it works with the theming system.

rahulkurumkar: Fixed it

PR #493, opened 2019-11-20
Logging addon dialog button

smanpathak:
This change adds a button to toggle logging feature. It is pretty much the same as Prometheus addon feature enablement.


src/app/plugins/kubernetes/components/logging/LoggingAddonDialog.js
```diff
@@ -0,0 +1,60 @@
...
+
+export const hasLoggingEnabled = compose(castFuzzyBool, path(['tags', 'pf9-system:logging']))
+
+const { appbert } = ApiClient.getInstance()
+
+const LoggingAddonDialog = ({ rows: [cluster], onClose }) => {
+  const enabled = hasLoggingEnabled(cluster)
+  const toggleLogging = () => {
+    try {
+      appbert.getPackages().then(pkgs => {
```
jruddell: We should be using `useEffect` here so that if this component is unmounted during the request we prevent throwing JS exceptions.

Also update to use async/await

src/app/plugins/kubernetes/components/logging/LoggingAddonDialog.js
```diff
@@ -0,0 +1,60 @@
...
+        const logPkg = (pkgs || []).filter(pkg => (
+          pkg.name === 'pf9-log'
+        ))
+        if (logPkg === null || logPkg.length === 0) {
+          console.log('no logging package found')
+          onClose()
+          return
+        }
+        const logId = logPkg[0].ID
```
brennancheung: Is `ID` really all caps?  We should go with `id` if at all possible because already have so many permutations of id, I don't want to introduce yet more variants.

src/app/plugins/kubernetes/components/logging/LoggingAddonDialog.js
```diff
@@ -0,0 +1,60 @@
...
+
+const LoggingAddonDialog = ({ rows: [cluster], onClose }) => {
+  const enabled = hasLoggingEnabled(cluster)
+  const toggleLogging = () => {
+    try {
+      appbert.getPackages().then(pkgs => {
+        const logPkg = (pkgs || []).filter(pkg => (
+          pkg.name === 'pf9-log'
+        ))
+        if (logPkg === null || logPkg.length === 0) {
```
jruddell: `logPkg` wont be null because `filter` returns an array. You can update this to `if (!logPkg.length) {`

src/app/plugins/kubernetes/components/logging/LoggingAddonDialog.js
brennancheung: It's clearer and easier if you use `async` / `await`

src/app/plugins/kubernetes/components/logging/LoggingAddonDialog.js
```diff
@@ -0,0 +1,60 @@
...
+export const hasLoggingEnabled = compose(castFuzzyBool, path(['tags', 'pf9-system:logging']))
+
+const { appbert } = ApiClient.getInstance()
+
+const LoggingAddonDialog = ({ rows: [cluster], onClose }) => {
+  const enabled = hasLoggingEnabled(cluster)
+  const toggleLogging = () => {
+    try {
+      appbert.getPackages().then(pkgs => {
+        const logPkg = (pkgs || []).filter(pkg => (
```
jruddell: Considering the usage here. Update `filter` to `find`. Array.find will return the element you are looking for rather than filtering and pulling the first item from the array. 

`const logPkg = (pkgs || []).find( pkg => pkg.name === 'pf9-log' )`

then on line 25 you can update to just `const logId = logPkg.id`

src/app/plugins/kubernetes/components/logging/LoggingAddonDialog.js
xantorres: You could also have set `const logId = pathStr('0.id', logPkg)` before the condition so here you would just check `if (!logId)`

src/app/plugins/kubernetes/components/logging/LoggingAddonDialog.js
xantorres: I think using `useCallback` is enough, but I would probably have a `loading` state variable to lock the UI while the operation is performed, so we prevent the component to be unmounted.
You can wrap your dialog in a `<Progress loading={loading}` to do so.

src/app/plugins/kubernetes/components/logging/LoggingAddonDialog.js
smanpathak: it's  in caps unfortunately, will have to fix appbert responses.

src/app/plugins/kubernetes/components/logging/LoggingAddonDialog.js
smanpathak: I replaced it with a find

PR #499, opened 2019-11-22
form validation and data gathering for nodes

jruddell:
WIP. Haven't worked on the submission yet.


src/app/core/components/Loading.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return <div className={flex}>
+    <div className={ `${rotateContainer} ${loading ? rotateCls : ''}` }>
+      <IconComponent {...iconProps} />
+    </div>
+    <span className={spacer} />
+    {children}  
+  </div>
+}
+
+export default Loading
```
xantorres: I think this is going to be confusing, we already have a `Progress` that we use for most of our loading operations, and with now a `Loading`, will be confusing to know which one to use (specially for newcomers).
Could we perhaps refactor Progress to have a prop to show the animated icon instead of having a new component?
Either that or just rename the component to a more specific name, such as `ClusterLoading` or something like this.

jruddell: I think `ClusterLoading` is probably fine, this component renders a spinning icon and a child to the side, I am repurposing it here as well https://platform9.atlassian.net/wiki/spaces/PM/pages/642155376/Wireframes+Bare+OS+Cluster+Create where the table has a refresh icon. Its a different kind of loading / feedback, what should we call it?

xantorres: I can't see this link :(

brennancheung: We might be able to abstract out this type of logic into its own hook or component.  But this kind of thing is not going to be used very often so it's probably fine.  It will mostly only be used for adding nodes when creating clusters and initial onboarding flows.

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/ClusterHostChooser.js
```diff
@@ -1,63 +1,124 @@
...
+                  <TableCell>
+                    <Checkbox checked={allSelected()} onChange={toggleAll} />
+                  </TableCell>
+                  <TableCell>Hostname</TableCell>
+                  <TableCell>IP Address</TableCell>
+                  <TableCell>Operating System</TableCell>
+                </TableRow>
+              </TableHead>
+              <TableBody>
+                { !orphanedNodes.length
```
brennancheung: `length` should be comparing against an integer value not casting to a boolean.

Not a big fan of the ternary with nulls.  It takes up a lot of space and doesn't need to be there.  As long as the conditional evaluates to an actual boolean you don't need the null clause.

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -94,11 +94,60 @@ const createAzureCluster = async (data, loadFromContext) => {
...
+    )
+  )
+
+  if (data.enableMetallb) {
+    body.metallbCidr = data.metallbCidr
+  }
+
+  // 1. Get the nodePoolUuid from the nodePools API and look for the pool with name 'defaultPool'
+  const nodePools = await qbert.getNodePools()
+  data.nodePoolUuid = nodePools.find(x => x.name === 'defaultPool').uuid
```
brennancheung: This would be a good place to try out the new optional chaining / existential operator.

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -94,11 +94,60 @@ const createAzureCluster = async (data, loadFromContext) => {
...
+    'metallbCidr',
+    'externalDnsName',
+    'containersCidr',
+    'servicesCidr',
+    'mtuSize',
+    'privileged',
+    'appCatalogEnabled',
+    'tags',
+  ]
+  const body = Object.fromEntries(
```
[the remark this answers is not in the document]
jruddell: This is generating an object from the keys specified in the array.

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/ClusterHostChooser.js
jruddell: I actually always default to using a ternary and a null. Because of the many times I'd see an extra `0` in the browser or a `1` ... etc.

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
jruddell: In what way?

src/app/core/components/Loading.tsx
jruddell: @xantorres sorry heres a screenshot of it. The refresh icon and text. 

image

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/ClusterHostChooser.js
brennancheung: It's a lot of code to work around something that can easily be resolve be ensuring it is a boolean.

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
brennancheung: `nodePools?.find(...)?.uuid`

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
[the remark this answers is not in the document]
brennancheung: That's what `pick` does.  https://devdocs.io/ramda/index#pick

```javascript
const body = pick(keysToPluck, data)
```

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
jruddell: Ah if nodePools is null from the request? yea good point

jruddell: issue is this is in a js file

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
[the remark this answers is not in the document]
jruddell: 👍

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/ClusterHostChooser.js
jruddell: 👍

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
brennancheung: Ahh, bummer.

PR #500, opened 2019-11-22
PMK-2258: Added "Finish And Review" button for AWS and Azure cluster creation.

rahulkurumkar:


src/app/plugins/kubernetes/components/infrastructure/clusters/AddAwsClusterPage.js
```diff
@@ -44,6 +44,7 @@ const initialContext = {
   runtimeConfigOption: 'default',
   isPrivate: false,
   internalElb: false,
+  showFinishAndReview: true,
```
brennancheung: We definitely don't want to mix this into the `wizardContext` as an ordinary value.  We should use a predicate function we can pass as a prop into the Wizard component.

Something like:

```typescript
const canReview = (wizardContext): boolean => hasAllRequiredFields(ctx)
<Wizard showFinishAndReview={canReview} />
```

rahulkurumkar: Passing `canReview` props to `<Wizard />` as the validation part is already taken care.

src/app/core/components/wizard/Wizard.js
```diff
@@ -90,7 +110,7 @@ class Wizard extends PureComponent {
 
   render () {
     const { wizardContext, setWizardContext, steps, activeStep } = this.state
-    const { showSteps, children, submitLabel, onCancel } = this.props
+    const { showSteps, children, submitLabel, finishAndReviewLabel, onCancel, canReview } = this.props
```
xantorres: `canReview` is a bit strange (seems something about permissions), perhaps `showFinishButton` is more self explanatory?

src/app/core/components/wizard/Wizard.js
```diff
@@ -116,12 +137,14 @@ Wizard.propTypes = {
   onCancel: PropTypes.func,
   context: PropTypes.object,
   submitLabel: PropTypes.string,
+  finishAndReviewLabel: PropTypes.string,
```
xantorres: missing the new `canReview` (or whichever name you end up calling it) prop here

src/app/core/components/wizard/Wizard.js
```diff
@@ -71,6 +72,25 @@ class Wizard extends PureComponent {
...
+    const { activeStep } = this.state
+
+    if (this.nextCb[activeStep] && this.nextCb[activeStep]() === false) {
+      return
+    }
+
+    this.setState(
+      state => ({ activeStep: state.steps.length - 1 }),
+      () => {
+        this.activateStep()
```
xantorres: I'm not sure why we are calling `this.activateStep` and therefore doing another `setState` after setting the active step. It looks like something that could be achieved simply by updating the state once (and avoiding one extra useless re-render).
I know we do the same when navigating between contiguous  steps, but still, something that could be reviewed and improved.

src/app/core/components/wizard/Wizard.js
rahulkurumkar: Changed it to `showFindAndReviewButton`

src/app/core/components/wizard/Wizard.js
rahulkurumkar: Refactored the code to update activeStepId and activeStep in single go

src/app/core/components/wizard/Wizard.js
rahulkurumkar: Added

PR #507, opened 2019-11-22
[PMK-2201] Force refetching of nodes when refreshing list

xantorres:
It turns out `loadNodes` is grabbing all its data from cached resources using `loadFromContext`, so even when we try to refetch the data, the data is being pulled out from the cache because `loadFromContext` doesn't refetch by default, unless either `refetchCascade` option is defined, or manually specifying so when calling `loadFromContext(cacheKey, params, refetch)` with `refetch = true`


jruddell:
this is what I was referring to the other day with refetch not happening

PR #508, opened 2019-11-23
Typescript support for cluster model

brennancheung:
The `ICluster` interface was autogenerated from [Paste JSON as Code](https://marketplace.visualstudio.com/items?itemName=quicktype.quicktype) VS Code extension.

I modified it in a few spots where I knew it the data schema but some of it might still be a little off.  We can change it in the future as we notice the discrepancies or if we want to restrict it down further in the case of some of the enum string values.

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/model.ts
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/model.ts
@@ -47,7 +47,7 @@ export interface ICluster {
   nodePoolName: string
   cloudProviderUuid: string
   cloudProviderName: string
-  cloudProviderType: 'aws' | 'azure' | 'bareos'
+  cloudProviderType: CloudProviders
   cloudProperties: CloudProperties
   tags: Tags
   endpoint: string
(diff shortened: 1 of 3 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/model.ts
```diff
@@ -0,0 +1,118 @@
...
+  etcdVersion: string
+  apiserverStorageBackend: string
+  enableCAS: number
+  numMinWorkers: number
+  numMaxWorkers: number
+  nodePoolUuid: string
+  nodePoolName: string
+  cloudProviderUuid: string
+  cloudProviderName: string
+  cloudProviderType: 'aws' | 'azure' | 'bareos'
```
jruddell: we should make this an enum and use it everywhere so we dont have to type these out in multiple places (thinking about choose a cloud provider screen ... etc.)

```
enum CloudProviders {
  Aws = 'aws',
  Azure = 'azure',
  BareOS = 'bareos',
}
```

src/app/plugins/kubernetes/components/infrastructure/clusters/model.ts
```diff
@@ -0,0 +1,118 @@
...
+  cloudProperties: CloudProperties
+  tags: Tags
+  endpoint: string
+  kubeconfigUrl: string
+  isUpgrading: boolean
+  nodes: any[]
```
xantorres: Should this be typed as well? We should avoid using `any` whenever possible

brennancheung: Yeah, it's just auto-generated.  We can add to it over time.

PR #509, opened 2019-11-23
Add yarn and error handling improvements and fixes

xantorres:
* Added `yarn.lock`, at doing so I had to remove `package-lock.json` to avoid consistency issues. Please visit https://yarnpkg.com/en/docs/install for instructions on how to install yarn on your machine.
* Fixed an issue with some api calls error handling being shadowed by localized `try catch` sentences in their respective api calls, and therefore not being captured by the loader hooks and not showing a toast notification to the user.
* Added keystone `getServiceEndpoint(service, type)` method for simplified resolution of specified service endpoint and centralized error handling.
* Improve error feedback when selected region does not have available endpoint for a given service
image
* Fixed stylings on toast notifications when the text is too long and fixed the use cases to properly display the error when there are network issues, eg http 503 or 401 errors, as shown in the picture:
image
* Remove duplicated messages to prevent flooding the screen with redundant toast notifications
* Prevent some requests that were requested by unauthorized users, eg credentials and all tenants/users list.
* Added Typescript typings for ramda.
* Fix misc warnings and errors


src/api-client/ApiClient.js
```diff
@@ -11,7 +11,8 @@ import Nova from './Nova'
 import Qbert from './Qbert'
 import ResMgr from './ResMgr'
 import { normalizeResponse } from 'api-client/helpers'
-import { pathStrOr } from 'utils/fp'
+import { hasPathStr, pathStr } from 'utils/fp'
+import { prop, has, cond, T, identity } from 'ramda'
```
jruddell: what is `T`? can we alias it to be more specific?

src/app/core/helpers/createContextLoader.js
```diff
@@ -155,6 +155,10 @@ const createContextLoader = (cacheKey, dataFetchFn, options = {}) => {
       )(params)
       // If not all the required params are provided, skip this request and just return an empty array
       if (requiredParams && values(providedRequiredParams).length < allRequiredParams.length) {
+        // Show up a warning when trying to refetch the data without providing some of the required params
+        if (refetch && !contextLoaderFn._invalidatedCache) {
```
jruddell: if `_invalidatedCache` is supposed to be private, it might be better to make a `get isCacheInvalidated()` to return the private method

src/app/plugins/kubernetes/components/logging/actions.js
```diff
@@ -10,8 +11,7 @@ export const loggingsCacheKey = 'loggings'
 const mapLoggings = (cluster, loggings) => {
   const logStorage = []
   const logDestination = []
-
-  loggings.items.forEach(item => {
+  propOr([], 'items', loggings).forEach(item => {
```
jruddell: why not just inline the assignment? `const mapLoggings = (cluster, loggings = []) => {`

src/app/plugins/kubernetes/components/userManagement/tenants/actions.js
```diff
@@ -14,7 +14,12 @@ export const mngmTenantsCacheKey = 'managementTenants'
 const reservedTenantNames = ['admin', 'services', 'Default', 'heat']
 export const filterValidTenants = tenant => !reservedTenantNames.includes(tenant.name)
 export const mngmTenantActions = createCRUDActions(mngmTenantsCacheKey, {
-  listFn: async () => keystone.getAllTenantsAllUsers(),
+  listFn: (params, loadFromContext, getContext) => {
+    const { role } = getContext(prop('userDetails'))
+    return role === 'admin'
+      ? keystone.getAllTenantsAllUsers()
```
jruddell: I would probably be more in favor of putting admin checks further into the process. Let the api handler check the user role. would be really cool to make a decorator `@adminRole` that you can wrap each method like `getAllTenantsAllUsers`.

src/app/plugins/kubernetes/components/userManagement/users/actions.js
```diff
@@ -19,7 +19,12 @@ export const isSystemUser = ({ username }) => {
   return uuidRegex.test(username)
 }
 export const mngmCredentialsCacheKey = 'managementCredentials'
-createContextLoader(mngmCredentialsCacheKey, () => keystone.getCredentials())
+createContextLoader(mngmCredentialsCacheKey, (params, loadFromContext, getContext) => {
+  const { role } = getContext(prop('userDetails'))
+  return role === 'admin'
```
jruddell: Same thing here about role checking

src/api-client/ApiClient.js
xantorres: Is a function that always returns true, I would say is quite common when using ramda and I would prefer not to alias these functions as it could confuse people who are used to work with the library.

src/app/core/helpers/createContextLoader.js
xantorres: No need to create a getter, this property is just used internally, I don't want it to be exposed. What it could be done is to use a Symbol as you mentioned, but is out of the scope of this PR

src/app/plugins/kubernetes/components/logging/actions.js
xantorres: I don't know about that, this code is not mine, I just fixed an error that was causing a fatal error and white screen when opening this page, I didn't want to spend much time on refactoring this code.

src/app/plugins/kubernetes/components/userManagement/tenants/actions.js
xantorres: The api handler doesn't know the role of the user as far as I'm aware of and I didn't want to add state into the api class as it would be another thing to be worried about and to keep updated when changing the Region.

src/app/plugins/kubernetes/components/userManagement/users/actions.js
xantorres: I know is not the most ideal thing right now but it's going to do the trick until we come up with a note generic way of protecting api calls from unauthorised users. The thing is right now this is just used in literally 2 places so I don't think is worth it spending too much time on a more complex and generic implementation, unless we see the need of it in the future.

src/app/core/helpers/createContextLoader.js
jruddell: 👍

src/app/plugins/kubernetes/components/logging/actions.js
jruddell: 👍

src/app/plugins/kubernetes/components/userManagement/tenants/actions.js
jruddell: Hmm, I really want to avoid having to do this role check in every action. Like if I have another action that calls the same method I have to duplicate the role logic. Maybe the decorator itself can look into the context for the user role?

src/app/plugins/kubernetes/components/userManagement/users/actions.js
jruddell: Yea, were going to have to start doing this everywhere. We should discuss a good way to do it and then implement in the next few weeks when we have a little more time to do it :)

src/app/plugins/kubernetes/components/userManagement/tenants/actions.js
xantorres: Fixed https://github.com/platform9/pf9-ui-plugin/pull/512

PR #512, opened 2019-11-26
Add requiredRoles option in context loaders

xantorres:
image
If the user doesn't have the required roles an empty array will be returned. This is just a life insurance as loaders that require this roles should not even be called in the first place (the routes are protected)

Unfortunately in some cases other loaders are used that reuse other loaders via `loadFromContext` function, that was the case in the Dashboard page where the users are being loaded, and as a side effect the credentials and tenants are being loaded to perform some extra logic that is just required in the user management section.

### DIFF
--- a/src/app/plugins/kubernetes/components/apiAccess/endpoints/EndpointsListPage.js
+++ b/src/app/plugins/kubernetes/components/apiAccess/endpoints/EndpointsListPage.js
@@ -2,7 +2,7 @@ import React from 'react'
 import { makeStyles } from '@material-ui/core/styles'
 import createCRUDComponents from 'core/helpers/createCRUDComponents'
 import endpointsActions from './actions'
-import SimpleLink from 'core/components/SimpleLink'
+import ExternalLink from 'core/components/ExternalLink'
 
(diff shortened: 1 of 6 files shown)


src/app/plugins/kubernetes/components/apiAccess/endpoints/EndpointsListPage.js
```diff
@@ -31,7 +31,7 @@ const EndpointsListPage = () => {
     <>
       <h2>API Endpoints</h2>
       <ListPage />
-      <SimpleLink className={classes.link} href="/clarity/docs/qbert/">See qbert API Documentation</SimpleLink>
+      <ExternalLink className={classes.link} url="/clarity/docs/qbert/">See qbert API Documentation</ExternalLink>
```
rahulkurumkar: Can we have URL in `constant.js` file

xantorres: definitely, actually, all hardcoded URL's should be in a constant file, but this is out of the scope of this PR

PR #518, opened 2019-11-27
PMK-2173 PMK-2140 PMK-2021 BareOS cluster scaling support

brennancheung:


src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -265,6 +265,17 @@ export const clusterActions = createCRUDActions(clustersCacheKey, {
...
     },
+    attachNodes: async ({ cluster, nodes }, prevItems) => {
+      await qbert.attach(cluster.uuid, nodes)
+      // TODO: how are we supposed to update the cache here
+      return prevItems
+    },
+    detachNodes: async ({ cluster, nodes }, prevItems) => {
+      await qbert.detach(cluster.uuid, nodes)
+      // TODO: how are we supposed to update the cache here.  Also
+      // TODO: Toast says "Successfully updated cluster" but should be something more specific
```
brennancheung: @xantorres How can we customize the toast notifications here?

brennancheung: Also, it's not really the cluster that needs to be updated in the cache in this instance, it is the nodes.  This falls a bit outside of our normal cache update pattern so maybe I need to go completely outside of `clusterActions`.  What do you think @xantorres

xantorres: You can customize the toast notification by using `successMessage` option. 
In this case I would suggest you to keep using the `clusterActions` even if you don't really update anything in the cache, just for the convenience of the toast notifications and compatibility with our data updater hooks. 
So you can just return `prevItems` without any modifications. Maybe clear the `nodes` cache if you think is necessary by using `loadNodes.invalidateCache()`

src/app/plugins/kubernetes/components/infrastructure/clusters/ScaleWorkersPage.tsx
```diff
@@ -127,8 +118,54 @@ const ScaleWorkers: FunctionComponent<ScaleWorkersProps> = ({ cluster, onSubmit
...
+        validations={[minScaleValidator, maxScaleValidator]}
+        required
+      />
+      <SubmitButton>{params.scaleType === 'add' ? 'Add' : 'Remove'} workers</SubmitButton>
+    </ValidatedForm>
+  )
+
   return (
     <div>
+      {!!params.scaleType || (
```
jruddell: this reads a little confusingly. Wouldn't it be better to  just do the inverse? `!params.scaleType && (`?

src/app/plugins/kubernetes/components/infrastructure/clusters/ScaleWorkersPage.tsx
```diff
@@ -155,27 +194,42 @@ const ScaleWorkersPage: FunctionComponent = () => {
...
 
-  const onComplete = (): void => {
-    history.push(listUrl)
-  }
-
+  const onComplete = () => history.push(listUrl)
   const [update, updating] = useDataUpdater(clusterActions.update, onComplete)
+
+  // TypeScript is not able to infer that the customOperations are actually there so we need to work around it
+  const [attach, isAttaching] = useDataUpdater((clusterActions as any).attachNodes, onComplete)
```
jruddell: dont need to do this, but when we go to clean up the casting to `any`, its easier if you just alias the actions as any once in a variable. `const anyClusterActions: any = clusterActions`. Then you can just use it as normal in the code `useDataUpdater(anyClusterActions.attachNodes`. This would be easier to refactor later. But small thing either way not required.

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/ClusterHostChooser.tsx
```diff
@@ -31,118 +24,118 @@ const useStyles = makeStyles<any, any>((theme: Theme) => ({
...
+export const isMaster = (node: ICombinedNode) => !!node.isMaster
+export const isNotMaster = (node: ICombinedNode) => !node.isMaster
+export const inCluster = (clusterUuid: string) => (node: ICombinedNode) => node.clusterUuid === clusterUuid
+
+// TODO: all the ValidatedForm stuff is in JS and we need the props to be merged
+// into this component.  Refactor this later on when we can convert
+// ValidatedForm.js to TypeScript.
+export interface IValidatedForm {
+  id: string
+  validations?: any[] // What's the best way to type this?
```
jruddell: hmm, each validation uses the FieldValidator (you can use the class as a type). so probably something like `(...args: string[] | number[]) => FieldValidator`? I'd have to test but something kinda like that

PR #519, opened 2019-11-27
Fix logging button

smanpathak:
I discovered this while testing toggle logging button.
The dataloader and contextloader pattern does not work/is not necessary.
The toggle button works on single cluster and the component is passed the cluster as argument. Toggle state is defined by cluster tags. By the time the dialog is loaded, the corresponding cluster is already cached.
I simplified the functionality to just use the tags from cluster arg. The async calls to appbert API are in place already.


smanpathak:
> We recently switched to `yarn` so delete the `package-lock.json` and just run `yarn` instead of `npm install`.
> 
> Also, it doesn't seem like the in-memory cache is being updated with the new tags? Is this something we should update in the cache or does it take some time before the backend will correctly report the tags. If that is the case, we have a slight data inconsistency issue that may cause some confusion.

ok, will delete
Yes, there's invalid cache issue till backend changes tags and fetch. Are there examples of updating cache?

smanpathak:
@brennancheung I have removed package-lock now

PR #522, opened 2019-11-28
Switch color change, removed header content border, reduced header sp…

kevinylee1991:
…acing

Made a few style changes madhura requested

### DIFF
--- a/src/app/core/components/widgets/WidgetCard.js
+++ b/src/app/core/components/widgets/WidgetCard.js
@@ -9,11 +9,17 @@ const useStyles = makeStyles(theme => ({
   },
   header: {
     padding: theme.spacing(1.5, 2),
-    backgroundColor: 'rgba(74, 163, 223, .2)',
+    backgroundColor: theme.palette.card.background,
   },
   content: {
     display: 'flex',
     justifyContent: 'center',
+    backgroundColor: theme.palette.card.background,
+    paddingTop: 0,
+    paddingBottom: theme.spacing(1.5),
+    '&:last-child': {
+      paddingBottom: theme.spacing(1.5),
+    }
   },
   headerContent: {
     display: 'flex',
--- a/src/app/core/containers/AuthenticatedContainer.js
+++ b/src/app/core/containers/AuthenticatedContainer.js
@@ -51,8 +51,8 @@ const useStyles = makeStyles(theme => ({
     display: 'flex',
     alignItems: 'center',
     justifyContent: 'flex-end',
-    padding: '0 8px',
-    ...theme.mixins.toolbar,
+    padding: theme.spacing(0, 1),
+    minHeight: theme.spacing(6),
   },
 }))
 
--- a/src/app/plugins/kubernetes/components/infrastructure/InfrastructurePage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/InfrastructurePage.js
@@ -19,12 +19,13 @@ const useStyles = makeStyles(theme => ({
     alignItems: 'flex-end',
     minHeight: 50,
   },
+
 }))
 
 const StatsToggle = ({ statsVisible, toggleStats }) => {
   return <FormControlLabel
     control={
-      <Switch onChange={toggleStats} checked={statsVisible} />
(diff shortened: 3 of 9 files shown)


src/app/plugins/kubernetes/components/infrastructure/InfrastructurePage.js
```diff
@@ -19,12 +19,13 @@ const useStyles = makeStyles(theme => ({
...
     minHeight: 50,
   },
+
 }))
 
 const StatsToggle = ({ statsVisible, toggleStats }) => {
   return <FormControlLabel
     control={
-      <Switch onChange={toggleStats} checked={statsVisible} />
+      <Switch onChange={toggleStats} checked={statsVisible} color='primary' />
```
brennancheung: double quotes for JSX attributes.

src/app/core/containers/AuthenticatedContainer.js
```diff
@@ -52,7 +52,7 @@ const useStyles = makeStyles(theme => ({
     alignItems: 'center',
     justifyContent: 'flex-end',
     padding: '0 8px',
-    ...theme.mixins.toolbar,
+    minHeight: '48px',
```
brennancheung: Specify from `theme.spacing`.  I believe `1` unit in `spacing` is equal to `8px` right now.  Spacing is a theme value that can be changed as well but we probably don't want to edit it ourselves.

brennancheung:
Also, please include before and after screenshots in the PR for theme changes.

kevinylee1991:
Screen Shot 2019-12-03 at 12 27 12 PM
Screen Shot 2019-12-03 at 12 24 32 PM

src/app/core/components/widgets/WidgetCard.js
```diff
@@ -9,11 +9,14 @@ const useStyles = makeStyles(theme => ({
...
   header: {
     padding: theme.spacing(1.5, 2),
-    backgroundColor: 'rgba(74, 163, 223, .2)',
+    backgroundColor: 'rgba(74, 163, 223, 0.1)',
   },
   content: {
     display: 'flex',
     justifyContent: 'center',
+    backgroundColor: 'rgba(74, 163, 223, 0.1)',
+    paddingTop: theme.spacing(0),
```
brennancheung: I'd just put `0` hard-coded in this case.

src/app/core/components/widgets/WidgetCard.js
```diff
@@ -9,11 +9,14 @@ const useStyles = makeStyles(theme => ({
   },
   header: {
     padding: theme.spacing(1.5, 2),
-    backgroundColor: 'rgba(74, 163, 223, .2)',
+    backgroundColor: 'rgba(74, 163, 223, 0.1)',
```
brennancheung: These values should come from theme.

kevinylee1991:
Screen Shot 2019-12-04 at 10 36 30 AM
Screen Shot 2019-12-04 at 10 36 47 AM
Screen Shot 2019-12-04 at 10 42 49 AM
Screen Shot 2019-12-04 at 10 43 00 AM

src/app/core/components/widgets/WidgetCard.js
```diff
@@ -9,11 +9,14 @@ const useStyles = makeStyles(theme => ({
   },
   header: {
     padding: theme.spacing(1.5, 2),
-    backgroundColor: 'rgba(74, 163, 223, .2)',
+    backgroundColor: theme.palette.common.teal10,
```
brennancheung: Name the palette keys by their key or what they represent, not what color they actually are.  It won't make sense during customization to change `teal10` to be another color.

src/app/core/components/widgets/WidgetCard.js
```diff
@@ -9,11 +9,14 @@ const useStyles = makeStyles(theme => ({
...
   },
   content: {
     display: 'flex',
     justifyContent: 'center',
+    backgroundColor: theme.palette.common.teal10,
+    paddingTop: 0,
+    paddingBottom: [theme.spacing(1.5), '!important'],
```
brennancheung: `!important` shouldn't be required. It's probably a sign we are doing things wrong or modifying the wrong style.

PR #530, opened 2019-12-02
[PMK-2338] Refactor some components & hooks to TS and add notifications popover

xantorres:
Refactored some hooks and AppProvider to Typescript.

Also added a notifications popover that will allow the user to check previous error messages and copy them if he wants to do so. Error notifications are stored in the global app context and will get reset when the user logs out.

image


src/app/core/hooks/useStateAsync.ts
```diff
@@ -0,0 +1,38 @@
+import { useState, useCallback, useRef, useEffect, SetStateAction } from 'react'
+
+const useStateAsync = <T>(initialState) => {
+  const resolvers = useRef<Array<(state: T) => void>>([])
+  const [state, setState] = useState<T>(initialState)
+
+  const setStateAsync = useCallback(async (stateUpdater: SetStateAction<T>) => {
```
brennancheung: Looks like you are hard-coding setting state with a Promise.  Wondering if it makes more sense to generalize the concept of setting state to any functor much like Ramda does.  That would allow any kind of Functor such as lenses, promises (with some modification), observables, arrays, etc, to have setState functionality all with the same function and interface.

xantorres: Actually, this `setStateAsync` function allows a Functor, you can pass anything of type `T`, which can be either a set function (passing a lens or whatever) or a hard value as an array or any object.
This just behaves like the nomal `useState` hook with the advantage of it returning a promise that will be fullfilled when the state update has been performed.

PR #533, opened 2019-12-03
PMK-2335: Fixed the reset password functionality.

rahulkurumkar:
Fixed the reset password functionality.


src/app/constants.js
```diff
@@ -4,10 +4,10 @@ export const appUrlRoot = '/ui'
 export const imageUrlRoot = `${appUrlRoot}/images`
 export const loginUrl = `${appUrlRoot}/login`
 export const loginWithCookieUrl = `${appUrlRoot}/pmkft/login`
-export const resetPasswordUrl = `${appUrlRoot}/reset_password`
+export const resetPasswordUrl = `${appUrlRoot}/reset/password`
```
brennancheung: I don't like this naming schema.  I like `reset_password` better.  REST is always noun first.

rahulkurumkar: Backend is sending the as a URL when reset email is sent. We need to ask them to change the URL.

brennancheung: I see.  In that case it's fine then.

PR #538, opened 2019-12-04
PMK-2351 change Qbert doc link to doc website

brennancheung:
Change Qbert API doc link to the doc website instead of the version hosted on the DU.

### DIFF
--- a/src/app/plugins/kubernetes/components/apiAccess/endpoints/EndpointsListPage.js
+++ b/src/app/plugins/kubernetes/components/apiAccess/endpoints/EndpointsListPage.js
@@ -31,7 +31,7 @@ const EndpointsListPage = () => {
     <>
       <h2>API Endpoints</h2>
       <ListPage />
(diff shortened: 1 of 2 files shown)


src/app/plugins/kubernetes/components/apiAccess/endpoints/EndpointsListPage.js
```diff
@@ -31,7 +31,7 @@ const EndpointsListPage = () => {
     <>
       <h2>API Endpoints</h2>
       <ListPage />
-      <ExternalLink className={classes.link} url="/clarity/docs/qbert/">See qbert API Documentation</ExternalLink>
+      <ExternalLink className={classes.link} url="https://docs.platform9.com/kubernetes/Qbert-API-Reference">See Qbert API documentation</ExternalLink>
```
rahulkurumkar: Can we move this to `constant.js` like https://github.com/platform9/pf9-ui-plugin/blob/master/src/app/constants.js#L12
It would be great if we have a dedicated file for all URLs so that in future it won't miss the places where it needs to replaced and easy to replace.

PR #539, opened 2019-12-04
Style tweaks

brennancheung:
Changed the primary and secondary back to their original values.  Really just the top nav was meant to be grey.

Quantized font size values so they are closer.  Having 12px, 13px, 14px, 15px, and 16px all on the same page was really bad.

Tweaked, dashboard colors slightly and added a subtle drop shadow.

Tweaked form look and and feel.

Screen Shot 2019-12-06 at 5 15 43 PM
Screen Shot 2019-12-06 at 4 20 34 PM
Screen Shot 2019-12-06 at 4 20 17 PM

### DIFF
--- a/src/app/core/components/validatedForm/TextField.js
+++ b/src/app/core/components/validatedForm/TextField.js
@@ -11,8 +11,8 @@ const styles = () => ({
   // https://github.com/mui-org/material-ui/issues/14530
   label: {
     backgroundColor: 'white',
-    // padding: '0 5px',
-    // margin: '0 -5px',
+    padding: '0 5px',
+    margin: '0 -5px',
   },
 })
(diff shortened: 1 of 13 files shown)


src/app/core/components/validatedForm/TextField.js
```diff
@@ -11,8 +11,8 @@ const styles = () => ({
   // https://github.com/mui-org/material-ui/issues/14530
   label: {
     backgroundColor: 'white',
-    padding: '0 5px',
-    margin: '0 -5px',
+    // padding: '0 5px',
+    // margin: '0 -5px',
```
xantorres: Why? This was about the fix I did for the label being line crossed

PR #540, opened 2019-12-04
PMK-2328: Disabled the "Network backend" option when the user selects "use platform9.net"

rahulkurumkar:
PMK-2328: Disabled the "Network backend" option when the user selects "use platform9.net"

PMK-2328Cannot choose network type unless use platform9 net is unchecked

PMK-2354: Tripling the current padding for the wizard. I also added a prop to skip it. Moved the URL to constant.js from AddResourcePage.

PMK-2354-Wizard input fields should have more left side margin

### DIFF
--- a/src/app/core/components/validatedForm/PicklistField.js
+++ b/src/app/core/components/validatedForm/PicklistField.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import Picklist from 'core/components/Picklist'
 import InfoTooltip from 'app/core/components/InfoTooltip'
 import withFormContext, { ValidatedFormInputPropTypes } from 'core/components/validatedForm/withFormContext'
-import { compose } from 'utils/fp'
+import { compose, noop } from 'utils/fp'
 import { makeStyles } from '@material-ui/styles'
 
 const useStyles = makeStyles(() => ({
@@ -23,7 +23,7 @@ const useStyles = makeStyles(() => ({
 const PicklistField = React.forwardRef(({
   DropdownComponent, id, info, placement,
   label, required, value, showNone, updateFieldValue, getCurrentValue,
-  hasError, errorMessage, options, ...restProps
+  hasError, errorMessage, options, disabled, ...restProps
 }, ref) => {
   const classes = useStyles()
   const [open, setOpen] = React.useState(false)
@@ -35,6 +35,7 @@ const PicklistField = React.forwardRef(({
     <InfoTooltip open={open} info={info} placement={placement}>
       <DropdownComponent
         {...restProps}
+        disabled={disabled}
         InputLabelProps={{
           classes: {
             root: classes.label,
@@ -42,9 +43,9 @@ const PicklistField = React.forwardRef(({
         }}
         formField
         label={required ? `${label} *` : label}
-        onMouseEnter={openTooltip}
+        onMouseEnter={disabled ? noop : openTooltip}
         onMouseLeave={closeTooltip}
-        onFocus={openTooltip}
+        onFocus={disabled ? noop : openTooltip}
         onBlur={closeTooltip}
         onClick={closeTooltip}
         ref={ref}
--- a/src/app/core/components/wizard/Wizard.js
+++ b/src/app/core/components/wizard/Wizard.js
@@ -1,6 +1,7 @@
 import React, { PureComponent } from 'react'
 import { withRouter } from 'react-router-dom'
 import PropTypes from 'prop-types'
+import { withStyles } from '@material-ui/styles'
 import WizardButtons from 'core/components/wizard/WizardButtons'
 import NextButton from 'core/components/buttons/NextButton'
 import PrevButton from 'core/components/buttons/PrevButton'
@@ -9,7 +10,13 @@ import WizzardStepper from 'core/components/wizard/WizardStepper'
 import CancelButton from 'core/components/buttons/CancelButton'
 
 export const WizardContext = React.createContext({})
+const styles = (theme) => ({
+  wizardContainer: {
+    paddingLeft: theme.spacing(9),
+  },
+})
 
+@withStyles(styles)
 class Wizard extends PureComponent {
   isLastStep = () => this.state.activeStep === this.state.steps.length - 1
   isComplete = () => this.state.activeStep > this.state.steps.length - 1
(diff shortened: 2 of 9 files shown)


rahulkurumkar:
Updated code to disable the network option if user uncheck "use Platform9.net"

PMK-2328Cannot choose network type unless use platform9 net is unchecked

src/app/core/components/wizard/Wizard.js
```diff
@@ -108,21 +115,34 @@ class Wizard extends PureComponent {
...
-    const { showSteps, children, submitLabel, finishAndReviewLabel, onCancel, showFinishAndReviewButton } = this.props
+    const {
+      showSteps,
+      children,
+      submitLabel,
+      finishAndReviewLabel,
+      onCancel,
+      showFinishAndReviewButton,
+      classes,
+      disableGutters,
```
jruddell: What is this for? why are we making gutters and a flag to disable them?

rahulkurumkar: Currently, we do not have any padding for wizard fields. I have added it for wizard by default it will triple the current one if anyone wants to disable that it can be done as I saw in some places it is not required example image attached below. If we try to implement it Page wise then I am okay for it but the downside of it is will create inconsistencies in UI depending on who is implementing page and how much padding he is adding. To avoid that we can have standards defined about the margin, padding, font size, etc.

image

rahulkurumkar: Currently, we do not have any padding for wizard fields. I have added it for wizard by default it will triple the current one if anyone wants to disable that it can be done as I saw in some places it is not required example image attached below. If we try to implement it Page wise then I am okay for it but the downside of it is will create inconsistencies in UI depending on who is implementing page and how much padding he is adding. To avoid that we can have standards defined about the margin, padding, font size, etc.

image

src/app/core/components/wizard/Wizard.js
```diff
@@ -9,7 +10,13 @@ import WizzardStepper from 'core/components/wizard/WizardStepper'
 import CancelButton from 'core/components/buttons/CancelButton'
 
 export const WizardContext = React.createContext({})
+const styles = (theme) => ({
+  wizardContainer: {
+    paddingLeft: theme.spacing(9),
```
xantorres: Please use jss props instead of conditionally setting the class, eg
```
  wizardContainer: {
    paddingLeft: ({ disableGutters }) => disableGutters ? 0 : theme.spacing(9),
```

src/app/core/components/wizard/Wizard.js
xantorres: I think this `gutters` padding should not be even in the `Wizard`, but in the `PageContainer` instead

src/app/plugins/kubernetes/components/namespaces/AddNamespacePage.js
```diff
@@ -12,20 +13,29 @@ import { namespaceValidator } from 'core/utils/fieldValidators'
...
-      <PicklistField
-        DropdownComponent={ClusterPicklist}
-        id="clusterId"
-        label="Cluster"
-        onChange={getParamsUpdater('clusterId')}
-        value={params.clusterId}
-        required
-      />
-      <SubmitButton>Add Namespace</SubmitButton>
+      <div className={classes.root}>
```
xantorres: please add the class to the `ValidatedForm` itself, we should avoid cluttering the DOM tree with extra containers whenever possible.

src/app/plugins/kubernetes/components/pods/AddResourcePage.js
```diff
@@ -20,12 +21,11 @@ import Progress from 'core/components/progress/Progress'
...
 
-const createPodUrl =
-  'https://kubernetes.io/docs/tasks/configure-pod-container/communicate-containers-same-pod/#creating-a-pod-that-runs-two-containers'
-const createDeploymentUrl =
-  'https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#creating-a-deployment'
-const createServiceUrl =
-  'https://kubernetes.io/docs/tutorials/connecting-apps/connecting-frontend-backend/#creating-the-backend-service-object'
+const useStyles = makeStyles((theme) => ({
+  root: {
+    paddingLeft: theme.spacing(9),
```
xantorres: This looks like a generic spacing that should be either added in the `FormWrapper` component or the `PageContainer`

src/app/core/components/validatedForm/PicklistField.js
```diff
@@ -35,16 +35,17 @@ const PicklistField = React.forwardRef(({
...
+        disabled={disabled}
         InputLabelProps={{
           classes: {
             root: classes.label,
           },
         }}
         formField
         label={required ? `${label} *` : label}
-        onMouseEnter={openTooltip}
+        onMouseEnter={disabled ? noop : openTooltip}
```
jruddell: Although this is technically fine, it would be better to explicitly pass `undefined` rather than noop. We dont need a callback function to run even if its an empty function.

src/app/core/components/wizard/Wizard.js
jruddell: Remove the `gutter` concept from the wizard. We dont want to put that here. Components should be smaller and less complicated. This makes them like an API where we can use it in multiple places. Common layout similarities could go in `PageContainer` or can implement their own version depending on the use case.

src/app/core/components/wizard/Wizard.js
```diff
@@ -108,21 +115,34 @@ class Wizard extends PureComponent {
...
-        {renderStepsContent({ wizardContext, setWizardContext, onNext: this.onNext })}
-        <WizardButtons>
-          {onCancel && <CancelButton onClick={onCancel} />}
-          {this.hasBack() && <PrevButton onClick={this.handleBack} />}
-          {this.canBackAtFirstStep() && <PrevButton onClick={this.handleOriginBack} />}
-          {this.hasNext() && <NextButton onClick={this.handleNext}>Next</NextButton>}
-          {this.isLastStep() && <NextButton onClick={this.handleNext}>{submitLabel}</NextButton>}
-          {showFinishAndReviewButton && this.isFinishAndReviewVisible() && <NextButton onClick={this.onFinishAndReview}>{finishAndReviewLabel}</NextButton>}
-        </WizardButtons>
+        <div className={disableGutters ? '' : classes.wizardContainer}>
```
jruddell: can remove the extra element with no gutters

src/app/plugins/kubernetes/components/namespaces/AddNamespacePage.js
jruddell: 👍

src/app/plugins/kubernetes/components/pods/AddResourcePage.js
jruddell: @rahulkurumkar @xantorres do we have a design / style guide that references this spacing of 9 units? I dont think `FormWrapper` should contain the padding for the same argument that wizard shouldn't. Forms are just forms. If we need spacing then parents should implement or use a common layout element.

src/app/core/components/wizard/Wizard.js
brennancheung: I agree, we should not have a component rendered 2 different ways in terms in spacing.  This makes it so now whenever we are doing design work we have to consider 2 possibilities and possibly do negative margins conditionally on the parent component.

src/app/plugins/kubernetes/components/pods/AddResourcePage.js
xantorres: I don't know about that, I assumed this is a change requested by product, but Rahul will tell.

rahulkurumkar: @jruddell  I don't think we have design guide @brennancheung do have design guide for UI?
PMK-2354 | Wizard input fields should have more left side margin
This ticket was created by Madhura

brennancheung:
We can probably cancel this one out in favor of #555

PR #541, opened 2019-12-04
[PMK-2333] Fix user & tenants roles assignment

xantorres:
* Branched off from https://github.com/platform9/pf9-ui-plugin/pull/530
* Fixed roles assignment when adding/editing Users & Tenants
* Fix patch request headers (fixes edit forms)
* Fix performance issue due to "registerNotification" being instantiated multiple times upon renders
* Fix minor styling issues
* Removed Groups tab to match the Clarity UI
* Fix region being written in user preferences twice

### DIFF
--- a/src/api-client/ApiClient.js
+++ b/src/api-client/ApiClient.js
@@ -120,7 +120,6 @@ class ApiClient {
 
   basicPatch = async (url, body) => {
     const config = this.getAuthHeaders()
-    config.headers['Content-Type'] = 'application/json-patch+json'
     const response = await this.axiosInstance.patch(url, body, config)
     return normalizeResponse(response)
   }
--- a/src/app/core/components/notificationsPopover/NotificationsPopover.tsx
+++ b/src/app/core/components/notificationsPopover/NotificationsPopover.tsx
@@ -40,7 +40,7 @@ const usePopoverStyles = makeStyles(theme => ({
     flexFlow: 'column nowrap',
     padding: 0,
     backgroundColor: '#FFF',
-    marginTop: 40,
+    marginTop: theme.spacing(5),
     width: 400,
     maxHeight: 600,
     border: 0,
@@ -120,6 +120,5 @@ const NotificationsPopover = ({ className }) => {
     </Popover>
   </div>
 }
-
 
 export default NotificationsPopover
--- a/src/app/core/components/notificationsPopover/NotificationsPopover.tsx
+++ b/src/app/core/components/notificationsPopover/NotificationsPopover.tsx
@@ -1,16 +1,10 @@
-import React, { useCallback, useContext } from 'react'
+import React, { useContext } from 'react'
 import { Popover, Typography, Tooltip, Button } from '@material-ui/core'
 import { makeStyles } from '@material-ui/core/styles'
 import FontAwesomeIcon from 'core/components/FontAwesomeIcon'
 import { AppContext } from 'core/providers/AppProvider'
-import { over, lensProp, prepend, always, take } from 'ramda'
-import uuid from 'uuid'
 import NotificationItem
   from 'core/components/notificationsPopover/NotificationItem'
-import moment from 'moment'
-import { emptyArr, pipe } from 'utils/fp'
-
-const maxNotifications = 30
 
 const useStyles = makeStyles(theme => ({
   title: {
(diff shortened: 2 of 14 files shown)


src/api-client/ApiClient.js
```diff
@@ -120,7 +120,6 @@ class ApiClient {
 
   basicPatch = async (url, body) => {
     const config = this.getAuthHeaders()
-    config.headers['Content-Type'] = 'application/json-patch+json'
```
brennancheung: This is needed in OpenStack.  That's why it's there.  Don't remove unless you are seeing error.

src/app/core/components/notificationsPopover/NotificationsPopover.tsx
```diff
@@ -0,0 +1,125 @@
...
+}))
+
+const usePopoverStyles = makeStyles(theme => ({
+  paper: {
+    overflow: 'visible',
+    display: 'flex',
+    flexFlow: 'column nowrap',
+    padding: 0,
+    backgroundColor: '#FFF',
+    marginTop: 40,
```
brennancheung: Use spacing here.  40 = 5 spacing = 5 x 8

src/app/plugins/kubernetes/components/userManagement/users/EditUserPage.js
```diff
@@ -40,16 +40,17 @@ const useStyles = makeStyles(theme => ({
   },
 }))
 
-const ToggableTextField = ({ id, label, value, required = false, TextFieldComponent = TextField }) => {
+const ToggableTextField = ({ id, label, initialValue, value, required = false, TextFieldComponent = TextField }) => {
```
kevinylee1991: Toggable -> Togglable

src/app/plugins/kubernetes/components/userManagement/users/EditUserPage.js
```diff
@@ -93,9 +94,11 @@ const EditUserPage = () => {
       {({ wizardContext, setWizardContext, onNext }) => <>
         <WizardStep stepId="basic" label="Basic Info">
           <ValidatedForm initialValues={wizardContext} onSubmit={setWizardContext} triggerSubmit={onNext}>
-            <ToggableTextField id="username" label="Username or Email" value={user.username} required />
-            <ToggableTextField id="displayname" label="Display Name" value={user.displayname} />
-            <ToggableTextField id="password" label="Password" value={user.password || '********'} TextFieldComponent={UserPasswordField} />
+            {({ values }) => <>
+              <ToggableTextField id="username" label="Username or Email" initialValue={user.username} required />
```
kevinylee1991: Togglable

src/app/plugins/kubernetes/components/userManagement/users/EditUserPage.js
```diff
@@ -93,9 +94,11 @@ const EditUserPage = () => {
...
+              <ToggableTextField id="displayname" label="Display Name" initialValue={user.displayname} />
+              <ToggableTextField id="password" label="Password" initialValue={user.password || '********'} value={values.password} TextFieldComponent={UserPasswordField} />
```
kevinylee1991: I don't think password will ever be present -- we could probably remove the initial value for the password field, unless it is required for the toggle to function properly

src/api-client/ApiClient.js
xantorres: I do, adding a comment in the main axios config

src/app/core/providers/AppProvider.tsx
```diff
@@ -84,7 +95,26 @@ const AppProvider = ({ children }) => {
       ? setterFn
       : context => ({ ...context, ...setterFn })
     )
-  }, [context, setContextBase])
+  }, [])
+
+  const notifLens = lensProp('notifications')
+  const registerNotification = useCallback(async (title, message, type) => {
```
kevinylee1991: Tried looking but couldn't find -- are these notifications going to be stored in localStorage somewhere like the old UI? Should they be using the same localStorage error notifications so that errors are shared between the old and new UI?

src/app/plugins/kubernetes/components/userManagement/users/EditUserPage.js
xantorres: good catch

src/app/core/providers/AppProvider.tsx
xantorres: no, these notifications can not be stored in localStorage, because then any user could see them, potentially leaking protected data.

src/app/plugins/kubernetes/components/userManagement/users/EditUserPage.js
xantorres: actually, I think it's "toggleable"

xantorres: fixed

PR #546, opened 2019-12-04
Bugfix/pmk 2262 cluster status update

murilovmachado:
This PR changes the way the `connection` and `health` status are calculated and split them into separate columns.

Screen Shot 2019-12-04 at 19 19 36


src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
```diff
@@ -43,81 +47,212 @@ const renderCloudProviderType = (type, cluster) => {
...
 })(taskStatus)
 
-const renderStatus = (status,
-  { highlyAvailable, healthyMasterNodes, masterNodes, cloudProviderType, progressPercent, taskStatus }) => {
-  if (!status || !taskStatus) {
-    // TODO probably a better way to handle this.
-    // But if we get undefined status The UI currently crashes
-    return null
+const getConnectionStatus = (nodes) => {
+  let connectionStatus
```
brennancheung: You don't need a variable.  Just return the value immediately.  This also eliminates the `else` clauses.

`node => node.status === 'ok' || node.status === 'failed'` is common to both branches.  Make it a simple 1-line function.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
```diff
@@ -43,81 +47,212 @@ const renderCloudProviderType = (type, cluster) => {
...
-        title={getClusterPopoverContent(healthyMasterNodes, masterNodes)}
-        status={highlyAvailable ? 'ok' : 'fail'}>
-        {highlyAvailable ? 'Healthy' : 'Unhealthy'}
+
+  return connectionStatus
+}
+
+const renderConnectionStatus = (_, { nodes }) => {
+  const connectionStatus = getConnectionStatus(nodes)
+  let message
```
brennancheung: Same here.  Don't use variables.  Actually you can just hard-code these in a map.

Pseudo-code:

```javascript
const fieldsByStatus = {
  connected: { message, clusterStatus, label },
  disconnected: { ... },
  ...
}
const fields = fieldsByStatus[connectionStatus]
````

File: src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js [hunk 3]
```diff
@@ -43,81 +47,212 @@ const renderCloudProviderType = (type, cluster) => {
...
+      title={message}
+      status={clusterStatus}>
+      {label}
+    </ClusterStatusSpan>
+  )
+}
+
+const getHealthStatusAndMessage = (healthyMasterNodes = [], nodes, numMasters, numWorkers) => {
+  const healthyMasterNodesCount = healthyMasterNodes.length
+  const healthyWorkersNodesCount = nodes.filter(node => !node.isMaster && node.status === 'ok').length
```
brennancheung: Protect against `nodes` being `undefined`.  You can probably just do `nodes = []` in the function params.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
```diff
@@ -43,81 +47,212 @@ const renderCloudProviderType = (type, cluster) => {
...
+  }
+
+  return healthStatus
+}
+
+const getClusterHealthStatusAndMessage = (mastersHealthStatus, workersHealthStatus) => {
+  let healthStatus
+  let message
+
+  if (mastersHealthStatus === healthy && workersHealthStatus === healthy) {
```
brennancheung: Create a lookup table for these instead of all the crazy conditionals.  Just determine a few of the conditional states and use those to key/index the status.

You might able to do something similar to this pattern:
https://github.com/platform9/pf9-ui-plugin/blob/master/src/app/plugins/kubernetes/components/infrastructure/clusters/ScaleMastersPage.tsx#L46

src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
```diff
@@ -43,81 +47,212 @@ const renderCloudProviderType = (type, cluster) => {
...
+  return [healthStatus, message]
+}
+
+const isConverging = (nodes) => !!nodes.find(node => node.status === 'converging')
+
+const isSteadyState = (taskStatus, nodes) =>
+  !isConverging(nodes) && (taskStatus === 'success' || taskStatus === 'error')
+
+const isTransientState = (taskStatus, nodes) =>
+  taskStatus === 'creating' ||
```
brennancheung: Use `includes`.

`[...].includes(taskStatus)`

Or maybe it's simpler to test the negative cases if there are fewer of them.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
```diff
@@ -43,81 +47,212 @@ const renderCloudProviderType = (type, cluster) => {
...
-      return <ClusterStatusSpan>{capitalizeString(status)}</ClusterStatusSpan>
-    }
+    )
+  } else if (status) {
+    return <ClusterStatusSpan>{capitalizeString(status)}</ClusterStatusSpan>
+  }
+}
+
+const renderClusterHealthStatus = (healthyMasterNodes, nodes, numMasters, numWorkers) => {
+  const [healthStatus, message] = getHealthStatusAndMessage(healthyMasterNodes, nodes, numMasters, numWorkers)
```
brennancheung: Same pattern here.  Use an object and look them up.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
```diff
@@ -43,81 +47,212 @@ const renderCloudProviderType = (type, cluster) => {
...
+const renderHealthStatus = (status, {
+  taskStatus,
+  cloudProviderType,
+  progressPercent,
+  healthyMasterNodes,
+  nodes,
+  numMasters,
+  numWorkers,
+}) => {
+  if (isTransientState(taskStatus, nodes)) {
```
brennancheung: Don't need `else` if you already have `return`ed.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
```diff
@@ -43,81 +47,212 @@ const renderCloudProviderType = (type, cluster) => {
...
+  }
+}
+
+const renderClusterHealthStatus = (healthyMasterNodes, nodes, numMasters, numWorkers) => {
+  const [healthStatus, message] = getHealthStatusAndMessage(healthyMasterNodes, nodes, numMasters, numWorkers)
+  let status
+  let label
+
+  switch (healthStatus) {
+    case 'healthy':
```
brennancheung: You already have these strings defined, don't hard-code strings again.

PR #548, opened 2019-12-05
[PMK-2312] Disable App Catalog

xantorres:


src/app/plugins/kubernetes/index.js
```diff
@@ -434,16 +435,17 @@ Kubernetes.registerPlugin = pluginManager => {
...
-      name: 'App Catalog',
-      link: { path: '/apps' },
-      icon: 'th',
-      nestedLinks: [
-        { name: 'App Catalog', link: { path: '/apps#appCatalog' } },
-        { name: 'Deployed Apps', link: { path: '/apps#deployedApps' } },
-        { name: 'Repositories', link: { path: '/apps#repositories' } },
-      ],
-    },
+    // TODO: enable this when backend side is implemented
```
rahulkurumkar: Can we just have flag `disable: true` so that router creation code will look for this and then build a route tree?

xantorres: In this case I don't want to disable it, just remove it entirely until we can re-enable it again. 
Also I don't think it's worth the effort of adding a `disable` option for the few (or none) cases that we will be using it.

PR #551, opened 2019-12-06
PMK-2335: Added missing "/" in URL. Minor change

rahulkurumkar:


src/app/core/containers/AppContainer.js
```diff
@@ -82,7 +82,7 @@ const AppContainer = props => {
 
     if (history.location.pathname === forgotPasswordUrl) return history.push(forgotPasswordUrl)
 
-    if (history.location.hash.includes(resetPasswordThroughEmailUrl)) return history.push(history.location.hash.slice().replace('#', 'ui'))
+    if (history.location.hash.includes(resetPasswordThroughEmailUrl)) return history.push(history.location.hash.slice().replace('#', '/ui'))
```
xantorres: this is a really ugly workaround, please add a FIXME comment explaining why we are doing this so that we remove it as soon as the backend corrects the url.

rahulkurumkar: Below URL will be formed when user clicks o`https://du-test-kplane-pacharya-2738.platform9.horse/ui/index.html#/reset/password/d484160fa02642741280086bd8476b752dd8fd8cdf93fb24f6c9808ebcd3801d02802fbe763a0a6aec28f2525ce3211b3a94e610b5b00263e0b35a9e0320c06f`

At the point of redirection, we get below data in the location
image

As a part of the password reset, we need to sent token in URL and for that, I am using the hash part of history and forming the correct URL.

In the future we can clean this after syncing up with backend and OLD UI.

xantorres: I understand, but please add a `FIXME` comment explaining this in the code itself.

PR #555, opened 2019-12-09
PMK-2328: Disabled the "Network options" option when the user selects "use platform9.net"

rahulkurumkar:
PMK-2328: Disabled the "Network options" option when the user selects "use platform9.net"

PMK-2328Cannot choose network type unless use platform9 net is unchecked


src/app/plugins/kubernetes/components/infrastructure/clusters/AddAwsClusterPage.js
```diff
@@ -457,24 +465,21 @@ const AddAwsClusterPage = () => {
                         wizardContext,
                       })}
 
-                      {/* API FQDN */}
-                      {values.usePf9Domain ||
-                      <TextField
+                      {/* API FQDN */}<TextField
```
jruddell: drop `<TextField` to a new line here

PR #556, opened 2019-12-09
PMK-2394: Added code to not show tooltip if component is in disabled state

rahulkurumkar:
Added code to not show tooltip if the component is in a disabled state

DoNotShowToolTipifComponentIsDisabled


brennancheung:
Did someone ask for this feature specifically?  I think the product would be worse off if we made this change.  I think it is useful to understand what a field is even if it is disabled; it provides more context as to why it might be disabled.

brennancheung:
Product confirmed no need for this.  Closing this PR.

PR #557, opened 2019-12-09
[PMK-2393] Create Clemency service to generate new user activation email

xantorres:


src/app/plugins/kubernetes/components/userManagement/users/actions.js
```diff
@@ -35,14 +35,23 @@ export const mngmUserActions = createCRUDActions(mngmUsersCacheKey, {
...
-      email: username,
-      name: username,
-      username,
-      displayname,
-      password: password || undefined,
-      default_project_id: head(keys(roleAssignments)),
-    })
+    const createdUser = password
+      ? await keystone.createUser({
+        email: username,
```
jruddell: is `username` here the email address?

src/api-client/ApiService.ts
```diff
@@ -0,0 +1,12 @@
+import ApiClient from 'api-client/ApiClient'
+
+abstract class ApiService {
+  protected constructor (protected client: ApiClient) {
+  }
+
+  abstract endpoint(): string | Promise<string>
```
jruddell: We should probably only have one type here. either `string` or `Promise<string>`. Feels like this will make usage harder because we dont know which type it is. Thoughts?

src/api-client/Clemency.ts
```diff
@@ -0,0 +1,44 @@
...
+    return `${this.endpoint}/clemency`
+  }
+
+  requestNewPassword = async (username) => {
+    const body = { username }
+    return this.client.basicPost(`${this.baseUrl}/request`, body)
+  }
+
+  verifyPasswordReset = async (secret) => {
+    return this.client.basicGet(`${this.baseUrl}/reset/password/#${secret}`)
```
jruddell: is the `#` intended here? is that supposed to be a `?` for QueryString parameters? or is the API really requiring a `#`?

src/app/plugins/kubernetes/components/userManagement/users/actions.js
brennancheung: Yes, we treat them the same for our purposes.  OpenStack has tons of fields that are all basically the same thing.

src/api-client/ApiService.ts
brennancheung: I'd lean towards `Promise<string>`.  Things can be composed better but how much will this require everything else to be changed?  Also, the con would be that code that is already simple would become infected with `await`s and make things more verbose.  I could go either way.

src/api-client/Clemency.ts
brennancheung: I think this is the backend setting this requirement and it's probably related to trying to be compatible with the old UI.

xantorres: I just copied and pasted this url from the old ui, literally

src/api-client/ApiService.ts
xantorres: I just checked our current services and seen that in some places we resolve the url by fetching some services, whereas other times we just return a hardcoded value

jruddell: If its an abstract class where we are extending other classes from I'd really prefer one type. Why do we have different return types for getting an endpoint?

PR #559, opened 2019-12-09
PMK-2397 only allow connected nodes to be added to cluster

brennancheung:


src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/ClusterHostChooser.tsx
```diff
@@ -24,6 +24,7 @@ const useStyles = makeStyles<any, any>((theme: Theme) => ({
   },
 }))
 
+export const isConnected = (node: ICombinedNode) => node.status === 'ok'
```
jruddell: we should probably start thinking about a cluster utils file. This will get messy as to where our utility functions are as we add more and more.

brennancheung: SG, we can do that in the future.  I'd also like to see some of the status things that are just `status: string` have the enumerated values but I'm not confident of all the possible backend values yet.

PR #562, opened 2019-12-10
Prompt to create cloud provider when none of type exist

jruddell:
Fixes PMK-2348

added logic for linking to creating a cloud provider when there are none of that type

deep linking to select the correct provider on first mount


src/app/plugins/kubernetes/components/infrastructure/cloudProviders/PromptToAddProvider.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+interface Props {
+  type: CloudProviders
+  src: string
+}
+export const PromptToAddProvider = ({ type, src }: Props) => {
+  const humanReadableType = capitalizeString(type)
+  const { history } = useReactRouter()
+  const handleClick = () => {
+    console.error(src)
```
brennancheung: remove console

PR #564, opened 2019-12-10
[PMK-2326] Logout user on session expiration

xantorres:
Also converted remaining Api Services to Typescript

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -42,7 +42,6 @@
     "no-multi-spaces": [2, { "exceptions": { "ObjectExpression": true } } ],
     "key-spacing": [2, { "mode": "minimum" }],
     "standard/no-callback-literal": 0,
-    "space-before-function-paren": "off",
     "@typescript-eslint/strict-boolean-expressions": "off",
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/no-floating-promises": "off"
--- a/.eslintrc
+++ b/.eslintrc
@@ -42,8 +42,10 @@
     "no-multi-spaces": [2, { "exceptions": { "ObjectExpression": true } } ],
     "key-spacing": [2, { "mode": "minimum" }],
     "standard/no-callback-literal": 0,
+    "space-before-function-paren": "off",
     "@typescript-eslint/strict-boolean-expressions": "off",
-    "@typescript-eslint/explicit-function-return-type": "off"
+    "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/no-floating-promises": "off"
   },
 
   "settings": {
--- a/src/api-client/Cinder.ts
+++ b/src/api-client/Cinder.ts
@@ -9,7 +9,7 @@ class Cinder extends ApiService {
 
   volumesUrl = async () => `${await this.endpoint()}/volumes`
 
-  async getRegionUrls() {
+  async getRegionUrls () {
     const services = (await this.client.keystone.getServiceCatalog()).find(
       x => x.name === 'cinderv3').endpoints
     const baseUrlsByRegion = services.reduce((accum, service) => {
@@ -19,26 +19,26 @@ class Cinder extends ApiService {
     return baseUrlsByRegion
   }
 
-  async getVolume(id) {
+  async getVolume (id) {
     const url = `${await this.volumesUrl()}/${id}`
     const response = await axios.get(url, this.client.getAuthHeaders())
     return response.data.volume
   }
 
   // Get volumes with details
-  async getVolumes() {
+  async getVolumes () {
     const url = `${await this.volumesUrl()}/detail`
     const response = await axios.get(url, this.client.getAuthHeaders())
     return response.data.volumes
   }
 
-  async getAllVolumes() {
+  async getAllVolumes () {
     const url = `${await this.volumesUrl()}/detail?all_tenants=1`
     const response = await axios.get(url, this.client.getAuthHeaders())
     return response.data.volumes
   }
 
-  async getAllVolumesCount(limit, allTenants, markerId) {
+  async getAllVolumesCount (limit, allTenants, markerId) {
     const baseUrl = `${await this.volumesUrl()}/detail`
     const limitUrl = `?limit=${limit}`
     const projectUrl = allTenants ? '&all_tenants=1' : ''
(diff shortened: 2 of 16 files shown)


src/api-client/Cinder.ts
```diff
@@ -1,18 +1,15 @@
...
+import ApiService from 'api-client/ApiService'
 
-class Cinder {
-  constructor (client) {
-    this.client = client
-  }
-
-  async endpoint () {
+class Cinder extends ApiService {
+  endpoint = () => {
```
jruddell: wont this break line 12? ```volumesUrl = async () => `${await this.endpoint()}/volumes` ``` using await when endpoint is no longer async? (this is why I dont like multiple return types on the base abstract class).

src/api-client/Qbert.ts
```diff
@@ -1,11 +1,23 @@
...
-const normalizeClusterizedResponse = (clusterId, response) => propOr([], 'items', response).map(
-  x => ({ ...x, clusterId }))
-const normalizeClusterizedUpdate = (clusterId, response) => ({ ...response, clusterId })
+interface GenericObject {
+  [key: string]: any
+}
+
+// TODO: Fix these typings
+const normalizeClusterizedResponse = (clusterId: string, response: GenericObject) =>
+  pipe<GenericObject, GenericObject[], Array<GenericObject & { clusterId: string }>>(
```
brennancheung: Can you leave the typing off here if it is generic?  Or maybe just type the return of the entire function only.  This code is more type information that actual executable code.  It's not helping the readability.

src/api-client/Qbert.ts
```diff
@@ -441,9 +450,13 @@ class Qbert {
   }
 
   createPrometheusInstance = async (clusterId, data) => {
-    const requests = {}
-    if (data.cpu) { requests.cpu = data.cpu }
-    if (data.memory) { requests.memory = data.memory }
+    const requests: { cpu?: string, memory?: string } = {}
+    if (data.cpu) {
```
brennancheung: If prefer the original on 1 line.

src/app/core/containers/AppContainer.js
```diff
@@ -50,6 +55,21 @@ const AppContainer = props => {
...
 
+  useEffect(() => {
+    const id = setInterval(() => {
+      if (!isEmpty(session)) {
+        // Check if session has expired
+        const { expiresAt } = session
+        const sessionExpired = moment().isAfter(expiresAt)
+        if (sessionExpired) {
+          destroySession()
+          showToast('The session has expired, please login again', 'warning')
```
brennancheung: Spelling should be `log in` not `login`.

`login` is a noun.  `log in` is the verb form.

src/app/core/containers/AppContainer.js
```diff
@@ -50,6 +55,21 @@ const AppContainer = props => {
...
+        }
+      }
+    }, 1000)
```
brennancheung: We can probably up this interval.

jruddell:
Why are we working on this currently? Its not related to the priorities given from product. We need to stay focused on the items that need immediate fixes

.eslintrc
```diff
@@ -42,8 +42,10 @@
     "no-multi-spaces": [2, { "exceptions": { "ObjectExpression": true } } ],
     "key-spacing": [2, { "mode": "minimum" }],
     "standard/no-callback-literal": 0,
+    "space-before-function-paren": "off",
```
brennancheung: Do not disable this one.  It is used to differentiate between function definition and function application.

src/api-client/Qbert.ts
jruddell: I don't think we need to write out types if they are just generic like this. Not really much of a value add.. more to read in the code ... etc.

.eslintrc
jruddell: Lets not change lint standards unless we all talk about it and all agree on it.

src/api-client/Cinder.ts
```diff
@@ -109,115 +105,115 @@ class Cinder {
...
-    const { extra_specs, ...rest } = params
+    const { extra_specs: extraSpecs, ...rest } = params
     const baseResponse = await axios.put(url, { volume_type: rest }, this.client.getAuthHeaders())
-    await axios.post(`${url}/extra_specs`, { extra_specs }, this.client.getAuthHeaders())
-    keysToDelete.forEach(async key => {
-      await axios.delete(`${url}/extra_specs/${key}`, this.client.getAuthHeaders())
-    })
+    await axios.post(`${url}/extra_specs`, { extra_specs: extraSpecs }, this.client.getAuthHeaders())
+    await Promise.all(keysToDelete.map(async key => {
+      return axios.delete(`${url}/extra_specs/${key}`, this.client.getAuthHeaders())
```
brennancheung: Do we need the `async key` and the `return`?  I think `axios.delete` is already a Promise.  So you can probably just `map(key => axios.delete(...))`

xantorres: Tslint was complaining about it so I had to add it

src/app/core/containers/AppContainer.js
xantorres: Why? It doesn't hace any performance impact

src/api-client/Qbert.ts
xantorres: I was forced to add these types because tslint or typescript was complaining

src/api-client/Cinder.ts
xantorres: This doesn't break anything, javascript is smart enough to inmediately resolve if the function is not async.
And about the double return type, I had to do it because some services are fetching the url from the lubernetes api whereas others are returning the value directly as a hardcoded value.

.eslintrc
xantorres: I agree on this one, but the promises one was preventing me to call an async function from within a hook basically, so i have been forced to do it.

src/app/core/containers/AppContainer.js
brennancheung: I guess it's fine.  Just seems overkill.

Probably safer to remove the remove the interval when you destroy the session to avoid weird scenarios like toast showing multiple times.

xantorres: I refactored it to just initiate the interval when session exists

src/api-client/Cinder.ts
xantorres: Also in this case, we are really returning a promise, even if the function is not async, `getServiceEndpoint` is, so a promise is returned. Unfortunately I could not make this function async because TSLINT again was complaining about having an `async` function without an explicity `await` statement

PR #567, opened 2019-12-11
PMK-2411 deauth node

brennancheung:
This adds a node deauthorization option to the node list table.

Data updating the view and AppContext doesn't seem to be working quite right but I think that is either a separate problem with the data loader or perhaps it might need to be done slightly differently.  @xantorres?


src/app/plugins/kubernetes/components/infrastructure/nodes/actions.js
```diff
@@ -49,3 +50,12 @@ export const loadNodes = createContextLoader(nodesCacheKey, async (params, loadF
   refetchCascade: true,
   uniqueIdentifier: 'uuid',
 })
+
+export const deAuthNode = createContextUpdater('nodes', async (node, prevItems, loadFromContext) => {
+  // TODO: This doesn't seem to be updating the node list table for some reason
+  await resmgr.unauthorizeHost(node.uuid)
+  const newItems = prevItems.filter(_node => node.uuid !== _node.uuid)
```
xantorres: if this is a delete operation, you can just pass an option: `operation: "delete"` so that you won't need to update the prevItems manually

xantorres:
Fixed issues with AppProvider, can you test your code after rebasing @brennancheung?

src/app/plugins/kubernetes/components/infrastructure/nodes/actions.js
brennancheung: It's a de-authorize, not a delete.  I'm not understanding all the magic stuff that is happening but I think whatever is fetching the API calls right afterwards is probably the correct thing to do.  I'd prefer to be doing it explicitly so there is more control and less magic.  But for now I think it can work.  This shifts the state management to the backend where it's actually the authoritative source.

src/app/core/components/ConfirmationDialog.tsx
```diff
@@ -17,10 +24,10 @@ class ConfirmationDialog extends React.PureComponent {
   render () {
     const {
       open,
-      title,
-      text,
-      cancelText,
-      confirmText,
+      title = 'Are you sure?',
```
jruddell: 😍

src/app/plugins/kubernetes/components/infrastructure/nodes/NodesListPage.js
```diff
@@ -73,7 +75,15 @@ export const options = {
   name: 'Nodes',
   title: 'Nodes',
   uniqueIdentifier: 'uuid',
-  showCheckboxes: false
+  multiSelection: false,
+  batchActions: [
+    {
+      cond: ([node]) => !node.clusterUuid,
```
jruddell: we have the utility method for this `isUnassignedNode`

brennancheung: It's in the cluster chooser component though.  Probably need to save for a bigger refactor where we can move it to a more common location.

PR #568, opened 2019-12-11
Monitor cache changes

xantorres:
This modifies the `useDataLoader` hook to monitor changes in its current loader cache array, so when the cache its updated it will automatically reload without the need of manually refreshing the data using the `reload` method provided by the hook.

### DIFF
--- a/src/app/core/hooks/useDataLoader.js
+++ b/src/app/core/hooks/useDataLoader.js
@@ -1,10 +1,11 @@
 import { useMemo, useEffect, useCallback, useContext, useRef, useReducer } from 'react'
 import moize from 'moize'
 import { emptyArr, emptyObj } from 'utils/fp'
-import { isEmpty } from 'ramda'
+import { isEmpty, path } from 'ramda'
 import { useToast } from 'core/providers/ToastProvider'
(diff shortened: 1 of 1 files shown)


xantorres:
This change refreshes the data returned by the hook, refetching from the cache, so no, we are not re-performing any api request with the risk of overwriting the cache with this. 
The reasoning behind this is to make sure all current active data loader hooks are kept up to date with the cache, as we could end up with outdated data in some components after performing an update or insert, as there's not always the possibility of calling the `reload` hook method, depending from where we realised the update or insert.

jruddell:
Have we seen bugs here that this is trying to fix? What is the catalyst? I think both Brennan and I are concerned with making changes to this stuff right before a release (2 days out). We are essentially **code freeze** aside from bug fixes / approved changes. Please refrain from changing stuff like this without running it by us first.

xantorres:
This is definitely a bug fix. I have seen problems with the user lists not being refreshed after a cache update (delete) can't remember the specific use case now but anyway, we can leave this PR as a Draft for now until I can find further confirmation of the issues that I had experienced.

PR #570, opened 2019-12-11
Reimplement health status to not use numMasters and numWorkers anymore

murilovmachado:
This PR reimplements the way `Health Status` is calculated so that we don't use `numMasters` and `numWorkers` anymore. We had to do this because BareOS clusters don't have these properties filled.

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
@@ -99,8 +99,8 @@ const renderErrorStatus = (taskError, nodesDetailsUrl) =>
     <SimpleLink src={nodesDetailsUrl}>Error</SimpleLink>
   </ClusterStatusSpan>
 
-const renderClusterHealthStatus = (nodes, masterNodes, workerNodes, healthyMasterNodes, healthyWorkerNodes, taskError, nodesDetailsUrl) => {
-  const [healthStatus, message] = getHealthStatusAndMessage(nodes, masterNodes, workerNodes, healthyMasterNodes, healthyWorkerNodes)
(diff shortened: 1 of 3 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
```diff
@@ -99,8 +99,8 @@ const renderErrorStatus = (taskError, nodesDetailsUrl) =>
     <SimpleLink src={nodesDetailsUrl}>Error</SimpleLink>
   </ClusterStatusSpan>
 
-const renderClusterHealthStatus = (healthyMasterNodes, nodes, numMasters, numWorkers, taskError, nodesDetailsUrl) => {
-  const [healthStatus, message] = getHealthStatusAndMessage(healthyMasterNodes, nodes, numMasters, numWorkers)
+const renderClusterHealthStatus = (nodes, masterNodes, workerNodes, healthyMasterNodes, healthyWorkerNodes, taskError, nodesDetailsUrl) => {
```
brennancheung: Use an object here.  This is too many for positional arguments.

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -183,8 +183,9 @@ export const clusterActions = createCRUDActions(clustersCacheKey, {
         grafanaLink: hasPrometheusEnabled(cluster) ? grafanaLink : null,
       }
       const masterNodes = nodesInCluster.filter(node => node.isMaster === 1)
-      const healthyMasterNodes = masterNodes.filter(
-        node => node.status === 'ok' && node.api_responding === 1)
+      const workerNodes = nodesInCluster.filter(node => node.isMaster === 0)
```
brennancheung: Ramda has the `partition` function that should allow you to pull out master and workers at the same time.

murilovmachado:
> Will this still work for AWS and Azure?

Yep, it works for all cluster types

PR #571, opened 2019-12-11
[PMK-2060] New dashboard styling

kevinylee1991:
Screen Shot 2019-12-11 at 3 20 58 PM

Screen Shot 2019-12-11 at 3 12 19 PM

Old vs new. Madhura noted that it needs more color, but that this is a good starting point to check in. Icons need to be updated according to spec, but plan to do that in a different PR.

### DIFF
--- a/src/app/core/components/buttons/CardButton.js
+++ b/src/app/core/components/buttons/CardButton.js
@@ -12,6 +12,7 @@ const styles = theme => ({
     color: theme.palette.dashboardCard.primary,
     border: `1px solid ${theme.palette.dashboardCard.primary}`,
     background: theme.palette.dashboardCard.button,
+    whiteSpace: 'nowrap',
   }
 })
 
--- a/src/app/core/components/buttons/CardButton.js
+++ b/src/app/core/components/buttons/CardButton.js
@@ -1,0 +1,28 @@
+import React from 'react'
+import Button from '@material-ui/core/Button'
+import { withStyles } from '@material-ui/styles'
+
+const styles = theme => ({
+  cardButton: {
+    margin: 0,
+    borderRadius: 2,
+    textTransform: 'uppercase',
+    height: 32,
+    padding: theme.spacing(0.25, 2),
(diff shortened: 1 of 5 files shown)


kevinylee1991:
Screen Shot 2019-12-11 at 4 18 44 PM

src/app/plugins/kubernetes/components/dashboard/StatusCard.tsx
```diff
@@ -35,101 +39,101 @@ const useStyles = makeStyles((theme: any) => ({
...
-  pendingText: {
-    color: '#cb9f53',
-    marginLeft: theme.spacing(2),
+  arrowIcon: {
+    background: theme.palette.dashboardCard.primary,
+    height: '32px',
+    width: '32px',
+    color: theme.palette.primary.contrastText,
+    borderRadius: '50%',
+    fontSize: '18px',
```
brennancheung: Hmm, should we be changing the font size in CSS?  Font Awesome has all the way from 2x to 5x I think.

src/app/core/components/buttons/CardButton.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import Button from '@material-ui/core/Button'
+import { withStyles } from '@material-ui/styles'
+
+const styles = theme => ({
+  cardButton: {
+    margin: 0,
+    borderRadius: 2,
+    textTransform: 'uppercase',
```
brennancheung: I actually prefer the uppercase because this is what everyone expects with Material Design buttons, but Madhura had us change it in other places to not use all caps.  We should make everything all caps for buttons or make this consistent.  We should discuss with product.

src/app/plugins/kubernetes/components/dashboard/StatusCard.tsx
kevinylee1991: In zeplin they're defined in font sizes so I feel like using that might be easier to match with the spec. If they are consistent across the UI we could probably stick the icon font sizes in the theme.

PR #574, opened 2019-12-12
Fix invalidate cache and issues with context updating

xantorres:


src/app/core/providers/AppProvider.tsx
```diff
@@ -58,25 +59,58 @@ const initialContext: IAppContext = {
...
+  'initSession',
+  'registerNotification',
+  'clearNotifications',
+  'updateSession',
+  'destroySession',
+  'getContext',
+  'setContext',
+]
+
+const WindowObject = (window as any)
```
jruddell:
```
interface IWindow {
  context: OurContextType
}
```
one way you could use it would be

```
declare var window: IWindow
```

xantorres: yes, there are many things to improve this, this was just a quick workaround

PR #575, opened 2019-12-12
PMK-2352: Added functionality to change the password.

rahulkurumkar:
POMK-2352: Added functionality to change the password.

PMK-2352-No-cancel-button-for-reset-password-page


src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,230 @@
...
+  hasOneUpperChar,
+  hasOneLowerChar,
+  hasMinLength,
+  requiredValidator,
+  passwordValidator,
+} from 'core/utils/fieldValidators'
+
+const styles = (theme) => ({
+  formContainer: {
+    margin: '0 37px',
```
brennancheung: Don't use hard coded values for spacing.  Use `theme.spacing`.  Also try to keep it a nice integer multiplier instead of weird fractional values.  This will make things flow visually and be more consistent.

Width and height are fine to hard-code though.  Just not margin and padding.

src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,230 @@
...
+    setContext,
+    userDetails: { userId },
+  } = useContext(AppContext)
+  const { keystone } = ApiClient.getInstance()
+
+  const handleCancel = () => setContext({ isChangePassword: false })
+
+  const handleFormSubmit = async (data) => {
+    if (params.isChangePasswordSuccessful) {
+      await setContext({ isChangePassword: false })
```
brennancheung: Don't repurpose the same exact page for multiple things and don't use weird global variables like this.  It introduces complexity.  Abstract out the common parts and create separate pages for change password and login.

src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,230 @@
...
+
+    try {
+      const body = { original_password: params.oldPassword, password: params.confirmPassword }
+      const updateUserPromise = keystone.updateUserPassword(userId, body)
+      await updateUserPromise.then((res) => (isNil(res) ? reject(isNil) : undefined))
+      await updateParams({ isChangePasswordSuccessful: true, loading: false })
+      await setContext({ isChangePassword: false })
+      history.push({
+        pathname: loginUrl,
+        toastData: {
```
brennancheung: Toasts and navigation are separate.  Don't complect them like this.

src/app/core/public/LoginPage.js
```diff
@@ -168,30 +169,52 @@ export class LoginPage extends React.PureComponent {
...
+              },
+            ]}
+          />
+        )}
         <Grid container justify="center" className={classes.root}>
           <Grid item md={4} lg={3}>
             <Paper className={classes.paper}>
-              <img alt="Platform 9" src={pathJoin(imageUrlRoot, 'logo-color.png')} className={classes.img} />
+              <img
+                alt="Platform 9"
```
brennancheung: No space in `Platform9`

src/app/core/components/ChangePasswordPage.js
rahulkurumkar: Addressed

src/app/core/public/LoginPage.js
```diff
@@ -168,30 +169,52 @@ export class LoginPage extends React.PureComponent {
...
     return (
       <div className="login-page">
+        {toastData && (
+          <ToastContainer
+            toasts={[
+              {
+                id: toastData.id,
+                isOpen: toastData.isOpen,
+                variant: toastData.variant,
+                text: toastData.text,
```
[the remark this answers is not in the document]
rahulkurumkar: Addressed

src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,230 @@
...
+
+    if (!isPasswordMatches()) {
+      updateParams({
+        isError: true,
+        errorMessage: 'Passwords do not match',
+      })
+      return
+    }
+
+    if (isNil(params.oldPassword)) {
```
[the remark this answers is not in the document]
rahulkurumkar: Addressed

src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,230 @@
...
+
+  const handleCancel = () => setContext({ isChangePassword: false })
+
+  const handleFormSubmit = async (data) => {
+    if (params.isChangePasswordSuccessful) {
+      await setContext({ isChangePassword: false })
+      history.push(loginUrl)
+    }
+
+    if (!isPasswordMatches()) {
```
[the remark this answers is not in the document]
rahulkurumkar: Addressed

src/app/core/components/ChangePasswordPage.js
rahulkurumkar: Addressed

src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,213 @@
...
+
+const styles = () => ({
+  alertContainer: {
+    width: '300px',
+  },
+})
+
+const ChangePasswordPage = (props) => {
+  const { params, updateParams } = useParams({
+    loading: false,
```
xantorres: Please declare the initial params outside of the render body so we prevent this object being created every time

src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,213 @@
...
+    isError: false,
+    newPassword: '',
+    confirmPassword: '',
+    isChangePasswordSuccessful: false,
+    errorMessage: 'Change password failed',
+  })
+
+  const passwordValidatorList = [
```
xantorres: Declare this object outside of the render body

src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,213 @@
...
+  } = useContext(AppContext)
+  const showToast = useToast()
+  const { keystone } = ApiClient.getInstance()
+
+  const handleCancel = () => setContext({ isChangePassword: false })
+
+  const handleFormSubmit = async (data) => {
+    if (params.isChangePasswordSuccessful) {
+      await setContext({ isChangePassword: false })
+      history.push(loginUrl)
```
xantorres: Please destroy the session before redirecting to the login page (`destroySession` from AppContext). 
Or maybe even better, just redirect to the logout page which already contains the destroy session logic and redirection to the login page.

src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    await updateParams({
```
xantorres: this method doesn't return a promise

src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,213 @@
...
+      loading: true,
+    })
+
+    try {
+      const body = { original_password: params.oldPassword, password: params.confirmPassword }
+      const updateUserPromise = keystone.updateUserPassword(userId, body)
+      await updateUserPromise.then((res) => (isNil(res) ? reject(isNil) : undefined))
```
xantorres: We should just capture the error in the catch statement and display the specific error to the user, instead of ignoring. Even if you don't want to display the server error for whatever reason, it should be handled in the catch statement, handling it here makes things confusing.

src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,213 @@
...
+      await updateParams({ isChangePasswordSuccessful: true, loading: false })
+      await setContext({ isChangePassword: false })
+      showToast('Successfully updated password. Please log in with new password.', 'success')
+      history.push(loginUrl)
```
xantorres: Redirect to the logout page instead to make sure the session is cleaned up.

src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,213 @@
...
+    } catch (err) {
+      updateParams({ isError: true, loading: false })
+    }
+  }
+
+  const updateValue = (key) => (value) => {
```
xantorres: You don't need to do this, just use the `getParamsUpdater` function provided by the `useParams` hook.

src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,213 @@
...
+
+  const CheckListItem = ({ children, checked }) => (
+    <ListItem>
+      <ListItemIcon>{checked ? <CheckIcon /> : <ClearIcon color="error" />}</ListItemIcon>
+      <ListItemText primary={children} />
+    </ListItem>
+  )
+
+  const renderPasswordValidationCheck = (values) => (
+    <Typography variant="body1" component="div">
```
xantorres: Please define this function outside of the render body.

src/app/core/components/ChangePasswordPage.js
```diff
@@ -0,0 +1,213 @@
...
+              </ValidatedForm>
+            </Grid>
+          </Grid>
+        </Progress>
+      </DialogContent>
+    </Dialog>
+  )
+}
+
+export default withStyles(styles)(ChangePasswordPage)
```
xantorres: Please follow the same pattern we are using in all the functional components and use the `useStyles` hook instead of this HoC

src/app/core/components/UserMenu.js
```diff
@@ -27,9 +27,13 @@ class UserMenu extends React.PureComponent {
   state = { anchorEl: null }
   handleClick = event => this.setState({ anchorEl: event.currentTarget })
   handleClose = () => this.setState({ anchorEl: null })
-  handleChangePassword = () => this.props.history.push(resetPasswordUrl)
   logout = () => this.props.history.push(logoutUrl)
 
+  handleChangePassword = () => {
+    this.props.setContext({ isChangePassword: true })
```
xantorres: Don't use the global context for this sort of things, just render the `ChangePasswordPage` modal (which, by the way, should be called `ChangePasswordModal`) in this view. Keep things simple.

src/app/core/containers/AppContainer.js
```diff
@@ -95,7 +95,14 @@ const AppContainer = () => {
     const currUnscopedToken = tokens && tokens.unscopedToken
     if (username && currUnscopedToken) {
       // We need to make sure the token has not expired.
-      const { unscopedToken, expiresAt, issuedAt } = await keystone.renewToken(currUnscopedToken)
+      const response = await keystone.renewToken(currUnscopedToken)
+
+      if (isNil(response)) {
```
xantorres: Is this necessary? If so, please redirect to logout page instead of this.

src/app/core/components/ChangePasswordPage.js
rahulkurumkar: Done

src/app/core/components/ChangePasswordPage.js
rahulkurumkar: Done

src/app/core/components/ChangePasswordPage.js
rahulkurumkar: Done

src/app/core/components/ChangePasswordPage.js
rahulkurumkar: Done

src/app/core/components/ChangePasswordPage.js
rahulkurumkar: Done

src/app/core/components/ChangePasswordPage.js
rahulkurumkar: Done

src/app/core/components/ChangePasswordPage.js
rahulkurumkar: Done

src/app/core/containers/AppContainer.js
rahulkurumkar: Yes. I got the error and then I debugged and come to this conclusion.  I am not sure when it happens but it was having some corner case.

src/app/core/components/ChangePasswordPage.js
rahulkurumkar: Done

src/app/core/components/ChangePasswordModal.js
```diff
@@ -0,0 +1,198 @@
...
+  </ListItem>
+)
+
+const renderPasswordValidationCheck = (values) => (
+  <Typography variant="body1" component="div">
+    Password must contain the following:
+    <List dense>
+      {passwordValidatorList.map((record) => (
+        <CheckListItem
+          key={uuid()}
```
jruddell: we shouldn't be using `uuid`'s as keys to react components. This will cause weird unexpected side effects. The `key` should always be related to the item its being rendered with. In this instance you could either define the uuid as the key in `passwordValidatorList` since that would create them one time and wouldn't change. Or you could also use the `displayText` as the key since thats unique also.

src/app/core/components/ChangePasswordModal.js
```diff
@@ -0,0 +1,198 @@
...
+      await setContext({ isChangePassword: false })
+      history.push(logoutUrl)
+    }
+
+    updateParams({
+      loading: true,
+    })
+
+    try {
+      const body = { original_password: params.oldPassword, password: params.confirmPassword }
```
jruddell: this should probably be made into an action that you just call here. Let the action handle this logic so we can potentially reuse it / move more complicated requesting logic outside of the view component.

src/app/core/components/ChangePasswordModal.js
```diff
@@ -0,0 +1,198 @@
...
+    </List>
+  </Typography>
+)
+
+const ChangePasswordModal = (props) => {
+  const { params, updateParams, getParamsUpdater } = useParams(initialState)
+  const classes = useStyles()
+  const { history } = useReactRouter()
+  const passwordValidators = [requiredValidator, passwordValidator]
+  const confirmPasswordValidator = [
```
xantorres: You should declare this outside of the render method

src/app/core/containers/AuthenticatedContainer.js
```diff
@@ -158,6 +159,7 @@ const AuthenticatedContainer = () => {
           [classes.contentShift]: drawerOpen,
           [classes['contentShift-left']]: drawerOpen,
         })}>
+          {isChangePassword && <ChangePasswordModal />}
```
xantorres: Just put this modal in the `UserMenu` component so we don't have to use weird global variables

src/app/core/components/UserMenu.js
```diff
@@ -24,17 +25,27 @@ const styles = theme => ({
...
 class UserMenu extends React.PureComponent {
-  state = { anchorEl: null }
+  state = { anchorEl: null, showChangePasswordModal: false }
   handleClick = event => this.setState({ anchorEl: event.currentTarget })
   handleClose = () => this.setState({ anchorEl: null })
-  handleChangePassword = () => this.props.history.push(resetPasswordUrl)
   logout = () => this.props.history.push(logoutUrl)
 
+  handleChangePassword = () => {
+    this.setState({ showChangePasswordModal: true })
```
xantorres: Just call `setState` once
`this.setState({ showChangePasswordModal: true, anchorEl: null })` so you prevent an extra render

rahulkurumkar: Addressed

src/api-client/Keystone.ts
```diff
@@ -363,6 +363,13 @@ class Keystone extends ApiService {
     return data.user
   }
 
+  updateUserPassword = async (id, params) => {
```
jruddell: We already have an update for the user in keystone that does a `basicPatch`. This is used to update the users password in the edit user view inside of the UsersListPage. Do we need a specific API for updating the password?

rahulkurumkar: existing `updateUser` do not check if the `currentPassword` is entered valid or not `updateUserPssword` is doing it.

PR #576, opened 2019-12-12
Cluster nodes details page

murilovmachado:
This PR implements Master Nodes and Worker Nodes tables.

Screen Shot 2019-12-12 at 19 28 38

### DIFF
--- a/src/app/core/helpers/createListTableComponent.js
+++ b/src/app/core/helpers/createListTableComponent.js
@@ -2,7 +2,6 @@ import React from 'react'
 import { withRouter } from 'react-router-dom'
 import ListTable from 'core/components/listTable/ListTable'
 import { withScopedPreferences } from 'core/providers/PreferencesProvider'
-import requiresAuthentication from 'openstack/util/requiresAuthentication'
 import { compose } from 'ramda'
 
 const createListTableComponent = ({
@@ -17,6 +16,7 @@ const createListTableComponent = ({
   showCheckboxes,
   onReload,
   onRefresh,
+  compactTable = false,
 }) => {
   const CustomListTable = ({
     data,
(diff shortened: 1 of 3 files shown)


src/app/plugins/kubernetes/components/infrastructure/nodes/NodesListPage.js
```diff
@@ -50,10 +54,54 @@ const renderNodeDetailLink = (name, node) =>
...
+
+const renderAPIServer = (_, { isMaster, api_responding: apiResponding }) => !!isMaster && (!!apiResponding).toString()
+
 export const columns = [
   { id: 'uuid', label: 'UUID', display: false },
   { id: 'name', label: 'Name', render: renderNodeDetailLink },
-  { id: 'status', label: 'Status', render: renderStatus },
+  { id: 'connectionStatus', label: 'Connection status', render: renderConnectionStatus },
+  { id: 'healthStatus', label: 'Health status', render: renderHealthStatus },
+  { id: 'api_responding', label: 'API server', render: renderAPIServer },
```
brennancheung: Go with `renderApiServer` to be more consistent with out existing naming conventions.

src/app/core/helpers/createListTableComponent.js
```diff
@@ -17,6 +16,7 @@ const createListTableComponent = ({
   showCheckboxes,
   onReload,
   onRefresh,
+  compactTable = false,
```
brennancheung: We want to eventually refactor `ListTable` to split it up and not have so many props.  Ideally we should be removing props from it not adding more.  If you can find an easy way to simplify this go with that, but if not, this is fine for now until we can refactor `ListTable` later on.

PR #577, opened 2019-12-12
Nodes remote support initial

kevinylee1991:
What do you guys think about this pattern of invalidating the caches of loaders when used within other loaders? This change ensures that when you press refresh on the nodes page, that an API request is made to resmgr and nodes APIs. Seems like a workaround that fights this loader/updater architecture but it's a solution for this node update issue.


brennancheung:
> What do you guys think about this pattern of invalidating the caches of loaders when used within other loaders?

It makes a certain amount of sense but I think we're going to refactor the data loading / context system to redux, remove a lot of the magic, and make it so actions need to be much more explicit about updating the cache and refreshing things.

src/api-client/ResMgr.ts
```diff
@@ -51,6 +51,14 @@ class ResMgr extends ApiService {
     return response.data
   }
 
+  async pushRole (hostId, role, body) {
```
brennancheung: Not sure `pushRole` is the correct term.  I prefer `addRole` or `putRole` unless there is a more specific reason for the this naming.

xantorres:
Close the other PR if you think it is not valid anymore

PR #579, opened 2019-12-13
Partial work on getting integration tests working again

brennancheung:
A lot of things have changed since we last worked on our integration tests.  Some things that still need to be done:

- [x] Session format in localStorage has changed so the mocks aren't allowing us to start from a logged in state.
- [x] Simulator does not send the expiration information need to keep a session open for a longer period of time

The following will be done in a separate PR:

- [ ] Cypress needs to be configured with the dashboard
- [ ] Figure out how to get tests to run in parallel
- [ ] Kick off test builds on every PR commit
- [ ] Figure out how to integrate cypress test results to allow / deny the PR to be merged.


xantorres:
Fixed stored session tests and added a login command.
Also made changes to keep the session alive even after reloading the server.

xantorres:
Added `--parallel` parameter to parallelize integration tests when running `cypress:record` script

brennancheung:
After a bit of research with GitLab looks like it won't work for us.  It forces us to run everything in their repo instead from GitHub.  They supposedly have sync capability in the paid version but it's still not ideal.  I'd rather we run directly from GitHub.

This PR has been open for a while so let's merge it in it's current state and we can add more tests in the future.

I will work on automating the CI/CD in another future PR.

Some of the Cypress tests are still failing but they are a lot closer to work than we had before.  Since we are not running Cypress in our pipeline yet I think it is safe to merge the improved functionality for now.

cypress/integration/clusters.spec.js
```diff
@@ -17,54 +17,55 @@ describe('clusters', () => {
...
     })
   })
 
   context('create cluster', () => {
     it('shows the cluster create form', () => {
-      cy.visit('/ui/kubernetes/clusters')
-      cy.contains('Add').click()
-      cy.contains('Add Cluster')
+      cy.visit('/ui/kubernetes/infrastructure')
+      cy.contains('+ Add Cluster').click()
```
jruddell: if any of the content / labels change this would break these tests. Isn't there a better way to click a button with cypress?

PR #581, opened 2019-12-13
Fix outlined inputs height and font size

xantorres:
Outlined inputs height were too tall and fonts too big and therefore looking inconsistent with other elements of the UI

Before:
image

After:
image

### DIFF
--- a/src/app/core/themes/defaultTheme.js
+++ b/src/app/core/themes/defaultTheme.js
@@ -52,11 +52,11 @@ const theme = {
     },
     MuiOutlinedInput: {
       root: {
-        fontSize: '16px',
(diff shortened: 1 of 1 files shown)


jruddell:
I think we can make the search bar the same size as the other fields (assuming thats what you were trying to fix with this pr). but I dont think we should be throwing out font-sizes all over the place. Good to keep it consistent.

xantorres:
The inputs are looking too big, in my opinion, also most base fonts are around 12-14px size, not 16, so the dropdowns look way too prominent compared to the rest of the UI. 
And multiples of 8 should be used just for paddings and margins, not heights, widths or font sizes.

PR #599, opened 2019-12-19
PMK-2454: Added ‘Copy to clipboard’ functionality

rahulkurumkar:
PMK-2454: Added ‘Copy to clipboard’ functionality
PMK-2451: New user creation flow minor updates (New user radio color is red - should be blue )

PMK-2454-Add-copy-to-clipboard-functionality-for-onboarding-a-node-CLI-instructions

PMK-2454-Add-copy-to-clipboard-functionality-for-onboarding-a-node-CLI-instructions

### DIFF
--- a/src/app/core/components/CopyToClipboard.tsx
+++ b/src/app/core/components/CopyToClipboard.tsx
@@ -1,0 +1,70 @@
+import React, { FunctionComponent, useRef } from 'react'
+import { FileCopy, Done } from '@material-ui/icons'
+import IconButton from '@material-ui/core/IconButton'
+import { makeStyles } from '@material-ui/styles'
+import useParams from 'core/hooks/useParams'
+
+interface Props {
+  copyText: string
+  children: any
+}
+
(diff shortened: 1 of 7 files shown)


src/app/core/components/CopyToClipboard.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+const defaultParams: State = {
+  showCopyIcon: false,
+  isCopySuccessful: false,
+}
+
+const useStyles = makeStyles(() => ({
+  textArea: {
+    position: 'absolute',
+    top: '-9999px',
```
jruddell: you can also just do `visibility: hidden; pointerEvents: none'` with absolute positioning.

rahulkurumkar: @jruddell , When we add `visibility: hidden`  in that case it is not able to copy the text. The element should not be hidden because we select the text first and then trigger a copy command.

PR #600, opened 2019-12-19
PMK-2439 Change success update operations messages

xantorres:
* Changing default successful update/delete/create operation message to a more readable "XXX created successfully" instead of "Successfully created XXX"
* Customized success messages for update/delete/create operation messages of users/tenants to display the name of the updated user or tenant.

### DIFF
--- a/src/app/plugins/kubernetes/components/userManagement/tenants/actions.js
+++ b/src/app/plugins/kubernetes/components/userManagement/tenants/actions.js
@@ -1,8 +1,10 @@
 import ApiClient from 'api-client/ApiClient'
-import { keys, pluck, pipe, prop, find, propEq, filter, always, isNil, reject, map } from 'ramda'
+import {
+  keys, pluck, pipe, prop, find, propEq, filter, always, isNil, reject, map, last,
+} from 'ramda'
 import { namespacesCacheKey } from 'k8s/components/namespaces/actions'
 import createCRUDActions from 'core/helpers/createCRUDActions'
-import { filterIf, pathStr, emptyArr } from 'utils/fp'
+import { filterIf, pathStr, emptyArr, objSwitchCase } from 'utils/fp'
 import createContextLoader from 'core/helpers/createContextLoader'
 import { tryCatchAsync } from 'utils/async'
 import { mngmUsersCacheKey } from 'k8s/components/userManagement/users/actions'
@@ -119,6 +121,17 @@ export const mngmTenantActions = createCRUDActions(mngmTenantsCacheKey, {
   refetchCascade: true,
   requiredRoles: 'admin',
   entityName: 'Tenant',
+  successMessage: (updatedItems, prevItems, { id }, operation) => objSwitchCase({
+    create: `Tenant ${prop('name', last(updatedItems))} created successfully`,
+    update: `Tenant ${pipe(
+      find(propEq('id', id)),
+      prop('name'),
+    )(prevItems)} updated successfully`,
+    delete: `Tenant ${pipe(
+      find(propEq('id', id)),
+      prop('name'),
+    )(prevItems)} deleted successfully`,
+  })(operation),
 })
 
 export const mngmTenantRoleAssignmentsCacheKey = 'managementTenantRoleAssignments'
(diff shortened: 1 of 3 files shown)


src/app/plugins/kubernetes/components/userManagement/tenants/actions.js
```diff
@@ -119,6 +121,17 @@ export const mngmTenantActions = createCRUDActions(mngmTenantsCacheKey, {
   refetchCascade: true,
   requiredRoles: 'admin',
   entityName: 'Tenant',
+  successMessage: (updatedItems, prevItems, { id }, operation) => objSwitchCase({
+    create: `Tenant ${prop('name', last(updatedItems))} created successfully`,
```
jruddell: updatedItems is the whole list of tenants if im not mistaken. @brennancheung had an issue where the wrong name was displayed. Isn't the 3rd parameter the entity in question? why iterate the list when you can just reference the name from the 3rd param?

xantorres: 3d param is the `params` object provided when calling the action, which typically should just contain the `id` or the required params to perform the update/delete/create operation.

jruddell: Why would we design it like that? having to find an element in an array when you are editing / mutating one entity. Why not just expose that entity here and reduce complexity / operations? Every place we use this we have to do extra work just to get the entity in question.

xantorres: Good point, this was added at a later point to handle the error/success messages after performing CRUD operations, and as such it was adapted to work with what it's available in the scope. I designed it to be as flexible as possible, that is, in some cases you would want to show the name of the deleted item, and to do so you have to look in to the previous items array, because the updated array doesn't contain the item anymore, whereas in other cases you want to name of the updated/added item.
The thing is that all CRUD actions (or other custom operations) are different and sometimes you don't have the entity you are editing or the fields you need because each action casuistry differs from each other. 
Regarding the extra performance hit, this really should not be a problem at all, as this will just be executed once after the action I performed, and right now the number of items we will work with is negligible. Is for that reasons too that we don't even have a pagination system in place, simply because we don't (and probably we ever won't) need it.

xantorres: In any case, there are already default messages defined in the helper, but sometimes like here they want to specifically display one field (the `name` in this case), so i added this additional option to allow customising it.

jruddell: I'm not concerned about performance, but rather excessive complexity because of a bad architecture. its fine for now, I'll end up changing this when setting up redux

xantorres: Redux has little to do with this architecture, and even if you dislike this, is aimed to solve some very specific requirements of data caching and indexing, while exposing an arguably simple yet flexible api to work with. 
It obviously has its flaws but its working fine so far, except for some edge cases with multiple layer of entities related, which greatly increases the complexity and would require a more suited solution like a full fledged ORM, thing that is already been discarded for being too much of an overkill.

xantorres: What I agree with though, is that Redux will allow some simplifications in all this process (as I already could do in my PR) as it basically allows to use the store from any function  at any time, whereas in the current implementation I had to make some sacrifices and over complications to keep the context in sync.

brennancheung: There are 2 points to the wrong name issue:

1) The success message function arguments was confusing and the first argument returns all the items while the 3rd argument returns the actual item that was changed.  So I mistakenly was not getting the right entity.

2) The string `Management Tenant` was almost universally assumed to be the name of an actual tenant and not the type of tenant.  We should just be referring them to them as `tenants` and that's it.

xantorres:
1) The third argument is the `param` object send to the action, not the current item (we don't have that because we could be editing many items in a custom operations, or no item at all if deleting)
2) The string `Management Tenant" comes from the fact that this name is automatically generated from the cache key name if no entity name is provided , which in that case was `managementTenants` so the generated entity name was "Management Tenant", this is what I solved by explicitly passing an `entityName` option

PR #601, opened 2019-12-19
Add an error control for when the server doesn't return a list of projects (tenants)

xantorres:


xantorres:
Because for many api requests we need to have a tenant ID, whereas the user id is already known by the server from the moment we generate a token (as far as I know)

PR #604, opened 2019-12-20
Render table with empty text instead of text only

murilovmachado:
JIRA Card: https://platform9.atlassian.net/browse/PMK-2405

Screen Shot 2019-12-20 at 18 10 58

### DIFF
--- a/src/app/plugins/kubernetes/components/apiAccess/ApiAccessPage.js
+++ b/src/app/plugins/kubernetes/components/apiAccess/ApiAccessPage.js
@@ -1,16 +1,17 @@
 import React from 'react'
-import Tabs from 'core/components/tabs/Tabs'
-import Tab from 'core/components/tabs/Tab'
+// import Tabs from 'core/components/tabs/Tabs'
+// import Tab from 'core/components/tabs/Tab'
 import PageContainer from 'core/components/pageContainer/PageContainer'
 import ApiAccessListPage from './ApiAccessListPage'
 
 const ApiAccessPage = () => (
   <PageContainer>
-    <Tabs>
+    <ApiAccessListPage />
+    {/* <Tabs>
       <Tab value="apiAccess" label="API Access">
         <ApiAccessListPage />
       </Tab>
-    </Tabs>
+    </Tabs> */}
   </PageContainer>
 )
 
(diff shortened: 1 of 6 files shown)


murilovmachado:
> I'm not understanding what all this code is for? I don't want to see more props added to our ListTable logic when we should just not show the table at all and just put something to the effect of `No clusters are available. Please create a cluster first.`

@brennancheung  Actually we should show the table in this case. There's an attached image in the Jira Card showing a table with the empty message printed in the table.

jruddell:
- No content
image

- Loading State
image

- With content
image

src/app/plugins/kubernetes/components/apiAccess/ApiAccessPage.js
```diff
@@ -1,16 +1,17 @@
...
+// import Tabs from 'core/components/tabs/Tabs'
+// import Tab from 'core/components/tabs/Tab'
 import PageContainer from 'core/components/pageContainer/PageContainer'
 import ApiAccessListPage from './ApiAccessListPage'
 
 const ApiAccessPage = () => (
   <PageContainer>
-    <Tabs>
+    <ApiAccessListPage />
+    {/* <Tabs>
```
brennancheung: I believe we are making all pages have tabs for consistency of formatting on the page.  Indentation and the title at the top of the page will look different if we are not consistent.

jruddell: It looks completely stupid to have a single tab, and then a header `API Endpoints` right below that.

brennancheung: We need to change the design everywhere to support this then.  This was discussed and decided on in the past.

jruddell: @brennancheung no these are two different things. We dont provide a bad UX for consistency sake. Tabs are specific elements for a purpose. When that purpose is not in a page, then we shouldn't be using them. It looks bad, and I don't know of a single website that does something like this. The only place where we should `change the design` is where there is a page with one tab. Talked with @smadhura about it and she agrees that we should take it out.

PR #607, opened 2019-12-30
PMK-2426: Correct the display message.

rahulkurumkar:
Minor change. PMK-2426: Corrected the display message.

### DIFF
--- a/src/app/core/public/ResetPasswordPage.js
+++ b/src/app/core/public/ResetPasswordPage.js
@@ -180,12 +180,7 @@ const ResetPasswordPage = props => {
   )
 
   return (
-    <Progress
-      loading={params.loading}
-      overlay
-      renderContentOnMount
-      message="Processing..."
-    >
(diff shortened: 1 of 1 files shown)


src/app/core/public/ResetPasswordPage.js
```diff
@@ -231,18 +226,15 @@ const ResetPasswordPage = props => {
...
                         )}
                         <SubmitButton label="RESET MY PASSWORD" />
                       </>
                     ) : (
                       <>
                         <Typography className={classes.paragraph} component="p">
-                          Your was reset successfully.
+                          Your password has been reset successfully.
                         </Typography>
                         <SubmitButton label="RETURN TO LOGIN SCREEN" />
```
brennancheung: Don't use all caps here.  We will use CSS if we want this styling.

rahulkurumkar: Done. This component needs refactoring as lot of things are not done the correct way and I am the owner of it :)

PR #609, opened 2020-01-06
PMK-2485: Refactored the Reset password page also added show password and typescript feature

rahulkurumkar:
PMK-2485: Refactored the Reset password page also added show password and typescript feature


src/api-client/Clemency.ts
```diff
@@ -20,10 +20,8 @@ class Clemency extends ApiService {
   }
 
   resetPassword = async (secret, username, password) => {
-    const body = {
-      username, password,
-    }
-    return this.client.basicPost(`${this.baseUrl}/reset/password/${secret}`, body)
+    const body = { username, password }
+    return this.client.basicPost(`${this.baseUrl}/reset/password${secret}`, body)
```
jruddell: what is the format of `secret`? is this still a valid url?

src/app/core/public/ResetPasswordPage.tsx
```diff
@@ -0,0 +1,267 @@
...
+  confirmPassword: string
+  isResetPasswordSuccessful: boolean
+  errorMessage: string
+  isNewPasswordMasked: boolean
+  isConfirmPasswordMasked: boolean
+}
+
+interface Validator {
+  displayText: string
+  validator(): boolean
```
jruddell: its better to type this like `validator: () => boolean`

src/api-client/Clemency.ts
rahulkurumkar: It's `/unquiecode` e.g. `reset/password/5dbfcf2092b50d973dec177b6756a97aa11065829314fbf05362eb5bdb451c4dae7d0658fcb478f601be2870c7d6eac8f9fa11f3c24e29c26357451e1fc18014`

src/app/core/public/ResetPasswordPage.tsx
rahulkurumkar: Done.

src/app/core/containers/AppContainer.js
```diff
@@ -14,7 +14,7 @@ import {
   activateUserUrl, dashboardUrl, resetPasswordUrl, resetPasswordThroughEmailUrl,
   forgotPasswordUrl, loginUrl,
 } from 'app/constants'
-import ResetPasswordPage from 'core/public/ResetPasswordPage'
+import ResetPasswordPage from 'core/public/ResetPasswordPage.tsx'
```
kevinylee1991: adding .tsx shouldn't be necessary I think

src/app/core/public/ResetPasswordPage.tsx
```diff
@@ -0,0 +1,268 @@
...
+      history.push(loginUrl)
+    }
+
+    updateParams({
+      loading: true,
+    })
+
+    try {
+      // ? .slice(1) is used to remove the extra /
+      const secret = history.location?.pathname.replace(resetPasswordUrl, '')?.slice(1)
```
kevinylee1991: Not sure if the question marks here are needed -- does it error out without it?

src/app/core/containers/AppContainer.js
rahulkurumkar: It was throwing error `
Error: ENOENT: no such file or directory, open '/Volumes/Rahul/Source/pf9-ui-plugin/src/app/core/public/ResetPasswordPage.js'` but after a server restart, it is working. Removed the extension.

src/app/core/public/ResetPasswordPage.tsx
rahulkurumkar: I had added this for safety but we are already in try-catch also pathname should be present.

src/api-client/Clemency.ts
```diff
@@ -1,42 +1,45 @@
...
     return config.apiHost
   }
 
-  get baseUrl () {
+  get baseUrl (): string {
     return `${this.endpoint()}/clemency`
   }
 
-  requestNewPassword = async (username) => {
+  requestNewPassword = async (username: string): Promise<object> => {
```
jruddell: `Promise<object>` is not a very helpful typing. Do you know what the return values are here? if not then just do `Promise<any>` (when we actually strongly type these things we would search for `any` types to find instances that need updated.)

src/app/core/public/ResetPasswordPage.tsx
```diff
@@ -0,0 +1,268 @@
...
+    validator: hasOneNumber,
+  },
+  {
+    displayText: '1 Special character - !@#$%^&*()?',
+    validator: hasOneSpecialChar,
+  },
+]
+const { clemency } = ApiClient.getInstance()
+
+const CheckListItem = ({ children, checked }: { children: any, checked: boolean }): JSX.Element => (
```
jruddell: you should use `React.FC` which is short hand for `React.FunctionComponent`. This automatically adds `children` as a type. Return type doesn't need to be specified, typescript will automatically infer that it returns a `JSX.Element`

`const CheckListItem: FC<{checked: boolean}> = ({ children, checked } => (`

src/app/core/public/ResetPasswordPage.tsx
```diff
@@ -0,0 +1,268 @@
...
+  <ListItem>
+    <ListItemIcon>{checked ? <CheckIcon /> : <ClearIcon color="error" />}</ListItemIcon>
+    <ListItemText primary={children} />
+  </ListItem>
+)
+
+const renderPasswordValidationCheck = (values: string): JSX.Element => (
```
jruddell: `values` plural form reads like its an array. Probably just update that to be `value` or `passwordValue`

src/app/core/public/ResetPasswordPage.tsx
```diff
@@ -0,0 +1,268 @@
...
+      history.push(loginUrl)
+    }
+
+    updateParams({
+      loading: true,
+    })
+
+    try {
+      // ? .slice(1) is used to remove the extra /
+      const secret = history.location.pathname.replace(resetPasswordUrl, '').slice(1)
```
jruddell: This is generally a bad way to get a secret from a url. If the format changes in the future we will have to update every place we do things like this. Lets talk about a better way to do this. What is the current URL format? can we use `UrlSearchParams` here? or even `Url`? Something that parses the current url and gives specific keys to values would be much less error prone

src/app/core/public/ResetPasswordPage.tsx
rahulkurumkar: Addressed

src/api-client/Clemency.ts
rahulkurumkar: Not sure about the all response format. Changing it to `Promise<any>`

src/app/core/public/ResetPasswordPage.tsx
rahulkurumkar: Changed to `passwordValue`

src/app/core/public/ResetPasswordPage.tsx
rahulkurumkar: I tried with both `Url` and `UrlSeachParams` but I am unable to figure it out.
We get url as `https://pmkft-1578430215-06056.platform9.io/ui/reset/password/c26e4e46d3bcdccc127e6fedbe0b2d027716a75f19a4bc66affd0038bab04fcbaf8e6a9bed093618b1b8a636931e96ca481963f4fa7236e211ac25dac74997ea` out of which we get `pathname` as `/ui/reset/password/c26e4e46d3bcdccc127e6fedbe0b2d027716a75f19a4bc66affd0038bab04fcbaf8e6a9bed093618b1b8a636931e96ca481963f4fa7236e211ac25dac74997ea`
From the above `pathname` we have `/ui/reset/password` as `resetPasswordUrl` 
After replacing that we get `/c26e4e46d3bcdccc127e6fedbe0b2d027716a75f19a4bc66affd0038bab04fcbaf8e6a9bed093618b1b8a636931e96ca481963f4fa7236e211ac25dac74997ea` and from which we just need the code

jruddell: What does React Router give you? If this is a dynamic url with the last parameter as `/ui/reset/password/:token` then react router should have an object that reflects this in key value pairs. `'params': { token: 'c26e4e46d3bcdccc127e6fedbe0b2d027716a75f19a4bc66affd0038bab04fcbaf8e6a9bed093618b1b8a636931e96ca481963f4fa7236e211ac25dac74997ea'}`

Try inspecting what react router gives u

rahulkurumkar: Done

src/app/core/utils/fieldValidators.js
```diff
@@ -20,7 +20,7 @@ export const customValidator = (validator, errorMessage) =>
   new FieldValidator(validator, errorMessage)
 
 const fieldIsUnset = value => isNil(value) || isEmpty(value) || value === false
-export const hasMinLength = curry((minLen, value) => both(is(String), val => val.length >= minLen)(value))
+export const hasMinLength = minLen => value => both(is(String), val => val.length >= minLen)(value)
```
xantorres: why? Curried functions are more flexible

rahulkurumkar: Type for validator is `(value: string | number) => boolean` and because of that, I converted to this. All of our field validator are of the same type.

xantorres: You should fix the curry function typings by passing the generic type of each of the calls, instead of downgrading the function from being curried to a nested function.

PR #611, opened 2020-01-07
PMK-2448: Added the color coding to ProgressBar component depending on % value.

rahulkurumkar:
PMK-2448: Added the color coding to ProgressBar component depending on % value.
Added absolute value with the total progress bar.

PMK-2448-Resource-utilization-should-report-absolute-values-as-well


src/app/core/components/Empty.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { makeStyles } from '@material-ui/styles'
+
+const useStyles = makeStyles(() => ({
+  root: {
+    fontSize: '20px',
+  }
+}))
+
+const Empty: React.FunctionComponent = (): JSX.Element => {
```
jruddell: We dont need a component to render empty data. `Empty` is also so generic its hard to know what this is about. Some of our columns in the table render `N/A` why not just keep that format? Is there a task asking for this?

rahulkurumkar: Currently, it was not displaying anything so I thought to add this. Will remove this component and will keep existing (not displaying anything)

jruddell: Sorry I wasn't asking you to not show anything. But rather to show something consistently. Meaning, we have `N/A` on some fields and the dash. We want to show something when its empty, lets just stick with `N/A`

src/app/core/components/progress/ProgressBar.js
```diff
@@ -54,7 +54,21 @@ const useStyles = makeStyles(theme => ({
       ? 'linear-gradient(45deg, rgba(255, 255, 255, 0.15) 25%, transparent 25%, transparent 50%, rgba(255, 255, 255, 0.15) 50%, rgba(255, 255, 255, 0.15) 75%, transparent 75%, transparent)'
       : null,
     backgroundSize: '40px 40px',
-    backgroundColor: ({ animated }) => animated ? '#4AA3DF' : '#4ADF74',
+    backgroundColor: ({ animated, percent }) => {
+      if (animated) {
+        return '#4AA3DF'
+      } else {
```
xantorres: You can remove this else statement. just leave the switch as the default fallback action for the function.

xantorres:
The tooltip is not necessary anymore, as we are already showing the total value in the label itself.

src/app/core/components/progress/ProgressBar.js
rahulkurumkar: Fixed

src/app/core/components/progress/ProgressBar.js
```diff
@@ -54,7 +54,21 @@ const useStyles = makeStyles(theme => ({
       ? 'linear-gradient(45deg, rgba(255, 255, 255, 0.15) 25%, transparent 25%, transparent 50%, rgba(255, 255, 255, 0.15) 50%, rgba(255, 255, 255, 0.15) 75%, transparent 75%, transparent)'
       : null,
     backgroundSize: '40px 40px',
-    backgroundColor: ({ animated }) => animated ? '#4AA3DF' : '#4ADF74',
+    backgroundColor: ({ animated, percent }) => {
```
brennancheung: Color values should be pulled from theme.

The switch / case statement doesn't seem warranted.  Simple if statements would be cleaner.  Especially if you use shorthand if statements on a single line.

Why is the rounding needed at all?

```javascript
if (animated) return '...'
if (percent >= 90) return '...'
if (percent >= 80) return '...'
return '...'
```

rahulkurumkar: Done

PR #613, opened 2020-01-07
Pie graph loader enhancements

kevinylee1991:
Screen Shot 2020-01-07 at 3 01 18 PM

### DIFF
--- a/src/app/core/components/graphs/PieGraph.tsx
+++ b/src/app/core/components/graphs/PieGraph.tsx
@@ -24,11 +24,12 @@ const emptyData = [
 const PieGraph = ({ data, sideLength, arcWidth, percent, primary, empty, ...rest }: Props) => {
   const theme: any = useTheme()
   const radius = Math.floor(sideLength / 2)
+  const items = empty ? emptyData : data
 
   return (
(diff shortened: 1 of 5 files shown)


src/app/core/components/graphs/PieGraph.tsx
```diff
@@ -2,28 +2,33 @@ import React from 'react'
...
+  data: PieDataEntry[]
   sideLength: number
   arcWidth: number
   percent: number
   primary: string
+  empty: boolean
 }
 
-const PieGraph = ({ data, sideLength, arcWidth, percent, primary, ...rest }: Props) => {
+const emptyData = [
```
jruddell: what is the purpose of this?

kevinylee1991: The emptyData thing is so that the gray pie arc shows up. Since the pie works by adding up the objects in the array passed to the data prop, if it is 0 then no arc shows up at all. So I added this preset data to represent an empty data set.

jruddell: Hmm so if you hover on the pie chart do you see a tooltip with the text `empty`? I think its better for there to be no string rather than `empty` displayed

PR #615, opened 2020-01-08
Onboard

jruddell:
image

image


src/app/core/components/expansionPanel/ExpansionPanel.js
```diff
@@ -1,83 +1,147 @@
...
+  children,
+  completed,
+  stepNumber,
+  summary,
+  onSkip = undefined,
+  skipConfirmTitle = undefined,
+}) => {
   const classes = useStyles()
 
+  const [overlay, setOverlay] = useState(false)
```
brennancheung: `showModal` or `showOverlay` would make it more obvious.  I wasn't sure by the name at first what it was for.

src/app/plugins/kubernetes/components/dashboard/DashboardPage.tsx
```diff
@@ -195,34 +207,92 @@ const bottomReports = [
...
+
+  const handleComplete = useCallback(() => {
+    forceUpdate()
+  }, [])
+
+  const initialExpandedClusterPanel = useMemo(
+    () =>
+      Math.min(
+        ...[hasClusters, hasAccess, hasMonitoring].reduce(
+          (ret, val, idx) => (!val ? ret.concat(idx) : ret),
```
brennancheung: I'm not sure what is going on here or why.  It kind of looks like you are trying to find the idx of the first one?  Really can't tell.  If that that is the case maybe use `find`(it has an index as well).  If you still need to use reduce then define the reducer as a function outside and give it a good name might clear things up.  At a minimum add some comments here.

src/app/plugins/kubernetes/components/onboarding/cluster-setup.tsx
```diff
@@ -0,0 +1,200 @@
...
+  !!localStorage.getItem(OnboardingMonitoringSetup)
+
+export const clustersHaveAccess = () => !!localStorage.getItem(OnboardingAccessSetup)
+
+interface Props {
+  onComplete: () => void
+  initialPanel?: number
+}
+
+enum Panels {
```
brennancheung: This seems a bit low level and maybe we should factor out this type of scenario (having multiple panels with 1 being active).

PR #619, opened 2020-01-14
PMK-2476: Corrected the enabling and disabling monitoring action for cluster

rahulkurumkar:


src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -283,11 +283,11 @@ export const clusterActions = createCRUDActions(clustersCacheKey, {
...
     updateTag: async ({ cluster, key, val }, prevItems) => {
-      const tags = cluster.tags || {}
+      const tags = cluster.tags ? { ...cluster.tags } : {}
+      tags[key] = val
+      const body = { tags }
 
-      if (key in tags) {
-        tags[key] = val
-      }
+      await qbert.updateCluster(cluster.uuid, body)
```
xantorres:
```
const body =  { 
   tags: { ...cluster.tags || {}, [key]: val }
}
await qbert.updateCluster(cluster.uuid,)
```

rahulkurumkar: Addressed

PR #620, opened 2020-01-14
PMK-2498 - Implement created at column at clusters table and order by it

murilovmachado:
Screen Shot 2020-01-14 at 19 25 15


src/app/core/components/listTable/ListTable.js
```diff
@@ -319,7 +319,7 @@ class ListTable extends PureComponent {
     const { searchTarget, data, filters, onSortChange } = this.props
     const { searchTerm } = this.state
 
-    const sortedData = onSortChange ? data : this.sortData(data)
+    const sortedData = onSortChange ? this.sortData(data) : data
```
xantorres: Please don't use ListTable to do the sorting logic, it is cluttered enough as it is now. Clusters are being ordered by the clusters loader, you shouldn't be changing anything about ListTable.

murilovmachado: @xantorres I've just updated the PR. Please take another look when you can.

PR #623, opened 2020-01-15
[PMK-2515] Manage plugin changes when switching regions and reload features.json with each region switch

kevinylee1991:
First pass at features.json loading by region in order to support ironic setup wizard for 4.2 release.

In the future, we will probably need to have some sort of feature flag system in place that extends beyond the AuthenticatedContainer and navbar. Perhaps with the proposed switch to redux or in the react context?


src/app/core/components/Navbar.js
```diff
@@ -465,14 +474,16 @@ class Navbar extends PureComponent {
...
-    const { classes, open } = this.props
+    const { classes, open, stack } = this.props
     return <div className={classes.sliderContainer}>
       {open && <a href={clarityDashboardUrl}>
         <ChevronLeftIcon className={classes.sliderArrow} />
       </a>}
       <div className={clsx(classes.sliderLogo, {
-        [classes.sliderLogoOpen]: open,
-        [classes.sliderLogoClosed]: !open,
+        [classes.kubernetesLogoOpen]: open && stack === 'kubernetes',
```
brennancheung: You can do this in the `makeStyles`.  You can access the props inside the theming code like this: https://material-ui.com/styles/basics/#adapting-based-on-props

PR #626, opened 2020-01-16
Integration tests and fixes for users and tenants management

xantorres:


cypress/fixtures/session.json
```diff
@@ -1,26 +1,41 @@
 {
   "userTenants": [
     {
-      "id": "52d5f76e-7af8-447f-a5a3-683d9c3b3c4a",
+      "id": "2915d709-0cea-4afc-b182-b0d118818e48",
```
brennancheung: Do we need to hard-code these ids in the simulator to make them match?

cypress/integration/tenants.spec.js
```diff
@@ -0,0 +1,76 @@
...
+
+      cy.contains('Complete').click().wait(1000)
+      cy.contains('Tenant Tenant#Test updated successfully')
+      cy.contains('Tenant#Test edited')
+    })
+  })
+
+  context('remove tenant', () => {
+    it('delete the newly created tenant', () => {
+      cy.contains('Tenant #1').click()
```
brennancheung: Might be brittle if we ever have `Tenant #11`.

xantorres: I think we'll never have that. In any way, what do you propose?

PR #628, opened 2020-01-18
PMK-2509 Enable AWS

brennancheung:


src/app/plugins/kubernetes/components/infrastructure/clusters/aws/AwsClusterRequirements.tsx
```diff
@@ -37,7 +37,7 @@ const AwsClusterRequirements = ({ onComplete }) => {
   return (
     <ClusterRequirementsContainer title="Amazon AWS Deployment">
       <Typography className={classes.text}>Use your existing AWS credentials to create and manage Kubernetes clusters and associated resources within your AWS public cloud environment.</Typography>
-      <Typography className={classes.text}>You can create multiple AWS cloud providers - each AWS cloud provider should be associated with a unique set of AWS credentials.</Typography>
+      <Typography className={classes.text}>You can create multiple AWS cloud providers; each AWS cloud provider should be associated with a unique set of AWS credentials.</Typography>
```
kevinylee1991: maybe just a period instead of semicolon?

brennancheung: I think semicolon makes the most sense here.  Closely related independent clauses are supposed to be separated by a semicolon.  https://www.grammarly.com/blog/semicolon/

Em-dash is when they are related but can't stand on their own.  Also, em-dash is more for when it is parenthetical or some kind of "aside".

PR #631, opened 2020-01-22
[PMK-2133] Added cluster filter to nodes page

kevinylee1991:
One thing I'm curious about being improved is the actual filtering that is done in the ListPage component -- it currently gets run 3x whenever the cluster picker gets changed.

Screen Shot 2020-01-21 at 3 38 52 PM


src/app/plugins/kubernetes/components/infrastructure/nodes/NodesListPage.js
```diff
@@ -21,6 +21,51 @@ import RemoteSupportDialog from './RemoteSupportDialog'
...
+
+const ListPage = ({ ListContainer }) => {
+  return () => {
+    const { params, getParamsUpdater } = usePrefParams(defaultParams)
+    const [data, loading, reload] = useDataLoader(loadNodes, params)
+    // Filter nodes based on cluster
+    const filteredNodes = data.filter(node => {
+      if (params.clusterId === allKey) {
+        return true
+      } else if (params.clusterId === '') {
```
brennancheung: Don't need `else` if you have `return` above.  Might be cleaner with these on a single line.  Some of this conditional branches might be able to be combined into a single compound conditional.

src/app/plugins/kubernetes/components/infrastructure/nodes/NodesListPage.js
```diff
@@ -21,6 +21,51 @@ import RemoteSupportDialog from './RemoteSupportDialog'
...
+    return <ListContainer
+      loading={loading}
+      reload={reload}
+      data={filteredNodes}
+      getParamsUpdater={getParamsUpdater}
+      filters={<>
+        <ClusterPicklist
+          onChange={getParamsUpdater('clusterId')}
+          value={params.clusterId}
+          noneLabel="No Cluster"
```
brennancheung: Not sure about the wording for `No Cluster` but it could work.  `Unassigned` might be another possibility but I would see what others think makes sense as well.

kevinylee1991: Chris specified that he wanted it called No Cluster, we can always change it easily later

PR #632, opened 2020-01-22
PMK-2502, Updates to the status so you can use it more easily

jruddell:
Cluster status page rework, cleanup of react errors, routing updates.

image

image

image


src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatus.tsx
```diff
@@ -0,0 +1,162 @@
...
+    '&:before': {
+      content: '\' \'',
+      height: ({ variant }) => variant === 'header' ? 30 : 14,
+      width: ({ variant }) => variant === 'header' ? 30 : 14,
+      marginTop: ({ variant }) => variant === 'header' ? theme.spacing() : 0,
+      marginRight: 3,
+      borderRadius: '50%',
+      // display: ({ status }) => !status || ['loading', 'error'].includes(status) ? 'none' : 'inline-block',
+      backgroundColor: ({ status }) => ({
+        ok: '#31DA6D',
```
brennancheung: Should we put these in theme.  Also, some are all caps for the hex values and other lowercase.

jruddell: That was existing code, just looks new because i converted it to typescript. I'll add to the theme though.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterDetailsPage.js
```diff
@@ -1,38 +1,168 @@
...
-      <Tab value="nodesAndHealthInfo" label="Nodes & Health Info"><ClusterNodes /></Tab>
-    </Tabs>
-  </PageContainer>
+  const [clusters, loading] = useDataLoader(clusterActions.list)
+  const cluster = clusters.find((x) => x.uuid === match.params.id) || {}
+  return (
+    <PageContainer
+      header={
+        <>
+          <span />
```
xantorres: What's the point of this?

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterDetailsPage.js
```diff
@@ -1,38 +1,168 @@
...
+          <SimpleLink src={pathToClusters()} className={classes.backLink}>
```
xantorres: calling `pathToClusters()` every time is useless, given that the result will always be the same, better to just call it outside of the render method and assign it to a `const`

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterDetailsPage.js
```diff
@@ -1,38 +1,168 @@
...
+          <UsageWidget units="GHz" title="Compute" stats={usage.compute} />
+          <UsageWidget units="GiB" title="Memory" stats={usage.memory} />
+          <UsageWidget units="GiB" title="Storage" stats={usage.disk} />
+        </div>
+      </Grid>
+    </Grid>
+  )
+}
+
+const HeaderCard = ({ title, body }) => {
```
xantorres: instead of a `body` prop better use just the generic `children`, also you will prevent having to pass a `<></>` to render multiple elements.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterInfo.js
```diff
@@ -102,32 +100,18 @@ const useStyles = makeStyles(theme => ({
 const ClusterInfo = () => {
   const { match } = useReactRouter()
   const classes = useStyles()
-  const [clusters, loading] = useDataLoader(clusterActions.list)
+  const [clusters] = useDataLoader(clusterActions.list)
```
xantorres: no `loading` here? Why?

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterNodes.js
```diff
@@ -20,64 +24,46 @@ const ClusterNodes = () => {
   const [clusters, loadingClusters] = useDataLoader(clusterActions.list)
   const [nodes, loadingNodes, reload] = useDataLoader(loadNodes)
   const handleRefresh = useCallback(() => reload(true), [reload])
-  const cluster = clusters.find(cluster => cluster.uuid === match.params.id)
-  const [masterNodes, workerNodes] = useMemo(() => {
+  const cluster = clusters.find((cluster) => cluster.uuid === match.params.id)
```
xantorres: why the added parentheses?

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterNodes.js
```diff
@@ -20,64 +24,46 @@ const ClusterNodes = () => {
...
+  const nodesInCluster = useMemo(() => {
     if (cluster) {
       const clusterNodesUids = pluck('uuid', cluster.nodes)
-      const clusterNodes = nodes.filter(node => clusterNodesUids.includes(node.uuid))
-      const isMasterNode = node => node.isMaster === 1
-      return partition(isMasterNode, clusterNodes)
+      return nodes.filter((node) => clusterNodesUids.includes(node.uuid))
     }
     return emptyArr
   }, [cluster, nodes, match])
```
xantorres: you are not using `match` inside of the memo callback

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterNodes.js
```diff
@@ -20,64 +24,46 @@ const ClusterNodes = () => {
...
-    onReload: handleRefresh,
-    showCheckboxes: false,
-    compactTable: true,
-  }
-
-  const removeColumnIds = (columns, ids) => columns.filter(column => !ids.includes(column.id))
-
   const addTaskStatusModal = (columns) => {
-    const healthStatusColumn = { ...columns.find(column => column.id === 'healthStatus') }
+    const healthStatusColumn = { ...columns.find((column) => column.id === 'healthStatus') }
```
xantorres: this `{...}` spread thing here is useless.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterNodes.js
```diff
@@ -20,64 +24,46 @@ const ClusterNodes = () => {
...
-    emptyText: 'No instances found.',
-    uniqueIdentifier: 'uuid',
-    onReload: handleRefresh,
-    showCheckboxes: false,
-    compactTable: true,
-  }
-
-  const removeColumnIds = (columns, ids) => columns.filter(column => !ids.includes(column.id))
-
   const addTaskStatusModal = (columns) => {
```
xantorres: you should declare this function outside of the render method

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterNodes.js
```diff
@@ -20,64 +24,46 @@ const ClusterNodes = () => {
...
 
   return (
-    <div>
-      <ClusterNodesOverview cluster={cluster} />
-      <h2>Master Nodes</h2>
-      <MasterNodesTable data={masterNodes} loading={loadingClusters || loadingNodes} />
-      <h2>Worker Nodes</h2>
-      <WorkerNodesTable data={workerNodes} loading={loadingClusters || loadingNodes} />
+    <Progress loading={loadingClusters || loadingNodes}>
+      <NodesTable data={nodesInCluster} loading={loadingClusters || loadingNodes} />
```
xantorres: the `loading` proop here is redundant, as you are already displaying a loading spinner using the `Progress` component

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatus.tsx
```diff
@@ -0,0 +1,161 @@
...
+      backgroundColor: ({ status }) => ({
+        ok: theme.palette.pieChart.success,
+        pause: theme.palette.pieChart.warning,
+        fail: theme.palette.pieChart.error,
+      }[status] || theme.palette.pieChart.error)
+    },
+  },
+  loading: {
+    marginRight: theme.spacing(0.375),
+    fontSize: ({ variant }) => variant === 'header' ? 14 : 12,
```
xantorres: is better that we get used to using theme variables for font sizes and colors instead of hardcoded numerical values (although I admit being guilty of not doing that sometimes too)

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterNodes.js
jruddell: No, I was seeing a `No data found` message flash. This addresses that

jruddell: No, I was seeing a `No data found` message flash. This addresses that

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterDetailsPage.js
jruddell: Spacing, the page container does a space between, but one element would render it on the left instead of the right. The empty span is for that. I'll be refactored once we have a breadcrumbs component

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterDetailsPage.js
jruddell: Assigning it to a const is useless unless you memorize or something. The overhead isnlt really a big deal so I didn't care to

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterDetailsPage.js
jruddell: children is a good use case for this 👍

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterNodes.js
jruddell: Auto formatter adds parentheses. It doesn't break linting, is there an issue with it?

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatus.tsx
jruddell: True, I got lazy because it was 3-4am 🤓 will update.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterInfo.js
jruddell: the cluster is loaded by a parent element with a loading spinner. The content here renders at first pass so the loader is redundant.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterNodes.js
jruddell: why? its using / setting local state?

PR #633, opened 2020-01-22
[PMK-2508] Create Banner component

xantorres:
Example:
image

This is just the implementation of the base components (the base `Banner` component + `OnboardingBanner`), until we have further instructions on the specific steps. It is not being shown anywhere yet.


src/app/core/components/notifications/BannerItem.tsx
```diff
@@ -0,0 +1,142 @@
...
+    height: 30,
+    display: 'flex',
+    flexFlow: 'row nowrap',
+    textAlign: 'center',
+    color: theme.palette.secondary.contrastText,
+    justifyContent: 'center',
+    width: '100%',
+  },
+  success: {
+    backgroundColor: green[600],
```
brennancheung: Do we want to allow these to be customized in the theme more?  If so, we should create or re-use a theme palette swatch value instead of hard-coding it to a green color.

src/app/core/components/notifications/BannerContainer.tsx
```diff
@@ -0,0 +1,30 @@
...
+import BannerItem from 'core/components/notifications/BannerItem'
+import { BannerContext } from 'core/providers/BannerProvider'
+
+const useStyles = makeStyles<Theme>(theme => ({
+  root: {
+    // Intercom uses a ridiculously high zIndex so we have to be even more ridiculous
+    zIndex: 9999999999,
+    display: 'flex',
+    flexFlow: 'column nowrap',
+    position: 'fixed',
```
brennancheung: Will we need to change the `top` offset if we change other things about the layout.  Wonder if there is a way we can make it more universal.

src/app/core/components/notifications/BannerItem.tsx
```diff
@@ -0,0 +1,142 @@
...
+    display: 'inline-block',
+  },
+  close: {
+    position: 'absolute',
+    right: 0,
+  },
+}))
+
+// Color overrides for links and other elements depending on the current banner type
+// TODO: Fix these colors when we have the definitive designs
```
brennancheung: Put these into the theme so they can be customized later instead of hard-coding below.

xantorres: This is very specific for this component, are you sure we want to put it in the theme file?

src/app/core/components/notifications/BannerItem.tsx
jruddell: would be best to create a custom banner entry inside palette for its different states.

src/app/core/components/notifications/BannerContainer.tsx
jruddell: I think we can absolutely position it to the `AuthenticatedContainer`'s top most element as that is full width but just below the header.

src/app/core/components/notifications/BannerItem.tsx
jruddell: yes, if we want to design future themes we need to be able to customize them.

src/app/core/components/notifications/BannerItem.tsx
```diff
@@ -0,0 +1,142 @@
...
+      size="small"
+      onClick={onClose}
+      className={classes.close}
+    >
+      <CloseIcon className={classes.icon} />
+    </IconButton>
+    <span className={classes.message}>
+      <Icon className={clsx(classes.icon, classes.iconVariant)} />
+      <div className={classes.text}>
+        <ThemeProvider theme={contentTheme}>
```
jruddell: why do you need to render the theme provider here?

src/app/core/containers/AuthenticatedContainer.js
```diff
@@ -138,8 +142,13 @@ const loadRegionFeatures = async (setRegionFeatures, history) => {
   }
 }
 
+const checkInitialRequirements = (showBanner) => {
+  // TODO: implement initial checks and show a banner (or banners) if required
```
jruddell: The logic used to determine if we should show onboarding or not can be moved to a helper method so we can reuse it here.

src/app/core/containers/AuthenticatedContainer.js
```diff
@@ -173,7 +180,9 @@ const AuthenticatedContainer = () => {
           [classes.contentShift]: drawerOpen,
           [classes['contentShift-left']]: drawerOpen,
         })}>
+          {/* <BannerContainer /> */}
```
xantorres: This will be decommented when we enable the Banner feature

src/app/plugins/kubernetes/components/onboarding/OnboardingBanner.tsx
```diff
@@ -0,0 +1,74 @@
...
+  button: {
+    lineHeight: 1.2,
+    minHeight: theme.spacing(4.5),
+    borderRadius: 0,
+    backgroundColor: theme.palette.secondary.contrastText,
+    color: '#F5A623',
+  },
+}))
+
+// TODO
```
jruddell: whats this todo for?

src/app/core/providers/BannerProvider.tsx
```diff
@@ -1,65 +1,22 @@
...
-        append(payload)
-      )(state)
-    case 'remove':
-      return except(payload, state)
-    default:
-      return state
+export const BannerContext = createContext<BannerContextType>({
+  bannerContainer: null,
+  setBannerContainer: container => {
+    console.error('BannerContainer not found')
```
jruddell: were calling `setBannerContainer` in `BannerContainer.tsx` but we just console.error here? does this still need to be implemented?

xantorres: Just in case that someone decides to use this provider somewhere else and forgets to add the container.

src/app/plugins/kubernetes/components/onboarding/OnboardingBanner.tsx
xantorres: This will be completed when we have more specifications about the implementation of the onboarding steps.

PR #634, opened 2020-01-22
[PMK-2519] Added etcd backup option to cluster create/edit

kevinylee1991:


src/app/plugins/kubernetes/components/infrastructure/clusters/EditClusterPage.js
```diff
@@ -25,16 +27,29 @@ const EditClusterPage = () => {
...
   const initialValues = useMemo(
     () => pipe(
       find(propEq('uuid', clusterId)),
       when(isNil, always(emptyObj)),
-      ({ tags = {}, ...cluster }) => ({
+      ({ tags = {}, etcdBackup = {}, ...cluster }) => ({
         ...cluster,
         tags: tagsObjToArr(tags),
+        etcdBackup: !!etcdBackup.isEtcdBackupEnabled,
+        etcdStoragePath: path(['storageProperties, localStorage'], etcdBackup) || '/etc/pf9/etcd-backup',
```
jruddell: make `'/etc/pf9/etcd-backup'` a constant default variable and reuse it rather than having the string literal in a few places

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -159,6 +159,23 @@ const createGenericCluster = async (body, data, loadFromContext) => {
     custom: data.customRuntimeConfig,
   }[data.runtimeConfigOption]
 
+  // This is currently supported by all cloud providers except GCP (which we
+  // don't have yet anyways)
+  if (data.etcdBackup) {
```
jruddell: need to make this a helper function so you aren't copy pasting it. build the object like this `body.etcdBackup = getEtcdPayload(data)`. let that helper return an object for when its enabled / disabled

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -249,7 +267,28 @@ export const clusterActions = createCRUDActions(clustersCacheKey, {
...
+        intervalInMins: params.etcdBackupInterval,
+      }
+    } else {
+      body.etcdBackup = {
+        isEtcdBackupEnabled: 0,
+      }
+    }
+
     await qbert.updateCluster(uuid, body)
+    // Doing this will help update the table, but the cache remains incorrect...
```
jruddell: if the cache is incorrect you should call `.invalidateCache()` on the action entity. I think in this case that would be `clusterActions.invalidateCache()` would have to test to be sure though

src/app/plugins/kubernetes/components/infrastructure/clusters/AddAwsClusterPage.js
```diff
@@ -48,6 +48,8 @@ const initialContext = {
   runtimeConfigOption: 'default',
   isPrivate: false,
   internalElb: false,
+  etcdStoragePath: '/etc/pf9/etcd-backup',
+  etcdBackupInterval: 1,
```
brennancheung: Add a comment about what the unit is?  1 second, 1 minute, hour, day?

src/app/plugins/kubernetes/components/infrastructure/clusters/AddAwsClusterPage.js
```diff
@@ -557,6 +559,37 @@ const AddAwsClusterPage = () => {
                         info={<div>Allows this cluster to run privileged containers. Read <ExternalLink url="https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities">this article</ExternalLink> for more information.</div>}
                       />
 
+                      {/* Etcd Backup */}
+                      <CheckboxField
+                        id="etcdBackup"
+                        label="Enable Etcd Backup"
```
brennancheung: `etcd` not `Etcd` I think.  The `etcd` website never shows it with a capital `E`.

src/app/plugins/kubernetes/components/infrastructure/clusters/AddAwsClusterPage.js
```diff
@@ -557,6 +559,37 @@ const AddAwsClusterPage = () => {
...
+                        info="Enable automated etcd backups on this cluster"
+                      />
+
+                      {/* Etcd Storage Path */}
+                      {values.etcdBackup && (
```
brennancheung: Just use 1 conditional that has both these fields in it.  You can use `React.Fragment` to combine them.

src/app/plugins/kubernetes/components/infrastructure/clusters/AddAzureClusterPage.js
```diff
@@ -44,6 +44,8 @@ const initialContext = {
   runtimeConfigOption: 'default',
   useAllAvailabilityZones: true,
   assignPublicIps: false,
+  etcdStoragePath: '/etc/pf9/etcd-backup',
```
brennancheung: This should not be repeated on every page.  Put it somewhere more common.

src/app/plugins/kubernetes/components/infrastructure/clusters/AddAzureClusterPage.js
```diff
@@ -394,6 +396,37 @@ const AddAzureClusterPage = () => {
                         info={<div>Allows this cluster to run privileged containers. Read <ExternalLink url="https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities">this article</ExternalLink> for more information.</div>}
                       />
 
+                      {/* Etcd Backup */}
+                      <CheckboxField
```
brennancheung: Looks like you are repeating these fields.  I think we should make this more common.  Perhaps create a separate file that has these and creating something called `renderEtcdBackupFields` or something like that.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterInfo.js
```diff
@@ -8,16 +8,20 @@ import useDataLoader from 'core/hooks/useDataLoader'
...
 const overviewStats = cluster => ({
-  Status:              cluster.status,
-  CloudProvider:       cluster.cloudProviderName,
-  'Cloud Provider Type': cluster.cloudProviderType,
-  'Kubernetes Version':  cluster.version || 'unavailable',
-  'Containers CIDR':     cluster.containersCidr,
-  'Services CIDR':       cluster.servicesCidr,
-  Privileged:          cluster.privileged,
-  'Unique ID':           cluster.uuid,
+  Status:                 cluster.status,
```
brennancheung: You might need to use `eslint-disable` with whatever the rule is for the indentation here.  If you are going to have some of the fields have single quotes around them, might as well make them all use single quotes to look more consistent.  It looks chaotic mixing the conventions.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
```diff
@@ -249,6 +249,7 @@ export const options = {
     { id: 'hasVpn', label: 'VPN' },
     { id: 'appCatalogEnabled', label: 'App Catalog', render: x => x ? 'Enabled' : 'Not Enabled' },
     { id: 'hasLoadBalancer', label: 'Load Balancer' },
+    { id: 'etcdBackupEnabled', label: 'Etcd Backup', render: x => x ? 'Enabled' : 'Not Enabled' },
```
brennancheung: Use `castBoolToStr` from `app/utils/misc`.

src/app/plugins/kubernetes/components/infrastructure/clusters/EditClusterPage.js
```diff
@@ -57,6 +72,40 @@ const EditClusterPage = () => {
         info="Name of the cluster"
         required
       />
+
+      {/* Etcd Backup */}
```
brennancheung: They are repeated again here too.  Move these fields to a common function.

src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
```diff
@@ -303,6 +305,37 @@ const AddBareOsClusterPage = () => {
                         info={<div>Allows this cluster to run privileged containers. Read <ExternalLink url="https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities">this article</ExternalLink> for more information.</div>}
                       />
 
+                      {/* Etcd Backup */}
```
brennancheung: Same here.  Don't repeat them.

kevinylee1991:
Noticed in the aws cluster creation workflow that api fqdn and services fqdn are required when I believe they should be optional, will confirm this tomorrow. They also don't show up in the review step.

PR #638, opened 2020-01-25
[PMK-2442] Expose user creation with clemency and fix activation page reroute

kevinylee1991:
Tested manually on paavan's DU

### DIFF
--- a/src/app/core/public/ActivateUserPage.tsx
+++ b/src/app/core/public/ActivateUserPage.tsx
@@ -49,7 +49,7 @@ export const ActivateUserPage = props => {
     const validateSecret = async () => {
       try {
         const response = await clemency.verifyActivateLink(username, otp)
-        history.push(`${resetPasswordUrl}/${response.value}`)
+        history.push(resetPasswordUrl.replace(':id', response.value))
       } catch (e) {
         updateParams({ loading: false })
       }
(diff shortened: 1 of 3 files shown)


src/app/plugins/kubernetes/components/userManagement/users/AddUserPage.js
```diff
@@ -65,7 +65,7 @@ const AddUserPage = () => {
                 <FormLabel component="legend"><p>Activate User Account</p></FormLabel>
                 <RadioGroup value={activationType} onChange={
                   e => setActivationType(e.target.value)}>
-                  {false && <FormControlLabel
+                  { <FormControlLabel
```
jruddell: I thought this was broken on the backend side. were you able to test and verify this works?

src/app/core/public/ActivateUserPage.tsx
```diff
@@ -49,7 +49,7 @@ export const ActivateUserPage = props => {
     const validateSecret = async () => {
       try {
         const response = await clemency.verifyActivateLink(username, otp)
-        history.push(`${resetPasswordUrl}/${response.value}`)
+        history.push(resetPasswordUrl.replace(':id', response.value))
```
jruddell: use my path functions for this. theres a file called `routes.ts`. Define the reset password url there and create a path function for it `const pathToResetPasswordUrl = ...` and then using it here would be `history.push(pathToResetPasswordUrl({id: response.value}))`

src/app/plugins/kubernetes/components/userManagement/users/AddUserPage.js
kevinylee1991: Yeah we tested this on paavan's DU (a decco DU)

PR #639, opened 2020-01-26
Updated styles

jruddell:
- Using cards for the form fields on cluster create.
- Updated wording / reduced text
- Better formatting / code formatting in a few places
- Unlocked add AWS Cloud Provider

## BareOS Cluster Create
image

image

image

## Add Resource
image


src/app/core/utils/routes.ts
```diff
@@ -79,9 +83,19 @@ const routes = {
   }
 }
 
-export const pathToCreateCluster = (params?: GenericKVP) => {
+// cluster
+export const pathToAddCluster = (params?: GenericKVP) => {
```
xantorres: All this routes should really be just constants, since most of them are not expecting any parameter anyways. Making them callable makes it one think that they could receive a `params` parameter when it's not the case because it won't make any difference in the output. 
The ones receiving parameter should be renamed to `getPathTo...` and expect a `params` parameter (not optional)

jruddell: This is inaccurate. The point of every url is that it can potentially have query string params. These path functions convert an object to query string params, and will also replace in url params that match react router. 

If you look at the typings to these functions, url's that have path parts that need to be replaced require those params. I'm wanting to put all urls in a central place, but it needs to be extensible. `pathToAddBareOSCluster({name: 'My Awesome Cluster'})` for instance could be used to populate the `name` field of the BareOS cluster form with `My Awesome Cluster` via QS params. We can also apply this logic for our list tables `pathToClusters({search: 'some table search value', rows: 100})`

src/app/core/components/validatedForm/CodeMirror.js
```diff
@@ -22,6 +23,27 @@ const defaultOptions = {
   },
 }
 
+// These styles are to match CodeMirrors. We need to find a good way
+// to re-define their styles so we can use common variables
+const useStyles = makeStyles(theme => ({
+  coderMirrorHeader: {
+    position: 'relative',
+    background: '#f3f3f3',
```
brennancheung: Do we want to make these color values themeable?

src/app/core/themes/defaultTheme.js
```diff
@@ -222,8 +222,8 @@ const theme = {
       medium: 'rgba(2, 194, 172, 0.4)',
     },
     code: {
-      background: '#FDF6F5',
-      text: '#E5513A',
+      background: '#606060',
```
brennancheung: Might as well toss the other hard-coded color values from CodeMirror in here as well.

src/app/core/utils/routes.ts
brennancheung: I thinking making it so there are params allows for flexibility in the future and it's good to keep the path calling convention consistent.

I'm wondering if we should create some kind of system for all these functions so we aren't just repeating ourselves over and over.  Looks like it is just a mapping to a route that accepts some params.

I prefer keeping the same structure of routes and just mapping over that structure to convert the strings to a function that accept params.

Something like `createMappings(routes)` that will output a structure with the functions.

That way we can just do something like `paths.cluster.addBareOs(params)` to generate our links.

src/app/plugins/kubernetes/components/dashboard/DashboardPage.tsx
```diff
@@ -25,6 +25,24 @@ import useDataLoader from 'core/hooks/useDataLoader'
 import Progress from 'core/components/progress/Progress'
 import identity from 'ramda/es/identity'
 import { isAdminRole } from 'k8s/util/helpers'
+import {
+  pathToNodeDownload,
```
brennancheung: This can be cleaned up quite a bit if everything is just on a single object called `paths` that we can drill into to get the necessary functions.

src/app/core/components/validatedForm/CodeMirror.js
jruddell: I do, if you notice the comment above, these colors are defined here and also in the `.css` file for codemirror. Could I put the whole css file in a `makeStyles` global section so we can properly theme?

src/app/core/utils/routes.ts
```diff
@@ -1,107 +1,108 @@
...
-    addDeployments: `${k8sPrefix}/pods/deployments/add`,
-    addServices: `${k8sPrefix}/pods/services/add`,
+    list: new Route(`${k8sPrefix}/pods#pods`),
+    deployments: new Route(`${k8sPrefix}/pods#deployments`),
+    services: new Route(`${k8sPrefix}/pods#services`),
+    add: new Route(`${k8sPrefix}/pods/add`),
+    addDeployments: new Route(`${k8sPrefix}/pods/deployments/add`),
+    addServices: new Route(`${k8sPrefix}/pods/services/add`),
   },
   storage: {
```
brennancheung: K8s has multiple resources that can be considered storage (storage classes, persistent volumes, persistent volume claims, and recently they added snapshots but I don't think we support those yet).

So we should either nest storage classes underneath storage as its own object, or just rename storage to storageClasses, and we can add the others later on.

src/app/core/utils/routes.ts
```diff
@@ -1,107 +1,108 @@
...
-    list: `${k8sPrefix}/storage_classes`,
-    add: `${k8sPrefix}/storage_classes/add`,
+    list: new Route(`${k8sPrefix}/storage_classes`),
+    add: new Route(`${k8sPrefix}/storage_classes/add`),
   },
   logging: {
```
brennancheung: Similar to storage, `logging` is very general and we make support multiple kinds of logging.  So we might want to make this more specific to which logging system we are using here.  Or do something like `logging.fluentd.list` or whatever service we are using.

src/app/plugins/kubernetes/components/dashboard/DashboardPage.tsx
```diff
@@ -43,31 +45,31 @@ const topReports = [
...
-    quantityFn: tenants => ({
+    quantityFn: (tenants) => ({
       quantity: tenants.length,
     }),
   },
   {
     entity: 'deployment',
-    route: '/ui/kubernetes/pods#deployments',
-    addRoute: '/ui/kubernetes/pods/deployments/add',
+    route: routes.pods.deployments.path(),
```
brennancheung: Same here.  `deployments` should be a sibling with `pods` not underneath it.

src/app/plugins/kubernetes/components/dashboard/DashboardPage.tsx
```diff
@@ -77,12 +79,12 @@ const topReports = [
   },
   {
     entity: 'service',
-    route: '/ui/kubernetes/pods#services',
-    addRoute: '/ui/kubernetes/pods/services/add',
+    route: routes.pods.services.path(),
+    addRoute: routes.pods.addServices.path(),
```
brennancheung: `services` should be sibling with `pods`

src/app/core/utils/routes.ts
```diff
@@ -1,107 +1,108 @@
...
+import { CloudProviders } from 'k8s/components/infrastructure/clusters/model'
+
 interface GenericKVP {[key: string]: string}
 
-const routes = {
+// This will make a parameter optional if no type definition is passed
+type RouterPathFn<T> = (...params: OptionalParamType<T>) => string
+type OptionalParamType<T> = T extends null ? [GenericKVP?] : [T & GenericKVP]
+
+class Route<T = null> {
```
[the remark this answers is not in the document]
xantorres: we could also add a default `toString()` implementation to the `Route` class so that we don't need to explicitly call `path()` every time.

src/app/core/utils/routes.ts
jruddell: until we define / support individual ones i dont see a reason to change it

src/app/core/utils/routes.ts
jruddell: same here. lets wait till we support more use cases to better define the path structure

PR #640, opened 2020-01-27
[PMK-2532] Fix issues with app context updating/accessing

xantorres:
Refactored AppProvider from FC to a class. 

I know this will look strange, since we are refactoring most of our stuff to functional components, but in this specific case I have given up, since the problems and complexities of using hooks are too high and is simply much more optimal and easier to use a class component.

Additionally, `this.setState` has the advantage of allowing to pass a callback for when the state update operation has been performed, whereas doing the same with the `useState` hook is much more complicated (I had to create a `useStateAsync` hook which seems to be somewhat buggy)

This should also fix some weird issues related to cache updating.

### DIFF
--- a/src/app/core/providers/AppProvider.tsx
+++ b/src/app/core/providers/AppProvider.tsx
@@ -1,19 +1,18 @@
-import React, { useCallback, SetStateAction, useRef, useEffect } from 'react'
+import React, { PureComponent, SetStateAction } from 'react'
 import {
-  assocPath,
-  flatten,
-  Dictionary,
-  assoc,
-  mergeLeft,
-  lensProp,
-  over,
-  take,
-  prepend,
   always,
+  assoc,
+  assocPath,
+  Dictionary,
+  flatten,
+  lensProp,
+  mergeLeft,
+  over,
+  prepend,
+  take,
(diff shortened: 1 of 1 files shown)


src/app/core/providers/AppProvider.tsx
```diff
@@ -87,26 +86,21 @@ const AppProvider = ({ children }) => {
...
   }
 
-  const setContext = useCallback(async setterFn => {
-    return setContextBase(setterFn instanceof Function
-      ? setterFn
-      : context => ({ ...context, ...setterFn }),
-    )
-  }, [])
+  // We must add a trailing comma to the generic (<T, >) to sidestep the JSX ambiguity
+  setContext = async <T, > (setterFn) => {
```
jruddell: is there supposed to be a second type parameter here (comma after first `<T, >`)?

xantorres: Read the comment above please

xantorres: There's no other way to do this, we need to use a property function rather than a regular function because these method gets passed down as a prop and we need the `this` being binded to the function.

PR #641, opened 2020-01-27
PMK-2534, fix the validator for min length on user password

jruddell:


src/app/plugins/kubernetes/components/userManagement/users/UserPasswordField.js
```diff
@@ -25,7 +25,7 @@ const UserPasswordField = ({ value }) => <>
   <Typography variant="body1" component="div">
     Password must contain the following:
     <List dense>
-      <CheckListItem checked={hasMinLength(8, value)}>
+      <CheckListItem checked={hasMinLength(8)(value)}>
```
brennancheung: I'd rather see `hasMinLength` be a curried function.  Just wrap it with ramda's `curry` function and you can call it with partial application.  A curried function looks cleaner and has flexibility in how it can be applied.

jruddell: we dont need flexibility here though right? this is pretty straight forward imo

PR #644, opened 2020-01-28
Cluster details

jruddell:
image

image

image

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterDetailsPage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterDetailsPage.js
@@ -20,7 +20,7 @@ import { routes } from 'core/utils/routes'
 import FontAwesomeIcon from 'core/components/FontAwesomeIcon'
 import PollingData from 'core/components/PollingData'
 
-const tenSeconds = 1000 * 10
+const oneSecond = 1000
 
 const useStyles = makeStyles((theme) => ({
   cardBoarder: {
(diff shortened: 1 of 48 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterDetailsPage.js
```diff
@@ -104,8 +109,9 @@ const ClusterDetailsPage = () => {
         </>
       }
     >
-      <Progress loading={loading}>
-        <ClusterStatusAndUsage cluster={cluster} />
+      <>
+        <PollingData hidden loading={loading} onReload={reload} refreshDuration={tenSeconds} />
```
brennancheung: `refreshDuration{10 * seconds}` so we have flexibility to change the duration without renaming variables in the future.

src/app/plugins/kubernetes/components/infrastructure/nodes/ConvergingNodeBreakdown.tsx
```diff
@@ -73,11 +82,27 @@ const useStyles = makeStyles<Theme, {}>((theme) => ({
...
+  }
+  if (all.length === completed.length) {
+    return 'success'
+  }
+  if (completed.length < all.length && !hasFailed) {
+    return 'primary'
+  }
+  return 'error'
+}
+const tenSeconds = 1000 * 10
```
brennancheung: Same here, just use `10 * seconds`.

PR #645, opened 2020-01-29
Added a few segment events for PMKFT GA

kevinylee1991:
Bare-bones segment event implentation for a few metrics (logout, login, cluster creation). Still need to confirm with chris for fields that he wants to include. A UI-wide implementation will for sure need more discussion.

### DIFF
--- a/src/app/core/public/LoginPage.js
+++ b/src/app/core/public/LoginPage.js
@@ -15,6 +15,7 @@ import { forgotPasswordUrl } from 'app/constants.js'
 import { pathJoin } from 'utils/misc'
 import { imageUrlRoot, dashboardUrl } from 'app/constants'
 import moment from 'moment'
+import track from 'utils/tracking'
 
 const styles = theme => ({
   root: {
@@ -97,6 +98,13 @@ export class LoginPage extends React.PureComponent {
     const authSuccess = await onAuthSuccess({ username, unscopedToken, expiresAt, issuedAt })
 
     if (authSuccess) {
+      track('event', {
+        eventName: 'PF9 Signed In',
+        eventProperties: {
+          username,
+          duDomain: window.location.origin,
+        }
+      })
       return this.props.history.push(dashboardUrl)
     }
     return this.setState({ loginFailed: true, loading: false })
--- a/src/app/core/public/LoginPage.js
+++ b/src/app/core/public/LoginPage.js
@@ -15,7 +15,7 @@ import { forgotPasswordUrl } from 'app/constants.js'
 import { pathJoin } from 'utils/misc'
 import { imageUrlRoot, dashboardUrl } from 'app/constants'
 import moment from 'moment'
-import track from 'utils/tracking'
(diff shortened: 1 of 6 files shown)


src/app/core/public/LogoutPage.js
```diff
@@ -2,12 +2,21 @@ import React, { useEffect, useContext } from 'react'
...
+import track from 'utils/tracking'
 
 // We are abusing the React component system a little bit here.  This is really
 // nothing but an action but I didn't want to clutter the Navbar component with
 // more code.  This gives us a nice clean separation.
 const LogoutPage = () => {
-  const { destroySession } = useContext(AppContext)
+  const { destroySession, session: { username } } = useContext(AppContext)
+
+  track('event', {
```
xantorres: Please put this inside of the `useEffect` callback below.

xantorres: Please put this inside of the `useEffect` callback below.

webpack.config.js
```diff
@@ -112,7 +112,8 @@ module.exports = {
     extractCSS,
     new HtmlWebpackPlugin({
       inject: true,
-      template: isProd ? './static/index-with-tracking.html' : './static/index.html',
+      // template: isProd ? './static/index-with-tracking.html' : './static/index.html',
```
xantorres: If this is not needed anymore please remove the comment along with the `index-with-tracking.html` file, otherwise we will end up trashing the already cluttered app with a lot of unused files.

jruddell: It looks like its still being used in the line below, just inversed? was this for development and needs to be reverted so tracking is on prod?

kevinylee1991: oops yeah that was for development, I just thought about that

src/app/core/public/LogoutPage.js
```diff
@@ -2,14 +2,23 @@ import React, { useEffect, useContext } from 'react'
...
+  const { destroySession, session: { username } } = useContext(AppContext)
 
   useEffect(() => {
+    track('event', {
+      eventName: 'PF9 Logged Out',
+      eventProperties: {
+        username,
+        du_domain: window.location.origin,
+      }
+    })
```
jruddell: add a closing `, []` to `useEffect` so this only runs once. without the empty array it'll run every render cycle

src/app/core/public/LoginPage.js
```diff
@@ -97,6 +98,13 @@ export class LoginPage extends React.PureComponent {
     const authSuccess = await onAuthSuccess({ username, unscopedToken, expiresAt, issuedAt })
 
     if (authSuccess) {
+      track('event', {
```
brennancheung: How often are we going to do anything other than `track('event', ...)`?  Maybe we should add a util for `trackEvent({ ... })` so that we can shorten the code a little bit if this is going to be added all over the place.

src/app/core/public/LogoutPage.js
kevinylee1991: wait it's already there

PR #646, opened 2020-01-29
[PMK-2552] Fix password reset error alerts

kevinylee1991:
Screen Shot 2020-01-28 at 6 24 03 PM
Screen Shot 2020-01-28 at 6 28 40 PM

### DIFF
--- a/src/app/core/public/ForgotPasswordPage.js
+++ b/src/app/core/public/ForgotPasswordPage.js
@@ -41,9 +41,6 @@ const useStyles = makeStyles(theme => ({
     marginTop: theme.spacing(1),
     textAlign: 'justify'
   },
-  errorContainer: {
-    width: '150px'
-  }
 }))
 
 const ForgotPasswordPage = props => {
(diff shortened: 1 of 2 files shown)


src/app/core/public/ForgotPasswordPage.js
```diff
@@ -126,9 +123,7 @@ const ForgotPasswordPage = props => {
                       onChange={handleEmailChange()}
                     />
                     {params.isError && (
-                      <div className={classes.errorContainer}>
-                        <Alert variant="error" message="Something went wrong" />
-                      </div>
+                      <Alert small variant="error" message="Something went wrong" />
```
xantorres: "Something went wrong" is not a very useful feedback message, we should probably provide a more user friendly message.

jruddell: I'd agree, we can get madhura / chris to give us better copy

kevinylee1991: oh yeah i'll ask her really quick

PR #647, opened 2020-01-29
add a back link

jruddell:
image


src/app/plugins/kubernetes/components/infrastructure/nodes/DownloadCliPage.tsx
```diff
@@ -9,18 +9,32 @@ import PageContainer from 'core/components/pageContainer/PageContainer'
...
-const ClusterDetailsPage: FunctionComponent = () => {
-  const { spacer } = useStyles({})
+const DownloadCliPage: FunctionComponent = () => {
+  const { spacer, backLink } = useStyles({})
   return (
-    <PageContainer>
+    <PageContainer
+      header={
+        <>
+          <span />
```
xantorres: You can prevent having to use a `<>` and this ugly empty `<span />` by adding `alignSelf: 'flex-end',` in the backLink styles. The same applies for the `ClusterDetailsPage`

jruddell: @xantorres have you tried this? It wont align it to the right.

PR #654, opened 2020-01-30
Feedback

jruddell:


src/app/core/utils/routes.ts
```diff
@@ -124,6 +124,13 @@ const createUrlWithQueryString = (url: string, params?: GenericKVP) => {
     return url
   }
   params = { ...params }
+  let hash = ''
+
+  if (url.includes('#')) {
```
brennancheung: JS has URL parsing built in already.

https://developer.mozilla.org/en-US/docs/Web/API/URL

src/app/core/utils/routes.ts
```diff
@@ -12,12 +12,37 @@ class Route<T extends OptionalGenericKVP = null> {
...
+  private createUrlWithQueryString (url: URL, params: OptionalParamType<T>) {
+    if (!params || Object.keys(params || []).length === 0) {
+      return this.url
+    }
+    const fields: GenericKVP = { ...params as any }
+
+    if (url.pathname.includes(':')) {
+      // nice utility to reconstruct urls from objects / models
+      (url.pathname.match(/:([0-9_a-z]+)/gi) || []).forEach((match) => {
+        const key = match.replace(':', '')
```
brennancheung: It should already be in the first capture group from your regex.

src/app/core/utils/routes.ts
```diff
@@ -12,12 +12,37 @@ class Route<T extends OptionalGenericKVP = null> {
...
+      new URL(this.url, window.location.origin),
+      params
+    )
   }
 
   public toString (): string {
     return this.path(null)
   }
+
+  private createUrlWithQueryString (url: URL, params: OptionalParamType<T>) {
```
brennancheung: This needs comments so I can see the inputs / outputs and maybe a sample usage example.  The code is very imperative and you have to carefully read all the code just to understand what it does and how to use it.  Very hard to tell what this is doing at a glance.

There's also a lot of mutations going on so that makes it harder to understand.  If it can be split up or made more declarative it would be easier to follow.

jruddell: There are tests for this so you can see usage. I'll add a comment but theres not much I can do about the mutations. If I'm going to use `URL` then I have to mutate the pathname of that class's instance to replace the dynamic parameters.

src/app/core/utils/routes.ts
jruddell: Not sure what you mean. 

```
const url = '/my/route/:id/:param'
url.match(/:([0-9_a-z]+)/gi) // [':id', ':param']
```
the match returns an array of matches that includes the `:`. So I'm looping over the array and replacing the colon so that I can look it up in the `params` object sent to this function.

brennancheung: Nevermind, looks like the `g` option makes it so it doesn't return capture groups anymore.

PR #655, opened 2020-01-30
Brigade

brennancheung:


brigade/helpers.js
```diff
@@ -0,0 +1,112 @@
...
+    printPrint(e.payload)
+
+    console.log(`Container registry set to: ${registry}`)
+
+    console.log(JSON.stringify({ prNumber, commit, ref }, null, 4))
+  }
+
+  const createJob = (name, image, tasks = []) => {
+    const fullImage = `${registry}/${image}`
+    const job = new Job(name, fullImage, tasks, true)
```
jruddell: can you make the boolean a variable so we know what it is for?

brigade/helpers.js
```diff
@@ -0,0 +1,112 @@
...
+
+  const logEventParams = () => {
+    // Webhook event received
+    console.log(`Received check_suite for commit ${e.revision.commit}`)
+
+    console.log('event info:')
+    prettyPrint(e)
+
+    console.log('project info:')
+    printPrint(project)
```
jruddell: is this supposed to be `prettyPrint`?

brigade/helpers.js
```diff
@@ -0,0 +1,112 @@
...
+
+    console.log('payload info:')
+    const payload = JSON.parse(e.payload)
+    printPrint(e.payload)
```
jruddell: same here `prettyPrint`?

PR #658, opened 2020-01-31
[PMK-2585] Fixed cluster tagging inconsistencies

kevinylee1991:
Some parts of the UI expected tags to be a list of objects, while others expected it to be a dict. The dict is the correct format.


src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -149,10 +153,12 @@ const createBareOSCluster = async (data = {}, loadFromContext) => {
     'mtuSize',
     'privileged',
     'appCatalogEnabled',
-    'tags',
+    // 'tags',
```
jruddell: comment you can remove

PR #665, opened 2020-01-31
Monitoring

jruddell:
image

image

image

image


src/app/core/utils/renderHelpers.ts
```diff
@@ -0,0 +1,2 @@
+
+export const boolToFormattedText = (value) => value ? 'Enabled' : 'Not Enabled'
```
xantorres: `boolToFormattedText` is too broad of a term, given that this function just returns `Enabled` or `Not Enabled`. 
Either we rename the function or we add parameters to allow specifying what should be returned in each case.

jruddell: what do you want it to be called? its casting a boolean value to a formatted / user friendly name.

brennancheung: We already have `castBoolToStr` in `app/utils/misc.js`.  Also, the text isn't formatted, it's just a literal string in this case.  The "formatted" part doesn't make sense to me.

If we want one specifically for enabled / disabled that we will use in a lot of places then:

`export const enabledStr = castBoolToStr('Enabled', 'Not Enabled')`

Or maybe `Disabled` instead of `Not Enabled` depending on how it is used.

Personally, I would just use `castBoolToStr` and partially apply it locally in the render method.  No need to define a separate function that gets imported just for this.

`<DataRow label="Auto scaling" value={castBoolToStr('Enabled', 'Not Enabled')(data.enableCAS)} />`

PR #672, opened 2020-02-04
[PMK-2587] Add prettier and pre-commit hook to run prettier and eslint before commit

xantorres:
This will run prettier with our eslint settings in js, ts and tsx files.
Please make sure of running `yarn` when updating your branch.


package.json
```diff
@@ -163,5 +166,14 @@
...
-  "repository": "https://github.com/platform9/pf9-ui-plugin.git"
+  "repository": "https://github.com/platform9/pf9-ui-plugin.git",
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{jsx,js,tsx,ts}": "eslint --cache --fix",
+    "*.{js,ts,tsx,css,md}": "prettier --write"
```
jruddell: does the order here matter? running prettier would fix linting issues.

package.json
```diff
@@ -88,11 +88,14 @@
     "faker": "^4.1.0",
     "hard-source-webpack-plugin": "^0.13.1",
     "html-webpack-plugin": "^3.2.0",
+    "husky": "^4.2.1",
```
jruddell: Do we really need two packages for this? what if we checked in a script that is a pre-commit hook? I'd like to avoid excess packages if needed. I wrote a bash script to run lint whenever I commit, happy to share that if we can make that work vs packages.

xantorres: Why reinvent the wheel when these guys have already done a great job with this multi platform package (a dev package so is not affecting build times)

package.json
xantorres: > Wheres the lint configuration? we should have it defined in either package.json as a `"prettier"` section or in a `prettierrc.yaml` file.
> 
> Here's one i've used before.
> 
> ```
> printWidth: 100
> tabWidth: 2
> useTabs: false
> semi: false
> singleQuote: true
> trailingComma: 'all'
> bracketSpacing: true
> jsxBracketSameLine: false
> arrowParens: 'always'
> proseWrap: 'preserve'
> ```

The configuration is taken from `.eslintrc`, as discussed in Slack

package.json
xantorres: Also you would need to run that script to add the pre-commit hook in their git config for every developer, is not so simple to do.

PR #673, opened 2020-02-05
PMK-2605: Corrected the back url links

rahulkurumkar:


src/app/plugins/kubernetes/components/infrastructure/nodes/NodeDetailsPage.tsx
```diff
@@ -55,10 +55,29 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+
+  if (searchParams.has('clusterDetailsNode')) {
+    return routes.cluster.nodes.path({ id: searchParams.get('clusterDetailsNode') })
+  }
+
+  if (searchParams.has('clusterList')) {
+    return routes.cluster.list.path()
+  }
+
+  if (searchParams.has('cloudProviderList')) {
```
jruddell: why are we trying to invent a custom redirect protocol? no one will know how to use this. you need to think about the architecture of what you're trying to build and see how it can be more extensible. Why dont we instead implement a standard `redirect=<url here>` where we can honor that from any view.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
```diff
@@ -58,7 +58,7 @@ const renderLinks = links => {
 
 const renderNodeLink = ({ uuid, name }) => (
   <div key={uuid}>
-    <SimpleLink src={`/ui/kubernetes/infrastructure/nodes/${uuid}`}>
+    <SimpleLink src={`/ui/kubernetes/infrastructure/nodes/${uuid}?clusterList`}>
```
jruddell: use the route helpers here. And this is not a valid url. Query string params are in the form of `key=value`

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/CloudProvidersListPage.js
```diff
@@ -41,7 +41,7 @@ const ClustersCell = ({ clusters }) => {
   </div>
 }
 const renderNodeLink = ({ uuid, name }) => <div key={uuid}>
-  <SimpleLink src={`/ui/kubernetes/infrastructure/nodes/${uuid}`}>{name}</SimpleLink>
+  <SimpleLink src={`/ui/kubernetes/infrastructure/nodes/${uuid}?cloudProviderList`}>{name}</SimpleLink>
```
jruddell: use the route helpers here. And this is not a valid url. Query string params are in the form of `key=value`

src/app/core/components/listTable/ListTable.js
```diff
@@ -391,7 +391,7 @@ class ListTable extends PureComponent {
             : <Radio checked={isSelected} color="primary" />}
         </TableCell>)}
         {this.getSortedVisibleColumns().map(columnDef =>
-          this.renderCell(columnDef, pathStr(columnDef.id, row), row, isSelected, classes.cell),
+          this.renderCell(columnDef, pathStr(columnDef.id, row), row, isSelected, classes.cell, source),
```
jruddell: I'm not a fan of this. the last thing we want right now is to tack on more conditional params to these functions. This source is too custom, we should find a better way to implement with a regular `redirect`

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterDetailsPage.js
```diff
@@ -139,7 +139,7 @@ const ClusterDetailsPage = () => {
         <ClusterStatusAndUsage cluster={cluster} loading={loading} />
         <Tabs>
           <Tab value="nodes" label="Nodes">
-            <ClusterNodes />
+            <ClusterNodes source="clusterDetailsNode" />
```
jruddell: If we were to do this type of redirecting, these sources should be a part of the route helper, so it is better maintained / not random strings in the views. But I don't think we want to implement this source concept

src/app/core/components/listTable/ListTable.js
```diff
@@ -619,6 +619,7 @@ ListTable.propTypes = {
   canDragColumns: PropTypes.bool,
 
   loading: PropTypes.bool,
+  source: PropTypes.string,
```
brennancheung: I don't know what this field does just from the name.  What is it supposed to be used for?  See if there is another name that has less chance of overloaded meanings.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
brennancheung: This is a weird convention for a backlink.  Not sure I like this.  Can't they just hit the back button if they want to go back?

src/app/core/components/listTable/ListTable.js
```diff
@@ -329,7 +329,7 @@ class ListTable extends PureComponent {
     return filters ? this.applyFilters(searchData) : searchData
   }
 
-  renderCell = moize((columnDef, contents, row, isSelected, cellClass) => {
+  renderCell = moize((columnDef, contents, row, isSelected, cellClass, source) => {
```
xantorres: `ListTable` should not know anything about navigation or back url's (if that's what `source` stands for) or anything else out of his main purpose of rendering a table being unaware of the insights of its parents or children.

PR #677, opened 2020-02-06
Rbac

jruddell:
image

image

image


src/app/plugins/kubernetes/components/rbac/AddClusterRoleBindingPage.js
```diff
@@ -30,36 +55,48 @@ export const AddClusterRoleBindingForm = () => {
...
-          info="Select users to assign this role"
-          onChange={getParamsUpdater('users')}
-        />
-        <GroupMultiSelect
-          id="groups"
-          info="Select groups to assign this role"
-          onChange={getParamsUpdater('groups')}
-        />
-        <SubmitButton>Add Cluster Role Binding</SubmitButton>
+        <div className={classes.formWidth}>
```
xantorres: if you pass `fullWidth={false}` to `ValidatedForm ` you avoid the need of creating another container. We could also add a `maxWidth` prop to `ValidatedForm` for cases like this.

src/app/plugins/kubernetes/components/rbac/AddClusterRoleBindingPage.js
```diff
@@ -30,36 +55,48 @@ export const AddClusterRoleBindingForm = () => {
...
+          <FormFieldCard title="Basic Details">
+            <div className={classes.inputWidth}>
```
xantorres: I'm not a fan of this customized one-time use containers, I would rather try to extend our existing components to fit our needs in different situations.

src/app/plugins/kubernetes/components/rbac/AddClusterRoleBindingPage.js
```diff
@@ -30,36 +55,48 @@ export const AddClusterRoleBindingForm = () => {
...
+              />
+              <GroupMultiSelect
+                id="groups"
+                info="Select groups to assign this role"
+                onChange={getParamsUpdater('groups')}
+              />
+            </div>
+          </FormFieldCard>
+        </div>
+        <div className={classes.submit}>
```
xantorres: Same thing here.

src/app/plugins/kubernetes/components/rbac/AddClusterRoleBindingPage.js
jruddell: the card grows as large as its parent. I was wanting to avoid a lot of configurations like max width on validated form, although I can see the benefit. If you guys are fine with adding more into the form I can do that.

src/app/plugins/kubernetes/components/rbac/AddClusterRoleBindingPage.js
jruddell: I did this because inputs have tooltips, and those overlay on top of the card, so I need the inputs to fit inside the card with its tooltip. And the contents vary in size (input vs table).

PR #678, opened 2020-02-06
PMK-2635: Display humanised dates across the app.

rahulkurumkar:


src/app/utils/misc.js
```diff
@@ -1,5 +1,29 @@
 import { path, equals } from 'ramda'
 import moize from 'moize'
+import moment from 'moment'
+
+// Date formatter
+export const defaultDateFormat = 'MMM Do YYYY, hh:mm A'
+export const formatDate = ({ ts, format = defaultDateFormat }) => moment(ts).format(format)
+
+export const durationBetweenDates = ({ ts, start = new Date().valueOf() }) => {
```
xantorres: What's the need for this? The previous function was working perfectly as expected but this one doesn't seem to anymore (I tried it in jsfiddle).
Also please add some unit tests for this (you can add them in `src/app/utils/__tests__/misc.test.js`) and some jsdoc comments explaining what the function does and what parameters expects would be good.

src/app/utils/misc.js
```diff
@@ -1,5 +1,29 @@
...
+   ${duration.minutes()} minutes`)
+  return [
+    { fnName: 'months', label: 'month' },
+    { fnName: 'days', label: 'day' },
+    { fnName: 'hours', label: 'hour' },
+    { fnName: 'minutes', label: 'minute' },
+  ].reduce((displayStr, curr) => {
+    const value = duration[curr.fnName]()
+    if (value > 0) {
+      displayStr += `${displayStr === '' ? '' : ', '}${value} ${curr.label}${value > 1 ? 's' : ''}`
```
jruddell: This line is getting a bit messy, we should probably handle the conditionals separately.

```
const value = duration[curr.fnName]()
if (value < 1) {
  return displayStr
}
const label = value > 1 ? 's' : ''
if (displayStr) {
  displayStr += ', '
}
return `${displayStr}${value} ${label}`

```

src/app/utils/misc.js
xantorres: Ok, I had passed a future date instead of a past date. The params names are misleading because `start` in this case stands for the end date whereas `ts` stands for the initial date.

Moreover, having to pass a date string in that specific is a bit weird, it was easier to pass the seconds as it was being done before or as a Date or Moment object, in my opinion.

src/app/utils/__tests__/misc.test.js
```diff
@@ -24,3 +23,28 @@ describe('sanitizeUrl', () => {
...
+    )
+  })
+  it('get the duration between start date and end date for days hours and minutes', () => {
+    expect(fn('2020-01-27T00:40:30.000Z', '2020-01-29T01:45:36.000Z')).toEqual(
+      '2 days, 1 hour, 5 minutes',
+    )
+  })
+  it('get the duration between start date and end date for n day, 0 hour and 0 minute', () => {
+    expect(fn('2020-01-26T00:40:30.000Z', '2020-01-27T00:40:30.000Z')).toEqual(
+      '1 day, 0 hour, 0 minute',
```
xantorres: I don't think it makes any sense to display this kind of message with 0 hours and 0 minutes, it should be just 1 day in this case, exactly like it was doing in the old function

jruddell: there are already value checks, how does this test even work? I'm fine with keeping the current formatter and not changing it as well.

jruddell: I think it makes sense for us to not change this function at this time. Squeaky wheels get the oil, this particular function works as intended so lets leave it as is.

PR #679, opened 2020-02-06
Bugfix/pmk 2605  added max width to on board node page

rahulkurumkar:


src/app/plugins/kubernetes/components/infrastructure/nodes/DownloadCliPage.tsx
```diff
@@ -23,48 +26,50 @@ const useStyles = makeStyles((theme: Theme) => ({
...
-    >
-      <Typography variant="h5">Onboard a new node</Typography>
-      <p> </p>
-      <Typography component="span">
-        In order to add a physical or virtual node to your BareOS cluster, you need to first
-        download and install the Platform9 CLI on that node. Follow the instructions below to
-        download and install the CLI on your node
-      </Typography>
-      <DownloadCliOnboardNodeWalkthrough />
+    <div className={onBoardNodeContainer}>
```
xantorres: This looks like something we could potentially need in other pages, it would be better to either allow passing a `className` to `<PageContainer />` or add a `maxWidth` prop, both of which will avoid the need to create yet another wrapper container.
In any way, probably having this `max-width` in the PageContainer by default would probably make more sense in this case (we need to be consistent across all pages), avoiding the need of additional props, but something like optionally allowing a `className` prop would not hurt anyway.

jruddell: I dont think page container should have a default max width. The page container is used in a lot of screens, and I dont think we want those to have a max width. This particular page is just a bunch of text so it makes more sense for this one page to be limited. I think a `className` on `PageContainer` is a good idea though

src/app/core/components/pageContainer/PageContainer.js
```diff
@@ -58,7 +59,7 @@ const PageContainer = ({ children, header = undefined, ...rest }) => {
   }, [])
 
   return (
-    <div className={classes.root}>
+    <div className={clsx(className, classes.root)}>
```
jruddell: probably better for className to be the second argument here, so it can override root styles as well if desired

src/app/core/components/pageContainer/PageContainer.js
```diff
@@ -47,7 +48,7 @@ const useStyles = makeStyles((theme) => ({
  * PageContainerHeader to render extra header contents dynamically within any children and also
  * exposes a "header" prop to render any arbitrary fixed header content
  */
-const PageContainer = ({ children, header = undefined, ...rest }) => {
+const PageContainer = ({ children, header = undefined, className = '', ...rest }) => {
```
xantorres: You don't need to default `className` to an empty string, `clsx` is smart enough to ignore `undefined` values

src/app/core/components/pageContainer/PageContainer.js
```diff
@@ -74,6 +75,7 @@ PageContainer.propTypes = {
   header: PropTypes.node,
   // eslint-disable-next-line react/no-unused-prop-types
   floatingHeader: PropTypes.bool,
+  className: PropTypes.string,
```
xantorres: `className` is a default prop in all React components, so you don't need to explicitly define it in the component PropTypes.

rahulkurumkar: Typescript complains if we don't specify this.
image

src/app/core/components/pageContainer/PageContainer.js
rahulkurumkar: Typescript complains about the default value.
image

PR #685, opened 2020-02-06
[PMK-2565 + PMK-2601] Tenants & Users changes and fixes

xantorres:
Also done some major homogeneization and simplification of the form stylings thorough the app.

### DIFF
--- a/src/app/core/components/validatedForm/CheckboxField.js
+++ b/src/app/core/components/validatedForm/CheckboxField.js
@@ -5,11 +5,13 @@ import { FormControl, FormControlLabel, FormHelperText } from '@material-ui/core
 import { withStyles } from '@material-ui/styles'
 import { withInfoTooltip } from 'app/core/components/InfoTooltip'
 import { compose } from 'app/utils/fp'
-import withFormContext, {
-  ValidatedFormInputPropTypes,
-} from 'core/components/validatedForm/withFormContext'
+import withFormContext, { ValidatedFormInputPropTypes } from 'core/components/validatedForm/withFormContext'
 
 const styles = (theme) => ({
+  root: {
+    minWidth: '50%',
+    width: 'fit-content',
+  },
   formControl: {
     marginTop: theme.spacing(1),
   },
@@ -41,7 +43,7 @@ class CheckboxField extends PureComponent {
     } = this.props
 
     return (
-      <div {...restProps}>
+      <div {...restProps} className={classes.root}>
         <FormControl
           id={id}
           onMouseEnter={onMouseEnter}
--- a/src/app/core/components/validatedForm/FormFieldCard.tsx
+++ b/src/app/core/components/validatedForm/FormFieldCard.tsx
@@ -4,17 +4,35 @@ import { makeStyles } from '@material-ui/styles'
 import Theme from 'core/themes/model'
 import clsx from 'clsx'
 
-export const useStyles = makeStyles<Theme>((theme) => ({
+interface ContainerProps {
+  title: string
+  link?: JSX.Element
+  topContent?: JSX.Element
+  className?: string
+  maxWidth?: number
+  inputsWidth?: number
+  limitInputsWidth?: boolean
+}
+const defaultMaxWidth = 715
+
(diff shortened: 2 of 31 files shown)


src/app/core/components/validatedForm/FormFieldCard.tsx
```diff
@@ -4,18 +4,36 @@ import { makeStyles } from '@material-ui/styles'
...
+}
+const defaultMaxWidth = 715
+
+export const useStyles = makeStyles<Theme, ContainerProps>((theme) => ({
   root: {
+    maxWidth: ({ maxWidth = defaultMaxWidth }) => maxWidth,
     borderRadius: 4,
     boxShadow:
-      '0 2.5px 2.5px -1.5px rgba(0, 0, 0, 0.2), 0 1.5px 7px 1px rgba(0, 0, 0, 0.12), 0 4px 5px 0.5px rgba(0, 0, 0, 0.14)',
+      '0 2.5px 2.5px -1.5px rgba(0, 0, 0, 0.2), ' +
```
jruddell: would a string template be more appropriate here?

src/app/core/components/validatedForm/ValidatedForm.js
```diff
@@ -202,13 +207,43 @@ class ValidatedForm extends PureComponent {
...
+          {nowrap ? (
+            inputs
+          ) : (
+            <FormFieldCard
+              title={title}
+              link={link}
+              topContent={topContent}
+              className={className}
+              maxWidth={maxWidth}
+              inputsWidth={inputsWidth}
```
jruddell: how do we handle content of varying size? for instance in this screenshot, the checkbox below the table is half width, the table though is full width. both in the same card
image

src/app/core/components/validatedForm/FormFieldCard.tsx
xantorres: I don't want to have a multilined attribute value here, I just want to be able to see what its meant to be in one line at a glance (as it originally was), so no multilined template is needed here.

src/app/core/components/validatedForm/ValidatedForm.js
xantorres: Thanks for noticing that, I fixed it by changing the input style itself, probably should check other types of inputs too though.

image

src/app/core/components/validatedForm/CheckboxField.js
```diff
@@ -5,11 +5,13 @@ import { FormControl, FormControlLabel, FormHelperText } from '@material-ui/core
...
 import { withInfoTooltip } from 'app/core/components/InfoTooltip'
 import { compose } from 'app/utils/fp'
-import withFormContext, {
-  ValidatedFormInputPropTypes,
-} from 'core/components/validatedForm/withFormContext'
+import withFormContext, { ValidatedFormInputPropTypes } from 'core/components/validatedForm/withFormContext'
 
 const styles = (theme) => ({
+  root: {
+    minWidth: '50%',
```
jruddell: Not a fan of percentage widths. What are we trying to accomplish here? Why not just use a class and pass to the form elements that we want to style differently?

src/app/core/components/validatedForm/ValidatedForm.js
```diff
@@ -202,13 +208,41 @@ class ValidatedForm extends PureComponent {
...
+      debug,
+      id,
+      title,
+      link,
+      className,
+      topContent,
+      formActions,
+      maxWidth,
+      inputsWidth,
+      nowrap,
```
jruddell: We should call this something more specific like `elevated` or `asCard`. nowrap sounds more like wrapping content to a new line `flex-direction: nowrap`

src/app/core/components/validatedForm/CheckboxField.js
xantorres: The 50% width is a way to give enough space to the information tooltip that has to be shown on the right.

src/app/core/components/validatedForm/ValidatedForm.js
xantorres: `elevated` sounds good

PR #686, opened 2020-02-07
PMK-2623: Added default cluster selection to 'All' in Storage, Namespace and RBC pages.

rahulkurumkar:
PMK-2623: Added default cluster selection to 'All' in Storage, Namespace and RBC pages.
Added could provider filed in create storage class flow.

### DIFF
--- a/src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
+++ b/src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
@@ -20,6 +20,7 @@ import { FormFieldCard } from 'core/components/validatedForm/FormFieldCard'
 import FontAwesomeIcon from 'core/components/FontAwesomeIcon'
 import ExternalLink from 'core/components/ExternalLink'
 import { clusterActions } from '../infrastructure/clusters/actions'
+import { CloudProvidersFriendlyName } from './../infrastructure/clusters/model'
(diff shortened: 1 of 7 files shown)


src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
```diff
@@ -105,6 +106,13 @@ const BasicStep = ({ onSubmit, triggerSubmit, wizardContext }) => {
                 onlyHealthyClusters
                 required
               />
+              <TextField
+                id="cloudProvider"
+                label="Cloud provider"
+                info="Cloud provider."
+                disabled
+                value={CloudProvidersFriendlyName[wizardContext.clusterType]}
```
jruddell: Does this work how you expect it to? wizardContext isn't always updated inside a validated form. I'd be careful here.

rahulkurumkar: I don't know it is working but I am getting the correct cloud provider

xantorres: WizardContext should be updated every time we change step or click `Complete`, so I don't see an issue here. 
This was not like this before but now it should work fine due a small change I introduced not long ago in the `withFormContext` HoC to deal with controlled input fields like these.
image

PR #687, opened 2020-02-07
PMK-2619, rework header row for infrastructure tab

jruddell:
image

image

image


src/app/plugins/kubernetes/components/common/ListPageHeader.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <UsageWidget title="Memory" stats={totals.memory} units="GiB" />
+        <UsageWidget title="Storage" stats={totals.disk} units="GiB" />
+      </div>
+    </>
+  )
+}
+
+export default ListPageHeader
+
+const useStyles = makeStyles<Theme>((theme) => ({
```
brennancheung: Should this be moved to top for consistency?

src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPageHeader.tsx
```diff
@@ -0,0 +1,41 @@
...
+const curUsagePaths = {
+  computeCurrPath: 'usage.compute.current',
+  memoryCurrPath: 'usage.memory.current',
+  diskCurrPath: 'usage.disk.current',
+}
+
+export const ClusterListHeader = () => {
+  const handleGetUsage = useCallback((clusters: ICluster[]) => {
+    const items = clusters.filter((cluster) => (cluster.nodes || []).length > 0)
+    return getUsageTotals(items, curUsagePaths, maxUsagePaths)
```
brennancheung: `getUsageTotals` seems like it fits a more general purpose pattern.  Might not need to hard-code the keys if they are already in the paths objects.

src/app/plugins/kubernetes/util/calcUsageTotals.js
```diff
@@ -29,4 +30,16 @@ const calcUsageTotals = curry((items, currentPathStr, maxPathStr) => {
...
-export default calcUsageTotals
+export default calcUsageTotalByPath
+
+export const getUsageTotals = moize(
+  (
+    items,
+    { computeCurrPath, memoryCurrPath, diskCurrPath },
+    { computeMaxPath, memoryMaxPath, diskMaxPath },
+  ) => ({
+    compute: calcUsageTotalByPath(items, computeCurrPath, computeMaxPath),
```
brennancheung: keys might not need to be hard-coded if path objects can already contain the key names.

src/app/plugins/kubernetes/components/common/ListPageHeader.tsx
jruddell: I hate having to scroll past all the styles to get to a component. Maybe we should move them all below the component definition?

src/app/plugins/kubernetes/util/calcUsageTotals.js
jruddell: Hmm, the paths are different based on the type of entity being passed. `nodes.combined.usage`. `clusters.usage`. What do you mean by hard-coded keys? The keys are passed in so they can be variable.

PR #692, opened 2020-02-07
[PMK-2095] Pass step by step segment events for cluster creation wizards

kevinylee1991:
Wanted to attach chris to this PR so he could review the steps, but I'll just slack him instead

### DIFF
--- a/src/app/core/components/wizard/Wizard.js
+++ b/src/app/core/components/wizard/Wizard.js
@@ -60,11 +60,13 @@ class Wizard extends PureComponent {
 
   handleNext = () => {
     const { onComplete } = this.props
-    const { activeStep } = this.state
+    const { steps, activeStep, wizardContext } = this.state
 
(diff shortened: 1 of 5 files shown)


src/app/core/components/wizard/Wizard.js
```diff
@@ -60,12 +60,14 @@ class Wizard extends PureComponent {
...
   handleNext = () => {
     const { onComplete } = this.props
-    const { activeStep } = this.state
+    const { steps, activeStep, wizardContext } = this.state
 
     if (this.nextCb[activeStep] && this.nextCb[activeStep]() === false) {
       return
     }
 
+    if (steps[activeStep].onNext) { steps[activeStep].onNext(wizardContext) }
```
xantorres: You should probably do this after updating the state, just like we do with the `onComplete` callback, otherwise we can run into weird issues if we update state inside of the `onNext` callback.

PR #693, opened 2020-02-07
PMK-2624, Dashboard page work

jruddell:
updated cards to use css grid, mosaic layout for the landing page, put count below header as it feels more natural

image

image

### DIFF
--- a/src/app/plugins/kubernetes/components/dashboard/DashboardPage.tsx
+++ b/src/app/plugins/kubernetes/components/dashboard/DashboardPage.tsx
@@ -34,9 +34,15 @@ export interface IStatusCardWithFilterProps extends StatusCardProps {
 }
 
 const useStyles = makeStyles<Theme>((theme) => ({
-  cardRow: {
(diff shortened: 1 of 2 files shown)


src/app/plugins/kubernetes/components/dashboard/DashboardPage.tsx
```diff
@@ -34,9 +34,15 @@ export interface IStatusCardWithFilterProps extends StatusCardProps {
 }
 
 const useStyles = makeStyles<Theme>((theme) => ({
-  cardRow: {
+  dashboardMosaic: {
     display: 'grid',
     gridTemplateColumns: 'repeat(4, 290px)',
+    gridTemplateAreas: `
```
kevinylee1991: this is cool, I didn't know this was a thing

jruddell: Right!?? You can do some really cool layouts with less html elements! :)

PR #704, opened 2020-02-10
Created stacked area chart component

kevinylee1991:
Looks like below:

Screen Shot 2020-02-10 at 7 42 40 PM

Todos are to create a legend for it and to fit it into the prometheus alarms page.


src/app/core/themes/defaultTheme.js
```diff
@@ -214,6 +214,11 @@ const theme = {
       aws: '#FF9800',
       azure: '#4AA3DF',
     },
+    stackedAreaChart: {
```
jruddell: I'd like to stop making custom theme colors for individual items. Its better if we point at existing types of colors. `palette.succes`, `palette.error`, `palette.warning`

src/app/core/components/graphs/StackedAreaChart.tsx
```diff
@@ -0,0 +1,61 @@
...
+  name: string
+  color: string
+}
+
+// types should be a list of strings that also show up in AreaChartEntry as a property
+interface Props {
+  data: AreaChartEntry[]
+  width: number
+  height: number
+  xAxis: string
```
jruddell: is there a required entry in `data` that has to match the `xAxis` prop?

src/app/core/components/graphs/PieGraph.tsx
```diff
@@ -45,7 +46,7 @@ const PieGraph = ({
         {...rest}
       >
         {items.map((entry, index) => (
-          <Cell key={entry.name} fill={theme.palette.pieChart[entry.color]} />
+          <Cell key={entry.name} fill={path(entry.color.split('.'), theme.palette)} />
```
brennancheung: We have a utility to save yourself needing to do the `split`.

`import { pathStr } from 'utils/fp'`

src/app/core/themes/defaultTheme.js
```diff
@@ -128,6 +128,7 @@ const theme = {
       contrastText: '#FFF',
     },
     warning: {
+      lighter: '#FED16C',
```
jruddell: is `light` too dark so we need this? it would be nice to keep things consistent if we can

PR #706, opened 2020-02-11
[PMK-2613] Fixed bug with cluster switching so that when cluster is c…

kevinylee1991:
…hanged, selected rows that do not exist in the updated list are removed

### DIFF
--- a/src/app/core/components/listTable/ListTable.js
+++ b/src/app/core/components/listTable/ListTable.js
@@ -101,7 +101,9 @@ class ListTable extends PureComponent {
   // only for a subset of above reasons.
   componentDidUpdate(prevProps) {
     const { data } = this.props
-    if (equals(prevProps.data, data)) { return }
+    if (prevProps.data.length === data.length
+      && equals(prevProps.data, data)
+    ) { return }
 
     this.setState({
       selected: [],
--- a/src/app/core/components/listTable/ListTable.js
+++ b/src/app/core/components/listTable/ListTable.js
@@ -13,7 +13,7 @@ import {
   TableRow,
 } from '@material-ui/core'
 import { withStyles } from '@material-ui/styles'
-import { compose, ensureFunction, except, matchArrayProperties } from 'app/utils/fp'
+import { compose, ensureFunction, except } from 'app/utils/fp'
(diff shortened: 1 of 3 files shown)


src/app/core/components/listTable/ListTable.js
```diff
@@ -93,6 +93,16 @@ class ListTable extends PureComponent {
     }
   }
 
+  componentDidUpdate(prevProps) {
+    const { data, uniqueIdentifier } = this.props
+    if (equals(prevProps.data, data)) { return }
```
jruddell: I feel like this could be a more heavy operation when we have lots of entries in the table. I think it makes more sense to drop selected items when we change a cluster. is the `uniqueIdentifier` the cluster id?

kevinylee1991: Yeah, the cluster ID. I'm fine with just dropping all selections

src/app/core/components/listTable/ListTable.js
```diff
@@ -93,6 +93,21 @@ class ListTable extends PureComponent {
...
 
+  // Reset selected when the list of items is updated
+  // This will be the case upon updating the cluster dropdown,
+  // refreshing the list, or if there is any other reason
+  // that the list changes.
+  // In the future we may want to look into triggering this
+  // only for a subset of above reasons.
+  componentDidUpdate(prevProps) {
+    const { data } = this.props
+    if (equals(prevProps.data, data)) { return }
```
jruddell: instead of doing an equals comparison on data which can be a large and expensive operation, would this work if we instead just looked at a change in `uniqueIdentifier`?

src/app/utils/fp.js
```diff
@@ -306,3 +306,11 @@ export const switchCase = (defaultValue, ...cases) => (input) =>
  */
 export const objSwitchCase = (casesObj, defaultValue) => (input) =>
   casesObj.hasOwnProperty(input) ? casesObj[input] : defaultValue
+
+export const matchArrayProperties = (originalArray, comparisonArray, property) => (
```
jruddell: probably dont need this anymore

PR #711, opened 2020-02-14
Notifications

brennancheung:
This has the basic infrastructure for push event notifications in place now.

The context is being updated when clusters are created and deleted but unfortunately the Cluster List Table is not being updated because it does not "react" to changes in the context.

I tried implementing a `setInterval` to `reload` from `context` but that just made it worse because it showed the loading animation over everything and obscured the table.

Unless anyone has any ideas about how to cleanly work around the data loader system I think this is about as far as we can get without the Redux refactor.


src/app/core/providers/NotificationsProvider.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+
+  componentDidMount () {
+    // TODO, it is attempting to establish a websocket but the auth is failing.
+    // How are we supposed to auth this?
+    console.log('componentDidMount')
+    const wsUrl = getWebsocketUrl()
+    this.socket = new WebSocket(wsUrl)
+    this.socket.addEventListener('open', this.handleOpen)
```
jruddell: I would put socket creation / event listeners in an explicit function to handle adding the event listeners. Same with with clean up. `create` and `dispose` would be fine. This way you can call it from `didMount` and `willUnmount` but also when we want to do things like reconnecting... etc.

src/app/core/providers/NotificationsProvider.tsx
```diff
@@ -0,0 +1,90 @@
...
+type UnsubscribeFnType = () => void
+type SubscriberFnType = (event: Object) => UnsubscribeFnType
+
+interface INotificationContext {
+  subscribe?: SubscriberFnType
+}
+
+export const NotificationsContext = createContext<INotificationContext>({})
+
+const getWebsocketUrl = () => {
```
xantorres: This would probably be better as an async function that loads the serviceCatalog (maybe reusing `Keystone.getServiceCatalog`)

src/app/core/providers/NotificationsProvider.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <NotificationsContext.Provider value={{ subscribe: this.subscribe }}>
+        {this.props.children}
+      </NotificationsContext.Provider>
+    )
+  }
+}
+
+export default NotificationsProvider
+
+export const withNotifications = (Component) => (props) => (
```
xantorres: There are already "notifications" in our current app, used to handle the top bar notifications (check `AppProvider`), we should find a different naming for this or for the other stuff or it will be very confusing.

src/app/core/providers/NotificationsProvider.tsx
```diff
@@ -0,0 +1,90 @@
...
+  handleMessage = (event: any) => {
+    const message = event.data
+    this.state.handlers.forEach(handler => handler(message))
+
+    // TODO: handle the messages
+    console.log(message)
+  }
+
+  subscribe (handler: SubscriberFnType) {
+    this.setState(state => ({ handlers: [...state.handlers, handler] }))
```
xantorres: Should we add a unique identifier to the handler? There will be many situations where we will end up with duplicated handlers if we are trying to subscribe from a hook, since the hook might be loaded multiple times.

.eslintrc
```diff
@@ -23,7 +23,7 @@
   ],
 
   "rules": {
-    "max-len": ["warn", { "code": 100, "tabWidth": 2 }],
+    "max-len": 0,
```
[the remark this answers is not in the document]
jruddell: should add a config here for prettier to be at 100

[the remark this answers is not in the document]
xantorres: I think is not a bad thing to have eslint reminding the devs by outlining the lines that they have too long, eg
image
this will make the devs be aware of it and try to fix it and improve readability (strings ad comments don't get fixed by prettier automatically):
image

jruddell:
@xantorres @brennancheung Should we pull the basic socket functionality out of a React Component and put it in a basic Singleton Class like we have for our ApiClient? Want to know your thoughts about that, maybe its ok in a React Component. But I feel like thats limiting potential usage in the future

xantorres:
> @xantorres @brennancheung Should we pull the basic socket functionality out of a React Component and put it in a basic Singleton Class like we have for our ApiClient? Want to know your thoughts about that, maybe its ok in a React Component. But I feel like thats limiting potential usage in the future

The problem I see with that is that we won't have an easy way to remove the listeners when we logout from the app, and we will have to release them manually from the `AuthenticatedContainer`, `LogoutPage` or so. And the same applies for the initialization as well. 
Not a big deal, but is just more things to worry about when logging in and logging out.

Another thing I see is that we won't have a way to prevent devs to subscribe to the service from non-authenticated views, since the singleton will be available from anywhere at any time without restrictions.

jruddell:
> Another thing I see is that we won't have a way to prevent devs to subscribe to the service from non-authenticated views, since the singleton will be available from anywhere at any time without restrictions.

This is the same as the ApiClient then. What prevents a dev from calling it? Also we could sync the socket instance with authentication.

Logout should dispose the socket instance because that action makes the socket invalid.

.eslintrc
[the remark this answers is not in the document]
jruddell: I too like to use the problems tab, and eslint showing me all these items is not ideal. I don't mind having the reminder, but the thing is prettier usually fixes that for me when I save. If we have edge cases where lines go long it's a bit annoying to be poked about that constantly. Either we should fix all of the warnings or turn it off IMO.

xantorres:
> > Another thing I see is that we won't have a way to prevent devs to subscribe to the service from non-authenticated views, since the singleton will be available from anywhere at any time without restrictions.
> 
> This is the same as the ApiClient then. What prevents a dev from calling it? Also we could sync the socket instance with authentication.
> 
> Logout should dispose the socket instance because that action makes the socket invalid.

With the ApiClient you don't want to prevent the devs from calling it from anywhere, because you want to be able to perform api calls even if the user is not logged in.

.eslintrc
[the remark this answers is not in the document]
xantorres: I vote for fixing all these cases then, and keep the rule enabled.

jruddell:
> With the ApiClient you don't want to prevent the devs from calling it from anywhere, because you want to be able to perform api calls even if the user is not logged in.

When there is no token the socket should throw an exception, the API client should do that as well for requests that are behind an authentication wall, I don't think it makes sense for us to tie socket management inside a react component, If I want to access the socket instance inside an API call, or if I want to dispose it in an action, right now it has to be done from a react component. This is limiting in the same way we do our state management. What happens when we refactor to redux? how would we access the socket instance in actions? Seems like it makes a lot of sense to write this in a way where we don't have to refactor it when moving to redux. Maybe i'm not seeing the same perspective you are, how would you handle this?

.eslintrc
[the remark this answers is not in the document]
brennancheung: Problem is it is warning me about things that are not my code that have been there from long before.  Hard to see any actual errors when all you see is lint warnings about length for a file you haven't even touched.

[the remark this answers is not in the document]
brennancheung: Also, we usually are pretty good about writing readable code and even if it goes over the length, it's usually for a good reason.  Also, prettier is going to eventually get it if we modify that file in the future.  Don't really see much benefit given that it is slow doing normal development.

src/app/core/providers/NotificationsProvider.tsx
brennancheung: I was looking for an action that does that but looks like it uses the current code uses the old system of data loading.  Since we are planning on swapping out a lot of the data loading stuff for redux I didn't want to add any more code to that system.

src/app/core/providers/NotificationsProvider.tsx
brennancheung: Definitely agree with this point.  PushEvents?

jruddell: I like `PushEvents` that makes more sense

src/app/core/PushEventManager.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+type SubscriberFnType = (event: Object) => void
+
+const getWebSocketUrl = () => {
+  const client = ApiClient.getInstance()
+  const catalog = client.serviceCatalog
+
+  if (!catalog) return console.error('Error loading serviceCatalog from ApiClient')
+
+  const notificationService = catalog.find(x => x.type === 'notifications')
```
jruddell: Should we use an enum for the catalog types? I could see us having other types in the future, but I'm not sure.

src/app/core/PushEventManager.tsx
```diff
@@ -0,0 +1,75 @@
...
+  if (!notificationService) return console.error('notification service not found in keystone service catalog')
```
jruddell: Line 15 could be a problem, maybe `if (!notificationService || !notificationService.endpoints || !notificationService.endpoints.length)` 

or we could maybe use `path` here? `notificationService.endpoints[0].url`

jruddell:
Fixed a few typings. 
image

.eslintrc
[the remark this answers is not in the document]
xantorres: As I said, if we remove this rule, all files from now on will be reformatted to 80 lines instead of the current 100, so if you remove this rule then make sure to add a `.prettierrc` with a `print-width` value of 100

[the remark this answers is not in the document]
jruddell: @brennancheung @xantorres manually fixing long strings and having to maintain that isn't the greatest. Ideally I'd not want to worry about formatting at all and just let the formatter handle that. Instead of removing the eslint `max-len` property, what about just telling webpack to ignore warnings? https://github.com/webpack/webpack.js.org/issues/147#issuecomment-253736058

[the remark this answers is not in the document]
brennancheung: Looks like the issue happens on `master` as well so I'm just going to leave figuring out a solution to this in another PR.

src/app/core/PushEventManager.tsx
```diff
@@ -0,0 +1,83 @@
...
+    }
+    return PushEventManager.instance
+  }
+
+  handleOpen() {
+    console.info('Websocket connection opened')
+  }
+
+  handleClose() {
+    console.info('Websocket connection closed')
```
jruddell: Do we want to do anything here around socket connection retries?

src/app/core/providers/PushEventsProvider.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const catchAll = (payload) => {
+      console.info(`No push event handler found for qbert::cluster_status with action: ${action}`)
+    }
+    const handler = actionHandlers[action] || catchAll
+    handler(payload)
+  }
+
+  const handleClusterStatusCreate = (payload) => {
+    const clusterUuid = payload[0].source
+    console.log(`Cluster created with uuid: ${clusterUuid}`)
```
jruddell: probably should do a pass through and remove the log messages

src/app/core/PushEventManager.tsx
brennancheung: Possibly.  But maybe in another PR.  We can either do  simple delay like 10 seconds or do an exponential backoff.

src/app/core/providers/PushEventsProvider.tsx
brennancheung: I wanted to leave these particular ones in place for now, at least until we have the data loading system refactored to redux.  It will be hard to know when the events are coming in if we don't have these messages.  I have the noisier ones from `cluster_update` already silenced.

PR #714, opened 2020-02-14
PMK-2668: Initial code check-in for button component.

rahulkurumkar:
PMK-2668: Initial code check-in for button component.


src/app/utils/__tests__/buttons/PrevButton.test.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
```
jruddell: These tests should be in typescript.

src/app/utils/__tests__/buttons/PrevButton.test.js
```diff
@@ -0,0 +1,46 @@
...
+import { ThemeProvider } from '@material-ui/core/styles'
+import { createShallow } from '@material-ui/core/test-utils'
+
+describe('<PrevButton />', () => {
+  let wrapper
+  let shallow
+  let mockCallBack
+
+  beforeAll(() => {
+    shallow = createShallow()
```
jruddell: We shouldn't be using shallow. Avoid using it as much as possible please. I'd recommend we bring in `@testing-library/react` and `import { render, fireEvent }` from there to do things like rendering and triggering a click.

src/app/core/components/buttons/PrevButton.js
```diff
@@ -17,7 +17,7 @@ const styles = (theme) => ({
 const PrevButton = ({ children, classes, disabled, ...rest }) => {
   const params = {
     className: classes.baseButton,
-    color: disabled ? 'primary' : 'primary',
+    color: disabled ? 'secondary' : 'primary',
```
jruddell: PrevButton here can render `test-ids` that we can use for testing purposes. 

```
return (
    <Button {...params}  data-test-id="test-prev-button">
      <Icon className={classes.leftIcon} data-test-id="test-prev-button-icon">arrow_back</Icon>
      {children || 'Back'}
    </Button>
  )
```

and then when testing you should be able to `const { getByTestId } = render(PrevButton)`
I haven't tested this so might have to tweak it some. But this would be a much better approach. And you are writing more valid tests. Shallow can pass even when the component is broken.

src/app/utils/__tests__/buttons/PrevButton.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  afterEach(cleanup)
+
+  test('render', () => {
+    const { getByTestId } = render(
+      <TestThemeProvider>
+        <PrevButton>Back Button</PrevButton>
+      </TestThemeProvider>,
+    )
+    expect(getByTestId('test-prev-button')).toHaveTextContent('Back Button')
```
jruddell: These test id's should be constants or better yet enums.

src/app/utils/__tests__/buttons/PrevButton.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+  })
+
+  test('disabled', () => {
+    const { getByTestId } = render(
+      <TestThemeProvider>
+        <PrevButton disabled>Back Button</PrevButton>
+      </TestThemeProvider>,
+    )
+
+    fireEvent.click(getByTestId('test-prev-button'))
```
jruddell: you should just alias / make a const on getByTestId instead of calling it twice.

src/app/utils/__tests__/buttons/PrevButton.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+      </TestThemeProvider>,
+    )
+
+    const parent = getByTestId('test-prev-button')
+    const child = getByTestId('test-prev-button-icon')
+    expect(parent).toContainElement(child)
+  })
+})
+
+// import React from 'react'
```
jruddell: clean up comments when you're done with them

src/app/utils/__tests__/buttons/SubmitButton.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <TestThemeProvider>
+        <SubmitButton disabled>Submit Button</SubmitButton>
+      </TestThemeProvider>,
+    )
+
+    fireEvent.click(getByTestId('test-submit-button'))
+    expect(getByTestId('test-submit-button')).toBeDisabled()
+  })
+
+  test('type=submit', () => {
```
jruddell: I think its better to have a more concise / specific test. Like one that tests the attributes is fine, but that can be a part of `render`. individual tests for attributes is going to add a lot of tests, which will take more time. Lets put these as an initial test for the default properties. Tests for different states like `disabled` for example are a good example of a more concise / specific test for functionality.

package.json
```diff
@@ -9,6 +9,8 @@
     "@material-ui/icons": "^4.2.1",
     "@material-ui/lab": "^4.0.0-alpha.20",
     "@material-ui/styles": "^4.2.1",
+    "@testing-library/jest-dom": "^5.1.1",
```
jruddell: shouldn't these be dev dependencies?

package.json
```diff
@@ -155,7 +157,7 @@
     "setupFilesAfterEnv": [
       "./setupJest.js"
     ],
-    "testEnvironment": "node",
+    "testEnvironment": "jsdom",
```
jruddell: this is going to break our existing tests.

src/app/core/components/buttons/CancelButton.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import Button from '@material-ui/core/Button'
 import { withStyles } from '@material-ui/styles'
+import { TestId } from 'utils/testId'
```
jruddell: I think it makes more sense for each component to export its own test id's

src/app/utils/__tests__/buttons/CancelButton.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+  })
+
+  test('disabled', () => {
+    const { getByTestId } = render(
+      <TestThemeProvider>
+        <CancelButton disabled>{text}n</CancelButton>
+      </TestThemeProvider>,
+    )
+
+    expect(getByTestId(TestId.TEST_CANCEL_BUTTON)).toBeDisabled()
```
jruddell: can we also verify that clicks / interactions on the button dont fire as a part of a disabled test?

src/app/utils/testId.ts
```diff
@@ -0,0 +1,26 @@
+export enum TestId {
+  // SubmitButton
+  TEST_SUBMIT_BUTTON = 'test-submit-button',
```
jruddell: These should be in each individual file. `import { PreviousButtonTestIds } from 'core/components/buttons/PreviousButton'`.

Enum casing should follow the rest of our enums (PascalCase).

```
export enum PreviousButtonTestIds {
  Button = 'prev-button',
  Icon = 'prev-button-icon'
}
```

src/app/utils/__tests__/buttons/NextButton.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+      </TestThemeProvider>,
+    )
+
+    expect(getByTestId(TestId.TEST_NEXT_BUTTON)).toHaveTextContent('Next')
+  })
+
+  test('render forward icon', () => {
+    const { getByTestId } = render(
+      <TestThemeProvider>
+        <NextButton showForward />
```
jruddell: isn't the default behavior that it shows the forward icon? we should put ALL default parameter testing into one test. the `render` test here is a good place for that. this test should rather pass false as the value for `showForward` to make sure it doesn't show up anymore when we pass false through

src/app/utils/__tests__/buttons/CloseButton.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+          <CloseButton to="backtoList">{text}</CloseButton>
+        </Router>
+      </TestThemeProvider>,
+    )
+
+    const parent = container.firstElementChild
+    const child = getByTestId(TestId.TEST_CLOSE_BUTTON_LINK)
+    const grandChild = getByTestId(TestId.TEST_CLOSE_BUTTON_ICON)
+
+    expect(parent).toContainElement(child)
```
jruddell: This is probably ok, but it also would be fine to verify that the element exists. Testing that its a parent to child may not be the best thing because if we change the order of rendering we will have to update the test. When really all this test is doing is verifying that the element was rendered... 

To that note, maybe you should look at these tests and make sure you're actually testing what the title of the test says. In this one I would expect that somewhere you verify that the link has a specific route that you passed through (**Especially since this test is supposed to be doing just that**, testing the `to` property).

package.json
brennancheung: ☝️

package.json
```diff
@@ -106,6 +106,8 @@
     "storybook-react-router": "^1.0.5",
     "style-loader": "^0.23.1",
     "terser-webpack-plugin": "^1.3.0",
+    "@testing-library/jest-dom": "^5.1.1",
```
brennancheung: We already have Cypress.  Let's stick with that for anything DOM related.

src/app/core/components/buttons/CardButton.js
```diff
@@ -23,8 +24,13 @@ const styles = (theme) => ({
 })
 
 const CardButton = ({ children, classes, showPlus = true, ...rest }) => (
-  <Button className={classes.cardButton} variant="contained" {...rest}>
-    {showPlus ? '+' : ''}
+  <Button
+    data-testid={TestId.TEST_CARD_BUTTON}
```
brennancheung: Can we just use a hard-coded string from this?  Or a string prefix with an index for collections.  Having an extra import clutters the code and adds overhead.

src/app/utils/TestThemeProvider.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+
+import { ThemeProvider, createMuiTheme } from '@material-ui/core/styles'
```
brennancheung: I don't want a separate code path for tests.  We will miss things if what we are testing and what we are shipping diverges.  We should be testing the actual app itself and go through Cypress for anything DOM related.

brennancheung: Also, we already have Storybook which can host the components in a more isolated environment.  We might be able to target those.

package.json
jruddell: This is for unit testing, I asked that we bring it in when we write unit tests vs shallow rendering.

jruddell:
I think we should hold off on unit tests for now, I'll setup a few example ones that we can use as a template in the future when we're ready to write them. Lets put the focus on integration tests for the time being.

PR #719, opened 2020-02-20
update wording to be more inline with k8s nomenclature

jruddell:


src/app/plugins/kubernetes/components/infrastructure/clusters/AddAwsClusterPage.js
```diff
@@ -534,7 +534,7 @@ const AddAwsClusterPage = () => {
                             {/* Workloads on masters */}
                             <CheckboxField
                               id="allowWorkloadsOnMaster"
-                              label="Make all Master nodes Master + Worker"
+                              label="Enable workloads on all master nodes"
```
brennancheung: This is roughly what it originally was and they wanted to change it to what is on on the left.  I prefer "enable workloads" so I'm good with this.

jruddell: Yep, people found it confusing so were changing it back. I like enable better than allow and we use enabled / disabled terminology everywhere else so it made sense to change it to this.

PR #724, opened 2020-02-24
PMK-2582, updates to cluster health status

jruddell:
update cluster status messages to show a more detailed breakdown of counts when appropriate

image

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatusUtils.ts
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatusUtils.ts
@@ -172,6 +172,18 @@ const getFormattedCounts = (masterNodes, workerNodes) => ({
   ),
 })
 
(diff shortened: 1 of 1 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatusUtils.ts
```diff
@@ -26,15 +27,15 @@ interface HealthStatusAndMessage {
   mastersHealthStatus: HealthStatus
   workersHealthStatus: HealthStatus
   healthStatus: HealthStatus
-  message: string
+  getMessage: (masters: INodeCount, workers: INodeCount) => string
```
brennancheung: Maybe `formatMessage` to remove confusion since nothing is being fetched here.

jruddell: used to just be a message property, now its a function that returns the message. I think `getMessage` is good enough.

PR #726, opened 2020-02-25
Monitoring dashboard page

kevinylee1991:
Read-only monitoring dashboard only


src/app/plugins/kubernetes/index.js
```diff
@@ -392,6 +398,7 @@ Kubernetes.registerPlugin = (pluginManager) => {
     { name: 'Storage Classes', icon: 'hdd', ...clarityLink('/kubernetes/storage_classes') },
     { name: 'Namespaces', icon: 'object-group', ...clarityLink('/kubernetes/namespaces') },
     { name: 'Prometheus Monitoring (BETA)', icon: 'chart-area', link: { path: '/prometheus' } },
+    { name: 'Monitoring', icon: 'chart-area', link: { path: '/alarms' } },
```
jruddell: is this not the same as the tab on the line above?

kevinylee1991: just a wip at the moment, but we def need to ask chris/madhura about if we still want to expose prometheus monitoring (beta)

kevinylee1991:
Screen Shot 2020-02-27 at 11 36 10 PM

src/app/plugins/kubernetes/components/alarms/SeverityPicklist.js
```diff
@@ -0,0 +1,50 @@
...
+        {...rest}
+        ref={ref}
+        onChange={onChange}
+        options={options}
+      />
+    )
+  },
+)
+
+SeverityPicklist.propTypes = {
```
jruddell: Should be typescript, we don't need to define propTypes anymore. And the default props should just be inline when you destructure properties

src/app/plugins/kubernetes/components/alarms/actions.js
```diff
@@ -0,0 +1,187 @@
...
+  },
+)
+
+// Maybe in the future we can add interval variable and make the
+// period & intervals dynamic, but for now just preset periods
+// and 7 intervals.
+const getTimestamps = (startTime, period) => {
+  const momentObj = moment.unix(startTime)
+
+  if (period === '24.h') {
```
jruddell: Need to clean this up. Manual mapping with `if` / `else` statements doesn't pass code smell. Think about how you're writing your code. If you're copy pasting something you probably should find a better way to do it. Here each of these are really similar, you could define an object mapping for these params

```
const timestampMutators = {
  // values are the arguments to pass to `moment.add`
  '12.h': [2, 'hours'],
  '6.h': [1, 'hours'],
  '3.h': [30, 'minutes'],
 ...
}

const getTimestamps = (startTime, period) => {
  const momentObj = moment.unix(startTime)
  const momentArgs = timestampMutators[period]
  return [momentObj.unix(), ...Array(6).fill().map(() => momentObj.add(...momentArgs).unix()]
}
```

src/app/plugins/kubernetes/components/alarms/actions.js
```diff
@@ -0,0 +1,145 @@
...
+    },
+    entityName: 'Alert',
+    uniqueIdentifier: 'id',
+    indexBy: 'clusterId',
+  },
+)
+
+const timestampSteps = {
+  // for use in moment.add
+  '24.h': [4, 'h'],
```
brennancheung: I'm confused.  24h is different than 4h.  Perhaps explain what's going on here.  Add comment about what the keys / values represent.

src/app/plugins/kubernetes/components/alarms/actions.js
```diff
@@ -0,0 +1,145 @@
...
+  '1.h': [10, 'm'],
+}
+
+// Maybe in the future we can add interval variable and make the
+// period & intervals dynamic, but for now just use preset periods
+// and 7 intervals.
+const getTimestamps = (startTime, period) => {
+  const momentObj = moment.unix(startTime)
+  const momentArgs = timestampSteps[period]
+  return [momentObj.unix(), ...Array(6).fill().map(() => momentObj.add(...momentArgs).unix())]
```
brennancheung: I have to play compiler in my head to understand what is going on here.  Add a comment showing what the resultant array is going to look like so I can tell what this is doing at a quicker glance.

src/app/plugins/kubernetes/components/alarms/actions.js
```diff
@@ -0,0 +1,145 @@
...
+      warning: 0,
+      critical: 0,
+      fatal: 0,
+    }
+    return accum
+  }, {})
+
+  const severityCountsByTimestamp = alertData.reduce((accum, current) => {
+    const severity = current.metric.severity
+    if (['warning', 'critical', 'fatal'].includes(severity)) {
```
brennancheung: `filter` this out before it even gets to the `reduce`.

src/app/plugins/kubernetes/components/alarms/actions.js
```diff
@@ -0,0 +1,145 @@
...
+      for (const dataPoint of current.values) {
```
brennancheung: `for` + `if` can be more idiomatically represented using a `filter`

src/app/plugins/kubernetes/components/alarms/actions.js
```diff
@@ -0,0 +1,145 @@
...
+
+    // Still need to calculate this manually as well because there
+    // is a chance that if no alerts were firing during that time,
+    // the API will not return those timestamps.
+    // Use unix timestamp to match the prometheus API
+    const timestamps = getTimestamps(timePast, chartTime)
+
+    const clusters = await loadFromContext(clustersCacheKey, { prometheusClusters: true })
+    if (chartClusterId === allKey) {
+      const all = await someAsync(pluck('uuid', clusters).map((cluster) => (
```
brennancheung: Javascript has a native `flatMap` since doing a `map` and then a `flat` are so common.

src/app/plugins/kubernetes/components/alarms/actions.js
kevinylee1991: In this case I create a starting template using the list of timestamps because the alertData may not contain data covering all these timestamps. I'm not sure the best way I could use filter in this particular case

src/app/plugins/kubernetes/components/alarms/actions.js
kevinylee1991: Tried converting to flatMap but it needs to be flattened twice so just converting this to flatMap does not work

src/app/plugins/kubernetes/components/alarms/actions.js
jruddell: I think this is ok

jruddell:
Lets refactor these to typescript, but do that in a following PR. I'd like to merge this so we can run a build including these changes.

PR #729, opened 2020-02-26
PushEvent retry and misc fixes

brennancheung:
This adds a 10 second delay and then retry when the websocket is closed.  Hopefully this won't hit the backend too hard.  If it turns out to be an issue we can add an exponential backoff.

This also fixes some misc issues with invalid message parsing.

### DIFF
--- a/src/app/core/PushEventManager.tsx
+++ b/src/app/core/PushEventManager.tsx
@@ -40,6 +40,13 @@ class PushEventManager {
   }
 
   handleClose() {
+    const RETRY_INTERVAL_SECONDS = 10
+
+    setTimeout(() => {
+      console.info(`PushEvent websocket closed, retrying in ${RETRY_INTERVAL_SECONDS} seconds.`)
+      this.disconnect()
+      this.connect()
+    }, RETRY_INTERVAL_SECONDS * 1000)
     console.info('Websocket connection closed')
   }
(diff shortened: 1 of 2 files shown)


src/app/core/PushEventManager.tsx
```diff
@@ -40,6 +40,13 @@ class PushEventManager {
   }
 
   handleClose() {
+    const RETRY_INTERVAL_SECONDS = 10
+
+    setTimeout(() => {
```
jruddell: Should this be an interval instead of a timeout?

brennancheung: Yeah, might make sense to do that in case there are network issues.  I'll need to add logic to remove the interval once we connect.

src/app/core/PushEventManager.tsx
```diff
@@ -37,9 +38,20 @@ class PushEventManager {
...
   handleOpen() {
     console.info('Websocket connection opened')
+    if (this.clearIntervalId) {
+      clearInterval(this.clearIntervalId)
+      this.clearIntervalId = null
+    }
   }
 
   handleClose() {
+    const RETRY_INTERVAL_SECONDS = 10
```
jruddell: I'd put the constant outside of the class

PR #731, opened 2020-02-27
PMK-2685: Initial code check-in for integration test cases

rahulkurumkar:


src/app/core/components/CRUDListContainer.js
```diff
@@ -119,9 +119,11 @@ const CRUDListContainer = ({
       {addEnabled && (
         <PageContainerHeader>
           {AddButton ? (
-            <AddButton onClick={handleAdd} />
+            <AddButton data-testid="list-container-add-btn" onClick={handleAdd} />
```
brennancheung: It would be more consistent as `data-test-id` instead of `data-testid`.  But with how much we will need to use this I'm wondering if we should use something even shorter.  Anyone have suggestions for a shorter name?  @xantorres @jruddell

src/app/plugins/kubernetes/components/userManagement/users/AddUserPage.js
```diff
@@ -83,14 +98,24 @@ const AddUserPage = () => {
...
                             label={activationByEmailLabel}
                           />
                         }
                         <br />
                         <FormControlLabel
                           value="createPassword"
-                          control={<Radio color="primary" />}
+                          control={
+                            <Radio
+                              data-testid="add-user-page-basic-info-activation-by-password"
```
brennancheung: I'm on the fence about these really long id's.  The BEM style makes it obvious where it is but are we really going to have that many issues?

Also we don't need id's on everything.  Cypress makes it very easy to find things using nested commands so if you find the containing field by it's `columnId` we can eliminate adding the id's everywhere.  This all depends on whether elements are being repeated within the same containing component but I don't want to see the normal code littered with test artifacts unnecessarily.

cypress/integration/login.spec.js
```diff
@@ -1,43 +1,11 @@
...
-      .click()
-
-    cy.contains(config.region)
-  })
-
-  it('remembers the login state on refresh', () => {
-    cy.login()
-    cy.visit('/')
-    cy.contains(config.region)
+    cy.get('[data-testid=login-email]').type(Cypress.env('userName'))
```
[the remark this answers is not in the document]
rahulkurumkar: @brennancheung, I have already added it in command but to test login as well I have created this spec file. If you look at other placed I am already using a login from command.
I have used `data-tesid` so that in future if we want to do the unit testing using Jest we can reuse the same id. `.id()` and `.testId()` are not supported by cypress

cypress/support/commands.js
```diff
@@ -1,97 +1,55 @@
...
-// The hard-coded token id is explicitly whitelisted in the simulator.
 Cypress.Commands.add('login', () => {
-  // FIXME we should be using cy.fixture instead of importing the fixture manually
-  window.localStorage.setItem('pf9', JSON.stringify(sessionJson))
+  cy.visit('/')
+  cy.get('[data-testid=login-email]').type(Cypress.env('userName'))
+  cy.get('[data-testid=login-password]').type(Cypress.env('password'))
+  cy.get('[data-testid=login-submit-btn]').click()
+  // For dev testing it is taking time to login
+  cy.wait(5000)
```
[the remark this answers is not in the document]
rahulkurumkar: @brennancheung, In development or from my side it takes time to log in and if I don't specify the explicit delay then test cases are failing. Please find the attached snapshot without delay
image

src/app/core/components/CRUDListContainer.js
rahulkurumkar: @brennancheung I have used `data-tesid` so that in future if we want to do the unit testing using Jest we can reuse the same id.

cypress/integration/login.spec.js
```diff
@@ -1,43 +1,11 @@
...
+    cy.get('[data-testid=login-password]').type(Cypress.env('password'))
+    cy.get('[data-testid=login-submit-btn]').click()
+    // For dev testing it is taking time to login
+    cy.wait(5000)
```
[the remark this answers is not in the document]
rahulkurumkar: @brennancheung , Cypress won't wait till the assertion is successful. Cypress up to 4 seconds for it to exist in the DOM
https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Timeouts
I have modified the timeout for some special cases.

cypress/support/commands.js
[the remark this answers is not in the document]
rahulkurumkar: @brennancheung , It waits up to 4 sec. I have modified the timeout instead of adding `.wait`.
https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Applying-Timeouts

cypress/integration/logout.spec.js
```diff
@@ -0,0 +1,17 @@
+describe('Logout user', () => {
+  before(() => {
+    cy.login()
+  })
+
+  it('logout user from application', () => {
+    cy.get('[data-testid=user-menu]')
+      .should('contain', Cypress.env('userName'))
```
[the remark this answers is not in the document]
rahulkurumkar: Done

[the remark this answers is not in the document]
rahulkurumkar: Done

cypress/support/commands.js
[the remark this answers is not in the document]
jruddell: I dont think we should ever be delaying. This has no guarantee that the thing will be complete and will slow down testing dramatically. Is there a way we can have a callback or something of that sort to handle the testing cases?

src/app/core/components/CRUDListContainer.js
jruddell: If were going this route we should not be hardcoding these in the elements. We should have a constant mapping so we can reuse / refactor as desired. I would prefer `data-test-id` in terms of format. 

If were moving the text content to be constants for different locals in the future we could possibly do away with the test id's and just get by the content. What do you guys think?

src/app/plugins/kubernetes/components/userManagement/users/AddUserPage.js
jruddell: i am also in agreement, this is too long of a test id. or well its unnecessarily too long. If these are not hardcoded on each element you can reduce the length by a decent margin. Meaning, if you are defining these as constants in each respective file you really dont need all of the path in the name. Lets think about how were approaching this, take a step back and view the big picture, this is currently quite messy, how would you like to read your own code in 6 months?

cypress.json
```diff
@@ -2,5 +2,13 @@
   "baseUrl": "http://localhost:3000",
   "viewportWidth": 1920,
   "viewportHeight": 1080,
-  "projectId": "k5ua2x"
+  "projectId": "k5ua2x",
+  "env": {
+    "userName": "test-user@platform9.com",
+    "password": "TestPassword@123",
+    "users": {
+      "userName": "__TestUser__",
```
jruddell: what are these underscores for?

cypress/integration/login.spec.js
```diff
@@ -1,43 +1,11 @@
-import config from '../../config'
-
-const signInText = 'Sign In'
-const username = 'admin@platform9.com'
-const password = 'secret'
-
-describe('login', () => {
-  it('reports failed logins', () => {
+describe('Login', () => {
+  it('login into application', () => {
```
jruddell: If you're going to change the name of a test please make sure it reads as plain English. `it login to application` doesn't make sense, did you mean something more like `it authenticates the client`?

cypress/integration/login.spec.js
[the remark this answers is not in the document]
jruddell: timeouts are generally a bad way to write a test, can you think of a better way to do this?

cypress/integration/users.spec.js
```diff
@@ -1,83 +1,69 @@
...
 
-    // TODO We should try to login with the newly created user
-  })
-
-  context('edit user', () => {
-    it('shows the edit user dialog', () => {
-      cy.contains('tr', 'User#Test').click()
-      cy.contains('div', 'Edit').click()
-      cy.contains('Edit User User#Test')
+    it('confirm newly created user', () => {
```
jruddell: `it validates a user was created`

src/app/plugins/kubernetes/components/userManagement/users/AddUserPage.js
rahulkurumkar: My approach for the naming convention is `component-name-form-context-element-name`

cypress.json
rahulkurumkar: @jruddell I have used this nomenclature so that in case of some test case failure we can clean up the data. e.g If deleting user test case fails then this user will be present in the system and we can clean up such data from the test environment.

src/app/plugins/kubernetes/components/userManagement/users/AddUserPage.js
jruddell: you don't need that explicit of a name if these were variables is what i'm saying. Its not necessary and overly verbose. As I've said a few times now, if you were to define these as constants in each file, then we can worry less about refactoring, how verbose the name is... etc. 

```
const testIds = {
  activateByEmail: 'activation-by-email'
  ...
}

data-test-id={activateByEmail}
```
I find this to be a much more readable implementation. And we can use it in our tests as well.

cypress.json
jruddell: These tests shouldn't be writing to a database.

jruddell:
@rahulkurumkar any updates on this?

src/app/core/components/UserMenu.js
```diff
@@ -57,7 +57,12 @@ class UserMenu extends React.PureComponent {
           anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
         >
           {false && <MenuItem onClick={this.handleChangePassword}>Change Password</MenuItem>}
-          <MenuItem data-testid="user-menu-sign-out" onClick={this.logout}>
+          <SimpleLink src={helpUrl} className={classes.link}>
+            <MenuListItem data-test-id="user-menu-sign-out" icon="question-circle">
```
jruddell: this test id is for sign out but its on the help link

PR #733, opened 2020-02-27
[PMK-2654] Prefill user name from link on user password reset page

xantorres:


src/app/core/public/ResetPasswordPage.tsx
```diff
@@ -154,10 +154,12 @@ const renderPasswordValidationCheck: IPasswordValidationCheck = (passwordValue)
...
 
 const ResetPasswordPage: React.FC = () => {
-  const { params, updateParams, getParamsUpdater } = useParams(initialState)
   const classes = useStyles({})
   const { history, match } = useReactRouter()
-
+  const { params, updateParams, getParamsUpdater } = useParams({
+    ...initialState,
+    emailId: match.params.id
+  })
```
rahulkurumkar: I don't think this will work. AFAIK for reset password API we need to send the unique id along with new user credential and backend was not return the email id and OTP instead it was returning the weird URL which is having unique id.

jruddell: In the ticket the querystring parameter is `username` (`?username=<email>&otp=<otp>`). Shouldn't you be using `UrlSearchParams` here to get the `username` from the search and populate the form?

xantorres: @jruddell is not that simple, the link `https://<du.url>/ui/users/activate?username=<email>&otp=<otp>` redirects to the `ActivateUserPage`, which in turns redirects the user to the password reset page after the username and otp has been validated.
I changed this redirection to also send the user email, so that we can show it in the input, since as @rahulkurumkar pointed out, the generated ID is not the same thing as the email.

src/app/core/utils/routes.ts
```diff
@@ -157,6 +157,6 @@ export const routes = {
     ),
   },
   password: {
-    reset: new Route<{ id: string }>(`${appUrlRoot}/reset/password/:id`),
+    reset: new Route<{ id: string; username: string }>(`${appUrlRoot}/reset/password/:id/:username`),
```
jruddell: not a fan of adding an additional `/:username` here. if you omit this param then the `username` will be appended as querystring parameters. I feel thats a better way to implement this.

PR #734, opened 2020-02-28
Auto deploy static content to target DUs after build

ParagDoke:
https://platform9.atlassian.net/browse/TESTS-1134

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build/*
+yarn/*
 output/*
 dist/*
 plugins/*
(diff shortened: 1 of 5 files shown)


ParagDoke:
> How is this supposed to be used? can you add a script in the package.json scripts section called `deploy:dev` with the execution script as the value?

IMO, since we build this on TeamCity as well as Travis CI, did not want to introduce this within `package.json`. Instead, planned to introduce a new build step on TeamCity with ui-dev FQDN being the default param value for `TARGET_DUS`.

PR #740, opened 2020-03-06
PMK-2673, added a tooltip to cluster info

jruddell:
Interactive tooltip with a link

image


xantorres:
You can have this same functionality in the material-ui tooltip by using the `interactive` prop

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterInfo.tsx
```diff
@@ -93,9 +96,26 @@ const clusterOverviewFields: IClusterDetailFields[] = [
...
+      <ExternalLink className={classes.link} url={applicationLoadBalancer}>
+        here
+      </ExternalLink>
+    </Typography>
+  )
+}
 const bareOsNetworkingFields = [
   { id: 'networkPlugin', title: 'Network Backend', required: true },
-  { id: 'hasLoadBalancer', title: 'Load Balancer', required: false, render: castBoolToStr() },
+  {
```
brennancheung: Weird this one is not a 1-liner like the others.  Might be able to make required default to false to save some space.

jruddell: could also be that this wasn't a component before and just inline, had to make it a component so i could style it.

PR #741, opened 2020-03-07
UX-18, UX-19, UX-16, UX-15 - updates to azure cluster create

jruddell:
Less copy on the platform selection screen

image

___

always show enable workloads on masters

image

___

disable availability zone selection until region is specified

image

___

maximum height on picklists so long lists are usable especially on small screens

image

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/AddAzureClusterPage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/AddAzureClusterPage.js
@@ -234,6 +234,16 @@ const AddAzureClusterPage = () => {
                           info="Set common options from one of the available templates"
                         />
 
+                        {params.template !== 'custom' && (
+                          <CheckboxField
+                            id="allowWorkloadsOnMaster"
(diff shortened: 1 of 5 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/AddAzureClusterPage.js
```diff
@@ -242,11 +252,17 @@ const AddAzureClusterPage = () => {
...
+                                  ? false
+                                  : wizardContext.useAllAvailabilityZones
+                              }
                               info=""
+                              disabled={!params.cloudProviderRegionId}
                             />
 
                             {/* Azure Availability Zone */}
-                            {values.useAllAvailabilityZones || (
+                            {!params.cloudProviderRegionId || values.useAllAvailabilityZones || (
```
brennancheung: Do we need to cast to boolean to avoid it rendering weird things like `'0'` or something like that?

jruddell: we are, im using `!` here to make sure its a boolean

PR #750, opened 2020-03-11
Hide the pf9-system:monitoring tag from the user

kevinylee1991:
Hide the pf9-system:monitoring tag from the user bc the backend does not detect cluster label changes to enable/disable prometheus on clusters

### DIFF
--- a/src/app/core/components/KeyValues.js
+++ b/src/app/core/components/KeyValues.js
@@ -7,7 +7,6 @@ import uuid from 'uuid'
 import { assoc } from 'ramda'
 import { makeStyles } from '@material-ui/styles'
 import { Button } from '@material-ui/core'
-import clsx from 'clsx'
 
 const useKeyValueStyles = makeStyles((theme) => ({
   root: {
@@ -28,9 +27,6 @@ const useKeyValueStyles = makeStyles((theme) => ({
   deleteButton: {
     flexGrow: 0,
     padding: 0,
-  },
-  hidden: {
-    display: 'none',
   },
 }))
 
--- a/src/app/core/components/KeyValues.js
+++ b/src/app/core/components/KeyValues.js
@@ -34,7 +34,7 @@ const useKeyValueStyles = makeStyles((theme) => ({
   },
 }))
 
-const KeyValue = ({ entry = {}, onChange, onDelete, keySuggestions, valueSuggestions, blacklistedTags }) => {
+const KeyValue = ({ entry = {}, onChange, onDelete, keySuggestions, valueSuggestions }) => {
   const classes = useKeyValueStyles()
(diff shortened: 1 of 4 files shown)


src/app/core/components/KeyValues.js
```diff
@@ -28,9 +29,12 @@ const useKeyValueStyles = makeStyles((theme) => ({
     flexGrow: 0,
     padding: 0,
   },
+  hidden: {
```
jruddell: We shouldn't be worrying this component with something specific like `hideMonitoringTags` thats not a good use case for this. Look at how we are doing it on the cluster create forms, we remove the item from the tags array completely.

src/app/core/components/validatedForm/KeyValuesField.js
```diff
@@ -43,6 +45,9 @@ KeyValuesField.propTypes = {
   label: PropTypes.string,
   onChange: PropTypes.func,
   initialValue: PropTypes.arrayOf(EntryShape),
+  // Hide pf9-system:monitoring tag from the display
+  // because that tag should be handled completely by appbert
+  hideMonitoringTags: PropTypes.bool,
```
jruddell: if we were going to do things like this, a much better way would be to pass a blacklist array of tags that we want to avoid, however I think just removing the item from the array before passing the KeyValuesField is acceptable

src/app/core/components/KeyValues.js
```diff
@@ -28,9 +29,12 @@ const useKeyValueStyles = makeStyles((theme) => ({
...
+    display: 'none',
```
jruddell: Its better to not render the element, display none is not needed.

src/app/core/components/KeyValues.js
```diff
@@ -28,9 +29,12 @@ const useKeyValueStyles = makeStyles((theme) => ({
...
+  },
 }))
 
-const KeyValue = ({ entry = {}, onChange, onDelete, keySuggestions, valueSuggestions }) => {
+const KeyValue = ({ entry = {}, onChange, onDelete, keySuggestions, valueSuggestions, blacklistedTags }) => {
```
jruddell: this component doesn't need the blacklist.

src/app/core/components/KeyValues.js
```diff
@@ -120,6 +126,7 @@ const KeyValues = ({ entries: _entries, onChange, keySuggestions, valueSuggestio
           entry={entry}
           onChange={handleChange}
           onDelete={deleteEntry(entry.id)}
+          blacklistedTags={blacklistedTags}
```
jruddell: Filter the entries with the blacklist before mapping / rendering

PR #759, opened 2020-03-20
dont make 2 requests on a delete

jruddell:
we need to tell the event to `preventDefault` so that the form `onSubmit` doesn't fire the same delete call simultaneously


jruddell:
> Just to check, in old code we are waiting to get the response from the server but in new code, we are not waiting to complete the operation if it is intended then code looks ok to me.

we aren't doing anything with the response.

PR #762, opened 2020-03-28
UX-3/Ironic setup wizard

kevinylee1991:
WIP still much fit & finish to do (validation, e2e testing, some css)


src/app/core/components/SimpleLink.tsx
```diff
@@ -60,6 +61,7 @@ const SimpleLink: ComponentType<Props> = forwardRef<HTMLElement, Props>(
         ref={ref}
         href={src || null}
         onClick={handleClick}
+        download={download}
```
jruddell, line 64: minor, but you probably dont need to pull download if its just passed directly though, you can let the `...rest` handle the pass through

src/app/core/components/validatedForm/ValidatedForm.js
```diff
@@ -183,9 +185,10 @@ class ValidatedForm extends PureComponent {
     return !results.includes(false)
   }
 
-  handleSubmit = (event) => {
-    const { clearOnSubmit, onSubmit } = this.props
+  handleSubmit = async (event) => {
+    const { clearOnSubmit, onSubmit, afterSubmit, apiCalls, showToast } = this.props
```
jruddell, line 189: what is `apiCalls` here? whats it supposed to be doing?

src/app/core/components/wizard/Wizard.js
```diff
@@ -5,12 +5,16 @@ import WizardButtons from 'core/components/wizard/WizardButtons'
...
 import { ensureFunction } from 'utils/fp'
-import WizzardStepper from 'core/components/wizard/WizardStepper'
+import WizardStepper from 'core/components/wizard/WizardStepper'
 import CancelButton from 'core/components/buttons/CancelButton'
 
 export const WizardContext = React.createContext({})
 
 class Wizard extends PureComponent {
+  componentDidMount() {
+    this.setState({ activeStep: this.props.startingStep })
```
jruddell, line 15: can probably just do this in the state definition

src/app/core/components/wizard/Wizard.js
```diff
@@ -58,12 +62,14 @@ class Wizard extends PureComponent {
...
-  handleNext = () => {
+  handleNext = async () => {
     const { onComplete } = this.props
     const { steps, activeStep, wizardContext } = this.state
 
-    if (this.nextCb[activeStep] && this.nextCb[activeStep]() === false) {
-      return
+    // This is triggerSubmit in the ValidatedForm
+    if (this.nextCb[activeStep]) {
+      const ok = await this.nextCb[activeStep]()
```
jruddell, line 71: migrating this to an async function may introduce regression on other components using the wizard. Have you tested that these work normally?

src/app/plugins/kubernetes/components/infrastructure/common/actions.js
```diff
@@ -13,12 +13,32 @@ export const combinedHostsCacheKey = 'combinedHosts'
 
 const { resmgr } = ApiClient.getInstance()
 
-const loadResMgrHosts = createContextLoader(
+const getIpPreview = (ips) => (
```
jruddell, line 16: `(ips = [])`

src/app/plugins/openstack/components/hosts/DownloadHostAgentWalkthrough.tsx
```diff
@@ -0,0 +1,165 @@
...
+  )
+}
+
+interface NumberedStepProps {
+  step: number
+  title?: string
+  description: string | JSX.Element
+}
+
+const NumberedSteps: FC<NumberedStepProps> = ({
```
jruddell, line 140: is this whole file copied from the kubernetes side? probably better to move to common and reuse

src/app/plugins/openstack/components/ironicSetup/AuthorizeHostStep.tsx
```diff
@@ -0,0 +1,73 @@
...
+  onNext: any
+  title: string
+}
+
+const AuthorizeHostStep = ({ wizardContext, setWizardContext, onNext, title }: Props) => {
+  const { getParamsUpdater } = useParams(wizardContext)
+
+  const [hosts, hostsLoading, reloadHosts]: IUseDataLoader<ResMgrHost> = useDataLoader(loadResMgrHosts) as any
+
+  const columns = [
```
jruddell, line 37: better to define this outside of the component since it doesn't use anything local to the component. just a constant variable

src/app/plugins/openstack/components/ironicSetup/IronicSetupPage.tsx
```diff
@@ -1,8 +1,225 @@
...
+  const [subnets, subnetsLoading] = useDataLoader(subnetActions.list)
+  const [hosts, hostsLoading] = useDataLoader(loadResMgrHosts)
+  const [startingStep, setStartingStep] = useState(0)
+  const [loading, setLoading] = useState(true)
+  const [submittingStep, setSubmittingStep] = useState(false)
+
+  // Determine current step
+  useEffect(() => {
+    if (!networksLoading && !hostsLoading && !subnetsLoading) {
+      (async () => {
```
jruddell, line 124: Why do you need to do an IIFE here? cant the function for useEffect be async?

src/app/plugins/openstack/components/ironicSetup/IronicSetupPage.tsx
```diff
@@ -1,8 +1,225 @@
...
+          setLoading(false)
+          return
+        }
+
+        setStartingStep(5)
+        setLoading(false)
+        return
+      })()
+    }
+  }, [networks, hosts, subnets])
```
jruddell, line 156: you are using the loading flags in an if expression inside this effect, you need to add them as a dependency here

src/app/plugins/openstack/components/hosts/DownloadHostAgentWalkthrough.tsx
kevinylee1991: Used the other component as a template, but they have fairly different contents

src/app/plugins/openstack/components/ironicSetup/IronicSetupPage.tsx
kevinylee1991: when I try making the useEffect function async I get this error:

TS2345: Argument of type '() => Promise<void>' is not assignable to parameter of type 'EffectCallback'.
  Type 'Promise<void>' is not assignable to type 'void | (() => void)'.
    Type 'Promise<void>' is not assignable to type '() => void'.
      Type 'Promise<void>' provides no match for the signature '(): void'.

Found an article about this on https://medium.com/javascript-in-plain-english/how-to-use-async-function-in-react-hook-useeffect-typescript-js-6204a788a435

src/app/core/components/validatedForm/ValidatedForm.js
kevinylee1991: apiCalls (for lack of a better descriptive word) is a place where they can provide an async function to be called when trying to submit the form. In the case of the ironic wizard, it is required to be able to submit API calls upon pressing next on the multistep wizard and determine whether or not those calls were successful or not before moving onto the next step of the wizard.

src/app/core/components/wizard/Wizard.js
kevinylee1991: Tested out the AWS cluster creation wizard, the steps seem to work correctly from manual testing.

src/app/core/components/validatedForm/ValidatedForm.js
jruddell: I dont think this should be inside the ValidatedForm. The Wizard has an onNext prop that is used with the forms triggerSubmit prop. I think it would be better to let a component implement its specific use case like this. You can let the component use a callback to determine if you should go to the next step in the wizard or not.

src/app/plugins/openstack/components/hosts/DownloadHostAgentWalkthrough.tsx
jruddell: The content isn't what im talking about. You've copied over all of the component part including styling... etc. I think it would make sense to have this be a common component and then both here and in the k8s side you can put in different content with the same components.

src/app/plugins/openstack/components/ironicSetup/IronicSetupPage.tsx
jruddell: Hmm, im not a fan of this leave it for now, i'll take a look when i have time.

src/app/core/components/validatedForm/ValidatedForm.js
```diff
@@ -184,8 +186,9 @@ class ValidatedForm extends PureComponent {
   }
 
   handleSubmit = (event) => {
-    const { clearOnSubmit, onSubmit } = this.props
+    const { clearOnSubmit, onSubmit, afterSubmit } = this.props
```
jruddell, line 189: what is `afterSubmit` used for?

src/app/plugins/openstack/components/ironicSetup/AuthorizeHostStep.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const submitStep = useCallback(async () => {
+    const isValid = validatorRef.current.validate()
+    if (!isValid) {
+      // don't let the user progress to next step
+      return false
+    }
+
+    try {
+      setSubmitting(true)
+      await addRole(wizardContext.selectedHost[0].id, 'pf9-onboarding', {})
```
jruddell, line 47: do we need to look at the response from adding a role here or just that the request doesn't blow up means its successful? Is there something a little more explicit we can use to know for sure that it worked?

src/app/plugins/openstack/components/ironicSetup/IronicSetupPage.tsx
jruddell: I think a better way is to just define the async function inside your effect and then call it.

```
useEffect(() => {
  async function setup() => {
    // code
  }
  setup()
}, [/* deps */ ])
```

src/app/plugins/openstack/components/ironicSetup/IronicSetupPage.tsx
```diff
@@ -1,8 +1,296 @@
...
+// Check for presence of provisioning network as well as neutron-server configured
+const networkConfigured = async (networks) => {
+  const neutronServerResponse = await getService('neutron-server')
+  const dnsDomain = path(['neutron', 'DEFAULT', 'dns_domain'], neutronServerResponse)
+  const dnsForwardingAddresses = path(['extra', 'dnsmasq_dns_servers'], neutronServerResponse)
+  const alreadyConfigured = path(['extra', 'configured'], neutronServerResponse)
+  const provisioningNetwork = getProvisioningNetwork(networks)
+
+  if (alreadyConfigured && provisioningNetwork) {
+    initialContext.provisioningNetwork = provisioningNetwork.id
```
jruddell, line 58: not a fan of assigning a value directly into the initialContext like this. The wizard wouldn't reflect the initial context changes inside its context. Can we keep the updates to the wizards context via the `setWizardContext` fn it exposes?

src/app/core/components/SimpleLink.tsx
```diff
@@ -15,6 +15,7 @@ interface Props {
   icon?: string
   variant?: ISimpleLinkVariant
   onClick?: (event: React.MouseEvent<HTMLAnchorElement, MouseEvent>) => void
+  download?: boolean
```
xantorres, line 18: Is this doing anything?

src/app/core/components/validatedForm/ListTableField.tsx
```diff
@@ -0,0 +1,60 @@
...
+  columns?: any
+  data: any
+  title?: string
+  loading: boolean
+}
+
+// TODO: all the ValidatedForm stuff is in JS and we need the props to be merged
+// into this component.  Refactor this later on when we can convert
+// ValidatedForm.js to TypeScript.
+export interface IValidatedForm {
```
xantorres, line 20: This doesn't belong here, either move it to ValidatedForm (after refactoring it to tsx) or create a new file

src/app/core/components/validatedForm/ListTableField.tsx
```diff
@@ -0,0 +1,60 @@
...
+  id: string
+  validations?: any[] // What's the best way to type this?
+  required?: boolean
+  initialValues?: any
+}
+
+// TODO: is forwardRef actually needed here?
+const ListTableField: React.ComponentType<Props> = forwardRef<HTMLElement, Props>(
```
xantorres, line 28: `forwardRef` is only needed if we are going to use this component wrapped in a `Tooltip` or when using the `withInfoTooltip` HoC

src/app/core/components/validatedForm/ValidatedForm.js
```diff
@@ -39,6 +40,7 @@ const styles = (theme) => ({
  * ValidatedForm is a HOC wrapper for forms.  The child components define the
  * data value schema and the validations.
  */
+@withToast
```
xantorres, line 43: Do we need this? I don't see the Toast being called anywhere in the component

src/app/utils/misc.js
```diff
@@ -203,3 +203,7 @@ export const normalizeUsername = (name = '') => {
 export const indefiniteArticle = (word = '') => (
   /^([aeiou])/i.test(word) ? 'an' : 'a'
 )
+
+export const sleep = async (ms) => (
```
xantorres, line 207: Please move this to `utils/async.js`

src/app/core/components/validatedForm/ValidatedForm.js
kevinylee1991: Oh i think it was a remnant of when I was trying to find a solution, I believe it can be removed

src/app/plugins/openstack/components/ironicSetup/AuthorizeHostStep.tsx
kevinylee1991: If it doesn't blow up then the UI did it's job initiating the backend to add the role, unfortunately there is no way to tell if the role adding was successful from the API response to the request that initiates the role.

src/app/plugins/openstack/components/ironicSetup/IronicSetupPage.tsx
kevinylee1991: I'm not sure how to access setWizardContext from outside of the Wizard component -- could you help point out how I could do that?

src/app/core/components/SimpleLink.tsx
kevinylee1991: probably not

src/app/core/components/validatedForm/ValidatedForm.js
kevinylee1991: remnant of earlier solution, good catch

src/app/core/components/validatedForm/ListTableField.tsx
kevinylee1991: I see, I could see the potential need for a tooltip in the future so I'll leave it in

PR #775, opened 2020-04-25
UX-196/update k8s dashboard url

jruddell:
update k8s dashboard url based on its version, fix an issue with storage classes throwing an exception


src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
```diff
@@ -61,7 +61,11 @@ const BasicStep = ({ onSubmit, triggerSubmit, wizardContext }) => {
   const { params, getParamsUpdater } = useParams()
   const defaultStorageClassForCurrentCluster = (storageClass) =>
     storageClass.clusterId === params.clusterId &&
-    storageClass.metadata.annotations['storageclass.kubernetes.io/is-default-class'] === 'true'
+    pathOr(
+      false,
```
kevinylee1991, line 65: should this be string 'false' to match how the affirmative value uses string 'true'?

jruddell: It doesn't make a difference since this `pathOr` is triple equal compared to `'true'`

PR #792, opened 2020-06-15
UX-264/Make wording on Sandbox less confusing

brennancheung:
The below button is confusing because we already have an `Add Cluster` button on the same page right below.  The button really asks them to sign up for Free Tier so we should word it accordingly.

Spoke with Madhura about the change already.

deploy-now

### DIFF
--- a/src/app/core/containers/AuthenticatedContainer.js
+++ b/src/app/core/containers/AuthenticatedContainer.js
@@ -163,8 +163,6 @@ const renderRawComponents = moize((plugins) =>
     .flat(),
 )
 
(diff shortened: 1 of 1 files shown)


jruddell:
@smadhura @brennancheung be careful about duplicating work. I think kevin is already doing/has completed the segment stuff.

@brennancheung do you have a JIRA for this? can you include it in the title of the PR under the format `UX-XXXX/Make wording on Sandbox less confusing`?

brennancheung:
Originally no JIRA, but just created one now after the fact for tracking.  This change originated initially from just a Slack conversation.

PR #797, opened 2020-06-25
UX-268, UX-257, UX-195, UX-157, UX-147/Bug fixes

jruddell:
image (15)

image (16)

image (17)

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
@@ -99,15 +99,14 @@ const IPValidator = customValidator((value, formValues) => {
 const IPValidatorRange = customValidator((value, formValues) => {
   const pairs = value.split(',')
   if (!pairs.length) return false
-  let isValid = true
 
-  pairs.forEach((pair = '') => {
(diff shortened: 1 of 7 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/AddBareOsClusterPage.js
```diff
@@ -81,17 +81,35 @@ const cidrBlockSizeValidator = customValidator((value) => {
...
 
+const IPValidatorRange = customValidator((value, formValues) => {
+  const pairs = value.split(',')
+  if (!pairs.length) return false
+  let isValid = true
+
+  pairs.forEach((pair = '') => {
+    const [startIp, endIp] = pair.split('-')
+    if (!isValidIpAddress(startIp) || !isValidIpAddress(endIp)) {
+      isValid = false
```
brennancheung, line 107: Can you just `return false` immediately and just skip the `isValid` boolean flag?

jruddell: `return false` will stop the `forEach` from running. A good optimization to add but that wont return false for the validator. I could switch the iterator to be a for loop so we can do what you're asking... But I find that to be less readable.

PR #801, opened 2020-07-01
UX-251/Validate azure ssh key

jruddell:
## With error
image (18)

## Without error
image (19)


jruddell:
> is the yarn.lock supposed to be updated?

yes, we should push the lock file

PR #802, opened 2020-07-03
UX-279/Added drift tracking code for pmk live demo

kevinylee1991:


src/app/utils/tracking.js
```diff
@@ -14,3 +14,37 @@ export const trackEvent = (name, body = {}) => {
   }
   window.analytics.track(name, body)
 }
+
+export const injectDrift = () => {
+  const existingScript = document.getElementById('driftCode')
+  if (!existingScript) {
```
brennancheung, line 20: I would just `if (existingScript) return` so you don't have to indent all the rest of the code.

PR #806, opened 2020-07-10
UX-99/Add banner for upgradable clusters and added to cluster health …

kevinylee1991:
…status

Link to JIRA for the wireframes (photos): https://platform9.atlassian.net/browse/UX-99

Looking for some feedback/thoughts on below bullets and screenshots:
- Not sure if the banner message makes the most sense with the wording.
- Held off on implementing the special cluster details banner, not sure if there is a very clean way to do so, and I don't see a huge benefit.

Screen Shot 2020-07-10 at 2 08 25 PM

Screen Shot 2020-07-10 at 1 57 56 PM

### DIFF
--- a/src/app/core/banners/ClusterUpgradeBanner.tsx
+++ b/src/app/core/banners/ClusterUpgradeBanner.tsx
@@ -1,0 +1,44 @@
+import React from 'react'
+import useReactRouter from 'use-react-router'
+import BannerButton from 'core/components/buttons/BannerButton'
+import { makeStyles, Theme } from '@material-ui/core'
+import BannerContainer from 'core/components/notifications/BannerContainer'
+import BannerContent from 'core/components/notifications/BannerContent'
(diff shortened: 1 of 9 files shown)


src/app/core/banners/ClusterUpgradeBanner.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+const ClusterUpgradeBanner = () => {
+  const classes = useStyles({})
+  const { history } = useReactRouter()
+
+  const [clusters, clustersLoading] = useDataLoader(clusterActions.list)
+  const upgradableClusters = clusters.filter(cluster => cluster.canUpgrade)
+
+  const redirectToClusters = () => {
+    history.push('/ui/kubernetes/infrastructure')
```
jruddell, line 28: Use the Route class that we have setup for getting a url. I believe the file is called `route.ts`

PR #807, opened 2020-07-13
Pendo POC

brennancheung:
We are performing a POC for https://www.pendo.io/.

The plan is to deploy this branch to `ui-dev` during the POC trial so the rest of the team can experiment with it.

The plan will be to evaluate if Pendo.io can be used to provide the guided tour to help onboard users and also to collect customer feedback for product improvement.


brennancheung:
Done with POC.  Closing out this branch.

PR #809, opened 2020-07-21
Appcues POC

brennancheung:
This branch is *NOT* meant to be merged.

We are currently doing a POC with Appcues to determine how good of a fit it is for collecting feedback from customers like NPS, and providing onboarding flows.


brennancheung:
We have decided to move forward with Appcues.  Leaving this branch open for now so we can use it later once we have more detailed specs.

PR #810, opened 2020-07-28
UX-305 Segment delete events

brennancheung:


src/app/plugins/kubernetes/components/infrastructure/cloudProviders/actions.js
```diff
@@ -32,7 +33,13 @@ export const cloudProviderActions = createCRUDActions(cloudProvidersCacheKey, {
     return { ...cloudProvider }
   },
   updateFn: ({ uuid, ...data }) => qbert.updateCloudProvider(uuid, data),
-  deleteFn: ({ uuid }) => qbert.deleteCloudProvider(uuid),
+  deleteFn: ({ uuid, name, type }) => {
+    trackEvent('Delete Cloud Provider', {
+      cloud_provider_name: name,
+      cloud_provider_type: type,
+    })
+    return qbert.deleteCloudProvider(uuid)
```
jruddell, line 41: in case this call fails, you should put the track event after the delete call

PR #819, opened 2020-08-12
UX-439/Add ability to switch between ironic and kubernetes stacks

kevinylee1991:
Piyush's DU has an ironic / kubernetes region and apparently mavenir may have something similar, added the ability to switch between the two stacks.

### DIFF
--- a/src/app/core/components/Navbar.js
+++ b/src/app/core/components/Navbar.js
@@ -22,7 +22,7 @@ import { except, pathStrOr } from 'app/utils/fp'
 import clsx from 'clsx'
 import { withHotKeys } from 'core/providers/HotKeysProvider'
 import moize from 'moize'
-import { assoc, flatten, pluck, prop, propEq, propOr, where, equals } from 'ramda'
+import { assoc, flatten, pluck, prop, propEq, propOr, where, equals, indexOf } from 'ramda'
 import { matchPath, withRouter } from 'react-router'
 import FontAwesomeIcon from 'core/components/FontAwesomeIcon'
 import {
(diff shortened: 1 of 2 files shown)


src/app/core/components/Navbar.js
```diff
@@ -272,20 +278,20 @@ const styles = (theme) => ({
     boxSizing: 'border-box',
     backgroundRepeat: 'no-repeat',
     backgroundImage: ({ stack }) =>
-      stack === 'openstack'
-        ? 'url(/ui/images/logo-sidenav-os.svg)'
-        : 'url(/ui/images/logo-kubernetes-h.png)',
+      stack === 'kubernetes'
```
jruddell, line 281: Theres an implicit default here that metalstack uses the same url as openstack. if you want to make this a function so that it can be more explicit that would be a good improvement

src/app/core/components/Navbar.js
```diff
@@ -566,19 +573,40 @@ class Navbar extends PureComponent {
     )
   }
 
+  switchStacks = () => {
+    const { stack, setStack, getContext, history } = this.props
+    const { features: { experimental: { ironic } } } = getContext()
+
+    // Until openstack is served in the new UI stack will only
```
jruddell, line 580: There is a lot of implicit behavior here. Can we be more explicit about how we change stacks? We know what the current stack is, we should know what the next stack is going to be based on the direction of the slider you click on. Can the click handler for the slider pass the next stack to switch to?

src/app/core/components/Navbar.js
kevinylee1991: I'm not sure if I did it the way you meant but I'll have you take a look when I post the update

PR #822, opened 2020-08-14
UX-40 UX-309 Segment misc events

brennancheung:


roopakparikh:
Brennan, why is the 'tracking' part of the places where we are referencing and NOT part of the qbert, keystone objects ?

jruddell:
> Brennan, why is the 'tracking' part of the places where we are referencing and NOT part of the qbert, keystone objects ?

I had thought about this on the first of these tracking pull requests. Technically its more error proof to put the tracking directly with where you make API calls. However, the UI has some "tech debt" that requires you to use these actions to make requests. So the only way the API methods are called are via these actions. Meaning, we dont interface with the api client classes except in these actions

PR #826, opened 2020-08-20
UX-96/Added CSI driver details to cluster overview

knnguy:
Screen Shot 2020-08-20 at 1 03 19 PM


src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterInfo.tsx
```diff
@@ -165,10 +165,29 @@ const azureCloudFields = [
...
 
+const csiDriverFields = [
+  { id: 'metadata.name', title: 'Driver Name', required: true },
+  {
+    id: 'spec.volumeLifecycleModes',
+    title: 'Capabilities',
+    required: true,
+    render: (modes) => {
+      return (
+        <ul style={{ listStyleType: 'none', paddingInlineStart: '0px', margin: '0px' }}>
```
jruddell, line 176: try to avoid inline styling. You can use a hook `useStyles` to define how this should look

PR #829, opened 2020-08-24
UX-136/Added link to kubectl docs

knnguy:
Screen Shot 2020-08-24 at 3 01 02 PM


src/app/plugins/kubernetes/links.ts
```diff
@@ -43,6 +43,7 @@ export const bareOSSetupDocumentationLink = `${k8sBaseUrl}/create-multimaster-ba
 
 // Clusters
 export const remoteMonitoringDocsLink = `${k8sBaseUrl}/enable-remote-monitoring/`
+export const kubectlInstallationDocumentationLink = 'https://docs.platform9.com/kubernetes/kubectl/'
```
jruddell, line 46: I believe we have a docs url base already. If you can follow the same pattern as the other urls here that would be good incase we ever change our docs base url :)

PR #830, opened 2020-08-25
UX-416/Modified the conditions that allow the 'Finish and Review' button to be shown

knnguy:
Screen Shot 2020-08-24 at 5 32 00 PM


knnguy:
Changes made:

- If multiple master nodes are selected, the 'Finish and Review' button only shows up on the Network step, after the VIP and VIP interface fields have been filled

Screen Recording Video Demonstration shows:

1. What happens if you select multiple master nodes
2. What happens if you select only one master node and check the 'Enable workloads on all master nodes' box


[Screen Recording 2020-08-26 at 11.20.12 AM.zip](https://github.com/platform9/pf9-ui-plugin/files/5131691/Screen.Recording.2020-08-26.at.11.20.12.AM.zip)

PR #831, opened 2020-08-25
UX-436/Fixed 'Create a sample pod' task completion problem

knnguy:
[Screen Recording 2020-08-24 at 11.00.33 PM.zip](https://github.com/platform9/pf9-ui-plugin/files/5125269/Screen.Recording.2020-08-24.at.11.00.33.PM.zip)


src/app/plugins/kubernetes/components/onboarding/PodSetup.tsx
```diff
@@ -35,6 +35,7 @@ const PodSetup = ({ onComplete, initialPanel }: Props) => {
 
   const handleCreatePod = useCallback(() => {
     history.push(routes.pods.add.path())
+    localStorage.setItem(onboardingPodSetup, 'true')
```
jruddell, line 38: I'd look closer as to why the pods list is empty after creating a pod. This history push doesn't guarantee that you have created a pod so we don't want to set a flag to mark it as complete here.

PR #834, opened 2020-08-27
UX-241/Modified Welcome header message on dashboard

knnguy:
Changes made:

- If a user has a displayname, the displayname will be displayed in the welcome header 

- If a user does not have a displayname, only 'Welcome!' will be displayed

Screen Shot 2020-08-27 at 11 56 15 AM

### DIFF
--- a/src/app/plugins/kubernetes/components/dashboard/DashboardPage.tsx
+++ b/src/app/plugins/kubernetes/components/dashboard/DashboardPage.tsx
@@ -14,7 +14,6 @@ import { cloudProviderActions } from '../infrastructure/cloudProviders/actions'
 // Components
 import StatusCard, { StatusCardProps } from './StatusCard'
 import { Typography } from '@material-ui/core'
-import { capitalizeString, normalizeUsername } from 'utils/misc'
 
 import ClusterSetup, {
   clustersHaveMonitoring,
@@ -313,7 +312,6 @@ const DashboardPage = () => {
   const [users, loadingUsers] = useDataLoader(mngmUserActions.list)
   const user = users.find((x) => x.username === session.username)
   const displayname = user?.displayname
-  const username = capitalizeString(normalizeUsername(session.username))
 
   if (loadingUsers) {
     return null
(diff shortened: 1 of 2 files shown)


src/app/plugins/kubernetes/components/dashboard/DashboardPage.tsx
```diff
@@ -310,9 +310,18 @@ const DashboardPage = () => {
     return [hasClusters, hasAccess, hasMonitoring].findIndex((item) => !item)
   }, [hasClusters, hasMonitoring, hasAccess])
 
+  const [users, loadingUsers] = useDataLoader(mngmUserActions.list)
```
jruddell, line 313: When we load the session we are storing user details, does the `displayName` not get stored in the session?

knnguy: No, it's not stored in session. That's why I had to load all the users and get it from there

src/app/utils/misc.js
```diff
@@ -190,16 +190,18 @@ export const getCookieValue = (name) => {
   return val ? val.pop() : ''
 }
 
-export const normalizeUsername = (name = '') => {
-  if (!name) return name
+export const normalizeUsername = (emailAddress = '') => {
```
jruddell, line 193: just call this `email` instead of `emailAddress`

src/app/utils/misc.js
```diff
@@ -190,16 +190,18 @@ export const getCookieValue = (name) => {
...
+  if (!emailAddress) return emailAddress
 
   // split emails from @ sign and take the left hand side
-  const [username] = name.split('@')
-  return username
+  const [username] = emailAddress.split('@')
+
+  // split username from  '.' for emails in this format: first.last@domain.tld => [first, last]
+  const name = username.split('.')
```
jruddell, line 200: I'd be careful here as emails arent always reflective of a users first and last name. If there is no `displayName` the ticket calls out just rendering `Welcome` lets just do that. No need to use the `session.username` anymore.

PR #835, opened 2020-08-28
UX-519/Added delete option to pods list

knnguy:
Screen Shot 2020-08-28 at 10 48 01 AM


src/app/plugins/kubernetes/components/pods/PodsListPage.js
```diff
@@ -122,6 +122,7 @@ const renderStatus = (phase) => {
 
 export const options = {
   showCheckboxes: false, // until we have actions to perform, disable this
```
jruddell, line 124: delete will not work if the user is unable to select the row. Go ahead and remove this flag

PR #841, opened 2020-09-03
Fix deployments list page

xantorres:


jruddell (approved):

src/app/plugins/kubernetes/components/pods/model.ts
```diff
@@ -17,6 +19,19 @@ export interface IPodSelector extends IGenericResource<GetClusterPodsItem> {
   clusterName: string
 }
 
+export interface IDeploymentSelector extends IGenericResource<GetClusterDeploymentsItem> {
+  dashboardUrl: string
+  id: string
+  name: string
+  namespace: string
+  // labels: string
```
jruddell, line 27: is there a conflict with labels? you can use `Omit<>` to remove a specific key from a type if you need to specify `string` as the labels type

PR #845, opened 2020-09-09
UX-278/Fixed cluster etcd backup problem

knnguy:
Changes to a cluster's etcd backup properties are now reflected in the UI

[Edit Etcd Backup Interval.mov.zip](https://github.com/platform9/pf9-ui-plugin/files/5191881/Edit.Etcd.Backup.Interval.mov.zip)

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
@@ -7,6 +7,7 @@ import {
   createBareOSCluster,
   getKubernetesVersion,
   getProgressPercent,
+  getEtcdBackupPayload,
 } from 'k8s/components/infrastructure/clusters/helpers'
 import {
   clustersSelector,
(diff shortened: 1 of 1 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -76,10 +76,17 @@ export const clusterActions = createCRUDActions(ActionDataKeys.Clusters, {
     }
   },
   updateFn: async ({ uuid, ...params }) => {
-    const updateableParams = 'name tags numWorkers numMinWorkers numMaxWorkers etcdBackup'.split(
-      ' ',
-    )
+    const updateableParams = 'name tags numWorkers numMinWorkers numMaxWorkers'.split(' ')
+
     const body = pick(updateableParams, params)
+    body.etcdBackup = {
```
jruddell, line 82: I believe there is a helper to generate this payload. Also... we should only be passing etcdBackup fields if the cluster has etcd enabled. If we pass this through on a cluster without it enabled I believe the update call would fail

jruddell (approved):

PR #846, opened 2020-09-09
Fix a bug with ListTable rows getting unselected when updating data

xantorres:


src/app/core/components/listTable/ListTable.js
```diff
@@ -100,14 +100,21 @@ class ListTable extends PureComponent {
...
-    ) { return }
+    const { uniqueIdentifier, data } = prevProps
+    const { selected } = this.state
+    const ids = this.pluckDataIds(data)
 
-    this.setState({
-      selected: [],
+    const existingSelectedRows = selected.filter((row) => {
+      const id =
+        uniqueIdentifier instanceof Function ? uniqueIdentifier(row) : row[uniqueIdentifier]
```
jruddell, line 109: maybe we should make this a helper function so we don't have to do this tedious check in many places.

jruddell (approved):

PR #848, opened 2020-09-09
UX-437/Modified Alert retrieval from API

knnguy:
Also added an 'Exported Namespace' column to the alerts table

Screen Shot 2020-09-09 at 2 33 09 PM


src/api-client/Qbert.ts
```diff
@@ -1143,24 +1144,46 @@ class Qbert extends ApiService {
...
+      const rule = alertsFromRules.find((ruleAlert) => {
+        return (
+          equals(ruleAlert.name, alert.labels.alertname) &&
+          equals(ruleAlert.labels.severity, alert.labels.severity)
+        )
+      })
+      return {
+        ...alert,
+        clusterId: clusterUuid,
+        id: `${alert.labels.alertname}${clusterUuid}${alert.labels.exported_namespace}${alert.labels.severity}`,
```
jruddell, line 1166: I think we should think about how we're defining the ID. Wouldn't the alert type, clusterUuid and timestamp be enough to have a unique ID for each alert?

src/api-client/Qbert.ts
```diff
@@ -1143,24 +1144,46 @@ class Qbert extends ApiService {
...
+    const url = `/clusters/${clusterUuid}/k8sapi/api/v1/namespaces/pf9-monitoring/services/http:sys-prometheus:9090/proxy/api/v1/alerts`
     const response = await this.client.basicGet<GetPrometheusAlerts>({
       url,
       options: {
         clsName: this.getClassName(),
         mthdName: 'getPrometheusAlerts',
       },
     })
+    const alertsFromRules = await this.getPrometheusAlertsFromRules(clusterUuid)
+    return response.alerts.map((alert) => {
```
jruddell, line 1156: We should probably do this data mapping inside a selector

src/api-client/Qbert.ts
```diff
@@ -1143,24 +1144,46 @@ class Qbert extends ApiService {
...
-    const url = `/clusters/${clusterUuid}/k8sapi/api/v1/namespaces/pf9-monitoring/services/http:sys-prometheus:9090/proxy/api/v1/rules`
+    const url = `/clusters/${clusterUuid}/k8sapi/api/v1/namespaces/pf9-monitoring/services/http:sys-prometheus:9090/proxy/api/v1/alerts`
     const response = await this.client.basicGet<GetPrometheusAlerts>({
       url,
       options: {
         clsName: this.getClassName(),
         mthdName: 'getPrometheusAlerts',
       },
     })
+    const alertsFromRules = await this.getPrometheusAlertsFromRules(clusterUuid)
```
xantorres, line 1155: You can create a DataLoader from this and call it here, so that then you will be able to access it from the selectors.

xantorres (changes requested):

src/api-client/Qbert.ts
xantorres: We don't need this ID here, if you really want to have a `uniqueIdentifier`, you can define a composite ID when creating the action (using createCRUDActions) by passing an array of identifiers such as `uniqueIdentifier: ['alert.labels.alertname', 'clusterId', 'alert.labels.exported_namespace', 'alert.labels.severity']`.
Either way, I don't even think we should be caching these alerts, so it might be better to just add a `cache: false` option so that we don't even need to use a unique identifier.

xantorres (changes requested):

src/api-client/Qbert.ts
xantorres: Backend should always provide us with unique identifiers for the data, it makes no sense that we have to deal with this kind of stuff.
Either way, in this specific case we probably don't want to cache these items as they are just used in this specific instance and they are being refreshed constantly so caching doesn't make a lot of sense. So I would just get rid of the cache overall by using the new option `cache: false` when creating the action, and maybe generate a unique ID here by using `uuid.v4()` or whatever, as a workaround so that react can track the items when rendering them.

xantorres (changes requested):

src/api-client/Qbert.ts
jruddell: We will have future use cases like dom manipulation based on an elements ID. I'd like these ID's to be derived from the data so we can recreate these id's if needed, lets use `<alert type>-<clusterUuid>-<alert timestamp>` here. Those fields should never be empty

knnguy:
Screen Shot 2020-09-11 at 2 56 12 PM

Screen Shot 2020-09-11 at 2 56 36 PM

src/app/plugins/kubernetes/components/monitoring/actions.js
```diff
@@ -0,0 +1,39 @@
+import ApiClient from 'api-client/ApiClient'
```
jruddell, line 1: we have an actions file for alerts already. I believe its `/components/alarms/actions.js`. Probably better to put this action there

knnguy: This was part of Kevin's change. I originally had the action in the alarm actions file, but Kevin added the same action in the monitoring alarms file. That's why we decided to just use the one he had. Should I copy and paste the action back into the alarm actions file as well?

jruddell (approved):

PR #849, opened 2020-09-09
Add cache option and fix alerts page

xantorres:
Added a `cache` option that is `true` by default but can be set to `false` to allow disabling the caching of the items so that they will always be fetched and replaced entirely in the redux store when using the `useDataLoader` hook, instead of "upserted".

Note that with `cache: false` option, data will be fetched and the store cached data replaced every time any of the `params` passed to the `useDataLoader` hook change, so beware of just updating those params when the data actually needs to be retrieved.

Whereas when using the `cache: true` option (the default) data will only be re-fetched and the store updated when the params defined by the `indexBy` option change.

### DIFF
--- a/src/app/core/components/progress/Progress.js
+++ b/src/app/core/components/progress/Progress.js
@@ -137,7 +137,8 @@ class Progress extends PureComponent {
     }
     return (
       <div
-        className={clsx(classes.content, {
+        className={clsx(classes.content, 'progressContent', {
+          loading,
           [classes.hiddenContent]: loading && !overlay,
           [classes.contentLoading]: loading,
         })}
--- a/src/app/plugins/kubernetes/components/alarms/actions.js
+++ b/src/app/plugins/kubernetes/components/alarms/actions.js
@@ -50,15 +50,11 @@ const timestampSteps = {
   '3.h': [30, 'm'],
   '1.h': [10, 'm'],
 }
+const selector = makeParamsClustersSelector()
 
 export const loadTimeSeriesAlerts = createContextLoader(
   ActionDataKeys.AlertsTimeSeries,
   async ({ chartTime, ...params }) => {
-    // Invalidate cache -- can't get cache working properly for this
-    // collection. Collection is somewhat different from all other
-    // types of collections.
-    loadTimeSeriesAlerts.invalidateCache()
-
     const timeNow = moment().unix()
     const [number, period] = chartTime.split('.')
     const timePast = moment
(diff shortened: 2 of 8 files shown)


src/app/core/components/progress/Progress.js
```diff
@@ -138,6 +138,7 @@ class Progress extends PureComponent {
     return (
       <div
         className={clsx(classes.content, {
+          hiddenContent: loading,
```
jruddell, line 141: what is the difference between this and the next line? we reference hiddenContent twice, once for just loading and one for loading without overlay? does this have any negative effects on other components that use progress?

src/app/plugins/kubernetes/components/alarms/actions.js
```diff
@@ -86,11 +81,8 @@ export const loadTimeSeriesAlerts = createContextLoader(
...
   {
     entityName: 'AlertTimeSeries',
-    // make uniqueIdentifier timestamp bc of autosort
-    // not sure if this has to do with the reason why the
-    // cache does not work properly
-    uniqueIdentifier: 'id',
     selectorCreator: makeTimeSeriesSelector,
-    indexBy: ['clusterId', 'chartTime'],
+    // indexBy: ['clusterId', 'chartTime'],
+    noCache: true,
```
jruddell, line 86: personally I would prefer this be called `cache` and the default is true. but you can explicitly set `cache: false` when needed

src/app/core/components/progress/Progress.js
kevinylee1991: This is a bit of a hack to ensure that the other components don't get affected and only applies to this one. There was an issue where when switching the chart time toggle, the chart showed a wavy chart with extra data points during the load (the issue we were fixing), but it was a big headache to debug and opted for a workaround. Fixing the issue with extra data points would be ideal for sure

src/app/core/components/progress/Progress.js
```diff
@@ -137,7 +137,8 @@ class Progress extends PureComponent {
     }
     return (
       <div
-        className={clsx(classes.content, {
+        className={clsx(classes.content, 'progressContent', {
+          hidden: loading,
```
jruddell, line 141: Lets call this loading instead of hidden. let the view decide if it should be hidden or not. Also `hidden` is a common class that most frameworks expose as a class, it could potentially be a problem

xantorres: We don't have any library targetting global classnames (as far as I'm aware of) but I changed it anyway

PR #853, opened 2020-09-11
UX-521/Added cluster update time to cluster table and cluster details

knnguy:
Screen Shot 2020-09-11 at 11 40 31 AM

Screen Shot 2020-09-11 at 11 41 56 AM


knnguy:
The cluster creation time was already displayed. As for the cluster update time, I'm assuming it's the `lastOP `value. According to the API reference doc, `lastOP` is the last operation such as node attach or node detach that was performed on the cluster.

jruddell (approved):

PR #854, opened 2020-09-12
UX-580/Using AlertManager API for alerts

jruddell:
Minor styling update, exposing more relevant data.
image (25)
image (26)
image (27)

### DIFF
--- a/src/api-client/Qbert.ts
+++ b/src/api-client/Qbert.ts
@@ -338,7 +338,7 @@ class Qbert extends ApiService {
   // @param clusterId = cluster.uuid
   // @param nodes = [node1Uuid, node2Uuid, ...]
   detach = async (clusterId, nodeUuids) => {
-    const body = nodeUuids.map((uuid) => ({ uuid }))
+    const body = nodeUuids.map((nodeUuid) => ({ uuid: nodeUuid }))
     const url = `/clusters/${clusterId}/detach`
     return this.client.basicPost({
       url,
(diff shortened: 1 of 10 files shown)


knnguy (approved):

src/api-client/qbert.model.ts
```diff
@@ -1104,3 +1104,32 @@ export interface IGetPrometheusAlertsOverTime {
...
+  fingerprint: string
+  receivers: Receiver[]
+  startsAt: string
+  status: Status
+  updatedAt: string
+  generatorURL: string
+  labels: { [key: string]: string }
+}
+
+export interface Annotations {
```
kevinylee1991, line 1122: If the annotations on alarms are similar to rules, I believe there can be some other potential values too, like summary or rules for example. Can check if the annotations in the existing alarms are the same as in rules and could possibly apply that to alarms that have not been active yet

src/api-client/Qbert.ts
```diff
@@ -1144,6 +1147,22 @@ class Qbert extends ApiService {
     return response
   }
 
+  getAlertManagerAlerts = async (uuid): Promise<AlertManagerAlert[]> => {
```
kevinylee1991, line 1150: change uuid to something else like clusterId or clusterUuid to avoid overlap with uuid import

src/api-client/Qbert.ts
```diff
@@ -1144,6 +1147,22 @@ class Qbert extends ApiService {
...
+    const url = `/clusters/${uuid}/k8sapi/api/v1/namespaces/pf9-monitoring/services/http:sys-alertmanager:9093/proxy/api/v2/alerts`
```
kevinylee1991, line 1151: uuid

src/api-client/Qbert.ts
```diff
@@ -1144,6 +1147,22 @@ class Qbert extends ApiService {
...
+    const alerts = await this.client.basicGet<AlertManagerRaw[]>({
+      url,
+      options: {
+        clsName: this.getClassName(),
+        mthdName: 'getPrometheusAlerts',
+      },
+    })
+    return alerts?.map((alert) => ({
+      ...alert,
+      clusterId: uuid,
```
kevinylee1991, line 1161: uuid

PR #855, opened 2020-09-14
UX-577/Nodes & clusters show resource usage even when offline

xantorres:
image


src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
```diff
@@ -80,9 +80,10 @@ const renderNodeLink = (_, { uuid, nodes }) => {
 
 const toMHz = (value) => value * 1024
 
-const renderStats = (_, { usage }) => {
+const renderStats = (_, { usage, connectionStatus }) => {
   const hasValidStats = !!path(['compute', 'current'], usage)
-  if (!hasValidStats) {
+  const hasValidConnectionStatus = ['connected', 'partially_connected'].includes(connectionStatus)
+  if (!hasValidStats || !hasValidConnectionStatus) {
```
jruddell, line 86: I think it would be better for us to not render null but uplift the null / empty state for resource usage. this is a more consistent UX and better for the user. Thoughts?

jruddell (approved):
The converging status is incorrect. I'll make a ticket for that :)

PR #861, opened 2020-09-17
UX-586/Added API FQDN field to cluster creation review tables and cluster de…

knnguy:
…tails

- BareOs, AWS, and Azure cluster creation review tables now all show the API FQDN field (BareOs already showed the field so I only modified AWS and Azure)
- API FDQN is also now shown in cluster details if the cluster has one

Screen Shot 2020-09-17 at 1 46 14 PM

Screen Shot 2020-09-17 at 1 47 27 PM

Screen Shot 2020-09-17 at 1 53 07 PM

Screen Shot 2020-09-17 at 2 03 12 PM


jruddell (approved):
👍

PR #862, opened 2020-09-17
UX-275/Modified payload for scaling master and worker nodes

knnguy:
- Payload for scaling worker nodes has been modified

Screen Shot 2020-09-17 at 3 29 31 PM

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
@@ -80,7 +80,6 @@ export const clusterActions = createCRUDActions(ActionDataKeys.Clusters, {
     const updateableParams = 'name tags numWorkers numMinWorkers numMaxWorkers'.split(' ')
 
     const body = pick(updateableParams, params)
-    body.etcdBackup = getEtcdBackupPayload('etcdBackup', params)
 
     await qbert.updateCluster(uuid, body)
     trackEvent('Update Cluster', { uuid })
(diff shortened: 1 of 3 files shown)


knnguy:
@jruddell I noticed that the `handleSubmit` function is passed to the `ScaleMasters` component in `ScaleMastersPage.tsx` , but it is never used. I still modified the `handleSubmit` function anyways, but not sure if that should be taken out.

Screen Shot 2020-09-17 at 4 15 12 PM

jruddell:
> @jruddell I noticed that the `handleSubmit` function is passed to the `ScaleMasters` component in `ScaleMastersPage.tsx` , but it is never used. I still modified the `handleSubmit` function anyways, but not sure if that should be taken out.
> 
> Screen Shot 2020-09-17 at 4 15 12 PM

I believe when Brennan wrote this the intention was to have the ability to pass an onSubmit prop to this component to override how we normally submit... Should the view need to override it. But if were not using it then we can omit for now.

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -80,7 +80,6 @@ export const clusterActions = createCRUDActions(ActionDataKeys.Clusters, {
     const updateableParams = 'name tags numWorkers numMinWorkers numMaxWorkers'.split(' ')
 
     const body = pick(updateableParams, params)
-    body.etcdBackup = getEtcdBackupPayload('etcdBackup', params)
```
jruddell, line 83: Does this break regular cluster updates? did you test updating a cluster and changing etcd fields?

jruddell: Maybe we should keep this line but wrap it in a conditional so we attach it to the payload when needed

knnguy:
etcdBackup is only sent in the payload when needed

Screen Shot 2020-09-21 at 1 05 50 PM

jruddell (approved):

PR #863, opened 2020-09-22
UX-593/Fixed deauthorize node bug

knnguy:
An exception occurred because there was no `node.name` or `node.primaryIp` to display in the `ConfirmationDialog` box when it was re-rendered. The node had already been deauthorized.

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/common/actions.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/common/actions.js
@@ -33,7 +33,7 @@ export const updateRemoteSupport = createContextUpdater(
       await resMgr.removeRole(id, supportRoleName)
     }
     // Reload the resMgrHosts
-    return await loadResMgrHosts()
+    return loadResMgrHosts()
   },
   {
     operation: 'updateRemoteSupport',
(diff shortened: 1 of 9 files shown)


jruddell (commented):
I don't think this is the fix for this bug. Why is the node undefined? Didn't the user just select a node and click deauthorize?

src/app/plugins/kubernetes/components/infrastructure/common/actions.js
```diff
@@ -33,7 +33,7 @@ export const updateRemoteSupport = createContextUpdater(
       await resMgr.removeRole(id, supportRoleName)
     }
     // Reload the resMgrHosts
-    await loadResMgrHosts()
+    return await loadResMgrHosts()
```
jruddell, line 36: Await is redundant here. You can omit it

src/app/plugins/kubernetes/components/infrastructure/nodes/actions.js
```diff
@@ -34,7 +34,7 @@ export const deAuthNode = createContextUpdater(
     trackEvent('Deauthorize Node', {
       node_name: node.name,
     })
-    await loadNodes()
+    return await loadNodes()
```
jruddell, line 37: Await is redundant here, omit as well

jruddell (approved):

PR #864, opened 2020-09-22
Create ActionSet class and useLoadAction hook

xantorres:


jruddell:
Putting a note here regarding our earlier conversation. It would be good to support the concept of a sub resource. We have an issue where when a cluster is unhealthy, every request to its sub resources will fail. A lot of our requests are dependent on the cluster being healthy. We can circumvent this and optimize the UI to not make requests when a resource like `clusters` is unhealthy.

PR #866, opened 2020-09-23
UX-598/Changed kubeconfig's code component from CodeBlock to CodeMirror

knnguy:
Screen Shot 2020-09-23 at 2 52 00 PM


knnguy:
Modified the styling of CodeMirror

Screen Shot 2020-09-24 at 12 05 54 PM

src/app/core/components/validatedForm/CodeMirror.js
```diff
@@ -115,7 +117,7 @@ const CodeMirror = ({
 
 CodeMirror.propTypes = {
   id: PropTypes.string.isRequired,
-  label: PropTypes.string,
+  label: PropTypes.node,
```
jruddell, line 120: This should be either string or node as the proptype.

jruddell (changes requested):
Can you create a jira ticket for this task and attach it here on the pr?

jruddell (approved):

PR #869, opened 2020-09-29
UX-604/Fixed Azure SKU selection problem

knnguy:
The problem was that for the `Region` field, we used` DisplayName` instead of` RegionName` and we weren't able to load SKUs with `DisplayName`. That is why I added a mapping for region name.

Screen Shot 2020-09-29 at 3 29 57 PM


src/app/plugins/kubernetes/components/infrastructure/clusters/AddAzureClusterPage.js
```diff
@@ -223,19 +223,29 @@ const AddAzureClusterPage = () => {
...
+  })
+
+  const [cloudProviderRegionDetails] = useDataLoader(loadCloudProviderRegionDetails, {
     cloudProviderId: params.cloudProviderId,
     cloudProviderRegionId: params.cloudProviderRegionId,
   })
-  const virtualNetworks = pathStrOr([], '0.virtualNetworks', details)
+  const virtualNetworks = pathStrOr([], '0.virtualNetworks', cloudProviderRegionDetails)
+
+  const mapRegionName = (displayName) => {
```
jruddell, line 246: you can `useMemo` here for optimization

src/app/plugins/kubernetes/components/infrastructure/clusters/AddAzureClusterPage.js
```diff
@@ -286,8 +296,10 @@ const AddAzureClusterPage = () => {
                           id="location"
                           label="Region"
                           cloudProviderId={params.cloudProviderId}
-                          onChange={getParamsUpdater('cloudProviderRegionId')}
-                          value={params.cloudProviderRegionId}
+                          onChange={(displayName) => {
```
jruddell, line 299: Don't do inline callbacks like this if you can avoid it. better to define it before the return in a `useCallback` hook

knnguy: So I used a `useCallback` hook, but didn't pass it any dependencies because it didn't seem like there were any. Maybe I'm just not understanding it right?

src/app/plugins/kubernetes/components/infrastructure/clusters/AddAzureClusterPage.js
knnguy: I'm not sure how to useMemo here. It doesn't seem to have any dependencies. `displayName` changes every time the `onChange` is called so shouldn't the `mapRegionName` be called every single time as well? Why do we need to use `useMemo`

jruddell: memoizing a function will cache the result of that function based on its parameters. This allows for one time processing, with cached results for the same parameters. So in this instance, `mapRegionName` is called, but the result is only processed once for the same `DisplayName`, if I click on a different item, this would run once for that one as well. However coming back to the same region you had previously selected, will be optimized because of memoization. 

I think you may need to use `useCallback` instead of `useMemo` though.

jruddell (approved):

PR #870, opened 2020-10-01
updated styling for input elements and buttons, fixed login page

jruddell:
Generic button component.
Generic input element.

image


kevinylee1991 (approved):
lgtm, cool this addresses that input light variant I was wondering about

PR #872, opened 2020-10-02
UX-600/Added tabs for ApiAccessPage and fixed ApiHelper

knnguy:
**API Tab**

Screen Shot 2020-10-02 at 6 24 07 PM

**KubeConfig Tab**

Screen Shot 2020-10-02 at 6 24 22 PM

**Terraform Tab**

Screen Shot 2020-10-02 at 6 24 32 PM


knnguy:
For the ApiHelper, this is what I tested:

- GET, PUT, POST, and DELETE for qbert
- GET, DELETE for nova

I couldn't test the rest of the services since I don't have the API docs for them, but if the API calls for qbert works, it should work with the rest of the services as well

jruddell:
Lets think about how we want to expose the api helper more. I think we should make it more intuitive for the user, take the implicit knowledge they need to know away. This way it'll be easier to consume and more helpful vs difficult to figure out.

PR #873, opened 2020-10-03
Document Meta

jruddell:
Generic document meta component to update the title of the page plus some extra functionality around body class and id manipulation

page-title-change

___

- ### Title Manipulation
```
<DocumentMeta title="Clusters" />
```
image

 - ### Body class manipulation for forms with a gray background
```
<DocumentMeta title="Add Cluster" bodyClasses={['form-view']} />
```
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "react-redux": "^7.1.3",
     "react-router": "^5.0.1",
     "react-router-dom": "^5.0.1",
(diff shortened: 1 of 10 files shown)


kevinylee1991 (approved):
Looks good. Just putting a note down regarding the potentially having a default DocumentMeta for when switching to views that don't have one defined

PR #877, opened 2020-10-08
UX-134/Show empty bar for deployed capacity for cloud providers with …

kevinylee1991:
…no capacity data

Screen Shot 2020-10-08 at 2 57 06 PM

Screen Shot 2020-10-08 at 3 36 37 PM


jruddell (changes requested):
Lets make this more reusable. This component is used on many views but re-defined in each view. Also make sure this change is on the nodes list page as well.

src/app/plugins/kubernetes/components/infrastructure/nodes/ConvergingNodeBreakdown.tsx
```diff
@@ -133,25 +133,8 @@ const useStyles = makeStyles<Theme, {}>((theme) => ({
...
-        valueConverter={toMHz}
-        units="MHz"
-        label="CPU"
-        stats={usage?.compute}
-      />
-      <ResourceUsageTable valueOff units="GiB" label="Memory" stats={usage?.memory} />
-      <ResourceUsageTable valueOff units="GiB" label="Storage" stats={usage?.disk} />
-    </>
-  )
+  return <ResourceUsageTables usage={usage} valueOff={true} />
```
jruddell, line 137: you don't need to specify true on a boolean prop. `<ResourceUsageTables valueOff usage={usage} />`

src/app/plugins/kubernetes/components/infrastructure/nodes/selectors.ts
```diff
@@ -25,15 +26,24 @@ export const nodesSelector = createSelector(
...
-      // if hostagent is not responding, then the nodes info is outdated
-      // set the status to disconnected manually
-      status: combinedHostsObj[node.uuid].responding ? node.status : 'disconnected',
-      combined: combinedHostsObj[node.uuid],
-      // qbert v3 link fails authorization so we have to use v1 link for logs
-      logs: `${qbertUrl}/logs/${node.uuid}`.replace(/v3/, 'v1'),
-    }))
+    return rawNodes.map((node) => {
+      const calcNodeUsage = calcUsageTotalByPath([combinedHostsObj[node.uuid]]) // expects array
+      const usage = {
```
jruddell, line 31: it probably makes sense to make a method that you can use in both selectors to generate this object considering they're the same. 

`const usage = getResourceUsage([combinedHostsObj[node.uuid]])`

jruddell (approved):
👍

PR #878, opened 2020-10-09
UX-623, UX-274/Fix cloud provider update UX to match API limitations

kevinylee1991:


src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AwsCloudProviderFields.tsx
```diff
@@ -137,7 +139,7 @@ const AwsCloudProviderFields = ({
         value={wizardContext.awsSecretKey}
         info="IAM User Secret Key"
         type="password"
-        required
+        required={optionalFields ? false : true}
```
jruddell, line 142: you don't need a ternary operator here. `required={!optionalFields}`

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/UpdateCloudProviderPage.tsx
```diff
@@ -54,20 +54,21 @@ const useStyles = makeStyles((theme: Theme) => ({
...
 
 const formCpBody = (data) => {
+  // Do not accept empty strings for these properties
+  // User may type and then delete the input before submitting
   if (data.type === CloudProviders.Aws) {
     return {
       name: data.name,
-      key: data.awsAccessKey,
-      secret: data.awsSecretKey,
+      key: !!data.awsAccessKey ? data.awsAccessKey : undefined,
```
jruddell, line 62: this is a typescript file so just use the optional chaining operator. Also include it on the `name` as that should be an optional parameter on edit

```
return {
  name: data?.name,
  key: data?.awsAccessKey,
  secret: data?.awsSecretKey,
}
```

jruddell (changes requested):
Few optimizations / cleaner code

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/UpdateCloudProviderPage.tsx
kevinylee1991: I figured for the name since it gets prepopulated by the update form that it would never really be optional. I also have it as a required field still so that they don't rename their cloud provider as empty string. Thoughts?

kevinylee1991: I tried using data?.awsAccessKey etc. but the UI sent empty strings in the request when I filled in the fields and deleted them. Not sure if there is a good other way to set undefined.

jruddell: you shouldn't be passing any form fields when they are empty. Someone may want to update their credentials without changing the name, or change the name without changing the credentials. If a field is empty it shouldn't be included as a key in the update call

src/app/core/components/wizard/Wizard.js
```diff
@@ -81,7 +81,7 @@ class Wizard extends PureComponent {
         ...this.getActiveStepId(state, state.activeStep + 1),
       }),
       () => {
-        if (steps[activeStep].onNext) {
+        if (steps[activeStep]?.onNext) {
```
jruddell, line 84: this is a javascript file, this'll be a problem

src/app/plugins/kubernetes/components/common/CloudProviderCard.tsx
```diff
@@ -65,7 +74,7 @@ const CloudProviderCard: FC<Props> = (props) => {
     history.push(src)
   }
   return (
-    <div className={classes.root} onClick={handleClick}>
+    <div className={clsx(classes.root, className)} onClick={handleClick}>
```
jruddell, line 77: we shouldn't need to pass class names to these cards. use css grid to space them out (`grid-gap`)

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AddCloudProviderCredentialStep.tsx
```diff
@@ -97,12 +100,13 @@ const AddCloudProviderCredentialStep = ({
...
       const [success, newCp] = await cloudProviderActions.create(body)
       if (!success) {
+        // TODO: surface the real API response error -- how to do this properly?
         throw 'Error creating cloud provider'
       }
       setWizardContext({ cloudProviderId: newCp.uuid })
     } catch (err) {
       setSubmitting(false)
-      showToast(err.message, MessageTypes.error)
+      showToast(err, MessageTypes.error)
```
jruddell, line 109: instead of showing a toast can we integrate the error on the page? just like you and I did for downloading the IAM policy (I think).

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AzureCloudProviderFields.tsx
```diff
@@ -53,18 +58,22 @@ const AzureCloudProviderFields = ({
...
         label="Subscription ID"
         onChange={(value) => setWizardContext({ subscriptionId: value })}
         info="The ID of the subscription that correlates to the service principal"
+        disabled={updateWizard}
         required
       />
       {showSubmitInCard && (
         <div className={inCardSubmit}>
-          <SubmitButton noMargin>Update Cloud Provider</SubmitButton>
+          <Button disabled type="submit">
```
jruddell, line 74: is this button always supposed to be disabled?

jruddell (changes requested):
I'd like us all to utilize css grid more. when you create elements that are opinionated on their margins you end up needing to override that in later use cases. css grid lets you space things out by defining a `gap`. This gives much more flexibility in layouts and also removes the spacing concerns from reusable components.

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AzureCloudProviderFields.tsx
kevinylee1991: yes for now for azure update until the API is improved

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AddCloudProviderCredentialStep.tsx
kevinylee1991: Screen Shot 2020-10-15 at 6 43 10 PM

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AddCloudProviderCredentialStep.tsx
```diff
@@ -97,12 +96,13 @@ const AddCloudProviderCredentialStep = ({
       const body = formCpBody(wizardContext)
       const [success, newCp] = await cloudProviderActions.create(body)
       if (!success) {
+        // TODO: surface the real API response error -- how to do this properly?
```
jruddell, line 99: dont need this todo anymore since you are tracking and surfacing the error in the form now

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AwsCloudProviderFields.tsx
```diff
@@ -120,25 +125,27 @@ const AwsCloudProviderFields = ({
...
         label="Secret Access Key"
-        onChange={(value) => setWizardContext({ awsSecretKey: value })}
-        value={wizardContext.awsSecretKey}
+        onChange={(value) => setWizardContext({ secret: value })}
+        value={wizardContext.secret}
         info="IAM User Secret Key"
         type="password"
         required
       />
+      <ErrorMessage>{errorMessage}</ErrorMessage>
```
jruddell, line 148: dont render an empty element if theres no error message.

jruddell (approved):
Great work!

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AddCloudProviderCredentialStep.tsx
kevinylee1991: regarding this I think it would still be more ideal for the UI to surface the API request error (user can know exactly what went wrong) but I'll phrase the todo better

PR #880, opened 2020-10-13
UX-615/remove the onboarding workflow

jruddell:
The current onboarding workflow is not very useful and we're seeing a lot of users never see the actual dashboard because they are always on onboarding.


knnguy (approved):

kevinylee1991 (approved):
I'm always on onboarding lol

PR #884, opened 2020-10-14
UX-611/Take into account all health statuses for infrastructure overv…

kevinylee1991:
…iew card and slight code cleanup

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatusUtils.ts
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatusUtils.ts
@@ -1,4 +1,10 @@
-import { ConnectionStatus, HealthStatus, IClusterSelector, IClusterStatus, TransientStatus } from './model'
+import {
+  ConnectionStatus,
+  HealthStatus,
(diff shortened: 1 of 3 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatusUtils.ts
```diff
@@ -312,7 +328,7 @@ export const clusterHealthStatusFields: {
     label: 'Unhealthy',
   },
   unknown: {
-    status: 'pause',
+    status: 'unknown',
```
jruddell, line 331: I would verify that this pause status isn't being used anywhere. I think maybe pods would show paused?

jruddell (approved):
assuming the pause status isn't being used then this is good to go!

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatusUtils.ts
kevinylee1991: It's just used for color, looks like all the places it's used looks better with the new status

PR #886, opened 2020-10-15
UX-626/Bugsnag source maps and send user info along with api URL and response metadata

xantorres:


webpack.config.js
```diff
@@ -109,9 +111,15 @@ module.exports = {
             },
           }),
         ]
-      : []),
+      : [
+          new BugsnagSourceMapUploaderPlugin({
+            apiKey: '3eb58c77ada2a9db70fc7a8e81e97b99',
```
jruddell, line 116: can you make this api key an environment variable? `process.env.BUGSNAG_KEY` or at least a constant so it can be reused here and when we `Bugsnag.start`?

webpack.config.js
```diff
@@ -8,10 +8,12 @@ const TerserPlugin = require('terser-webpack-plugin')
 const contextPath = path.resolve(__dirname, './src/app')
 const contentBase = path.resolve(__dirname, './build')
 const outputPath = path.resolve(contentBase, './ui')
+const { BugsnagSourceMapUploaderPlugin } = require('webpack-bugsnag-plugins')
 
 const env = process.env.NODE_ENV || 'development'
 const isDev = env === 'development'
 const isProd = env === 'production'
+const { version } = JSON.stringify(require('./package.json'))
```
jruddell, line 16: I believe this is invalid, probably dont need the stringify correct?

jruddell (approved):
🚢

PR #887, opened 2020-10-16
UX-630, UX-648/Wrap wizard steppers, fix OpenStack RC step of ironic …

kevinylee1991:
…setup wizard, added form DocumentMeta to ironic and cluster forms

Before
Screen Shot 2020-10-16 at 12 58 47 PM
After
Screen Shot 2020-10-16 at 1 41 40 PM

### DIFF
--- a/src/app/core/components/validatedForm/FormFieldCard.tsx
+++ b/src/app/core/components/validatedForm/FormFieldCard.tsx
@@ -26,6 +26,7 @@ export const useStyles = makeStyles<Theme, ContainerProps>((theme) => ({
     justifyContent: 'space-between',
     alignItems: 'stretch',
     color: theme.palette.grey[700],
+    marginBottom: theme.spacing(2),
   },
   inputWidth: {
     maxWidth: ({ maxWidth = defaultMaxWidth, inputsWidth = maxWidth / 2 }) => inputsWidth,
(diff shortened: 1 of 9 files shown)


src/app/core/components/validatedForm/FormFieldCard.tsx
```diff
@@ -26,6 +26,7 @@ export const useStyles = makeStyles<Theme, ContainerProps>((theme) => ({
     justifyContent: 'space-between',
     alignItems: 'stretch',
     color: theme.palette.grey[700],
+    marginBottom: theme.spacing(2),
```
jruddell, line 29: why are we adding margins to all of these components? this will effect other instances where we use them. Please make the component rendering these define the layout

src/app/core/components/wizard/WizardStepper.js
```diff
@@ -69,6 +69,8 @@ const useWizardStepperStyles = makeStyles((theme) => ({
   stepperRoot: {
     backgroundColor: 'transparent',
     padding: 0,
+    flexWrap: 'wrap',
+    marginBottom: theme.spacing(4),
```
jruddell, line 73: does the wizard stepper need this bottom margin?

jruddell (approved):

PR #888, opened 2020-10-17
UX-600/Created API Helper

knnguy:
**Exposed API Services**

- Qbert
- Keystone
- Appbert (only if localStorage.enableDevPlugin === true)
- Resmgr (only if localStorage.enableDevPlugin === true)

**DevTools ApiHelper**

- Works now (Couldn't make requests before)

API Details Page

**Search bar allows user to search by method name**

- User can also filter by request verb (PUT, PATCH, GET, POST, DELETE)

**API Request Helper**

- ClusterUuid and CloudProviderUuid params have a dropdown menu so users can select an options vs. having to type it out

- The URL field will dynamically show the the url with the parameters that the user enters

- The Make API Request button and the JSON response CodeMirror will only show up on GET methods which means that making PUT, PATCH, DELETE, and POST requests will be disabled for users. However, if dev plugin is enabled, then everything is allowed


src/app/core/components/Filter.tsx
```diff
@@ -0,0 +1,127 @@
...
+}))
+
+const Filter = ({ data, setFilteredData, filters, searchTarget }: Props) => {
+  const classes = useStyles()
+  const [searchTerm, setSearchTerm] = React.useState('')
+  const [filterProperties, setFilterProperties] = React.useState({})
+
+  const { history, location } = useReactRouter()
+  const url = Route.getCurrentRoute().url
+  const route = new Route({ url, name: 'Filter' })
```
jruddell, line 50: why do you need to make a new route here?

src/app/core/components/Filter.tsx
```diff
@@ -0,0 +1,127 @@
...
+      setSearchTerm(searchTerm)
+    }
+
+    filters.map((filter) => {
+      const param = searchParams.get(filter.target)
+      if (param) {
+        setFilterProperties(assocPath([filter.target], param))
+      }
+    })
+  }, [data])
```
jruddell, line 65: aren't `filters` a dependency of this `useEffect`?

src/app/core/components/Filter.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const params = Object.assign({}, filterProperties)
+    if (!!searchTerm) {
+      params['search'] = searchTerm
+    }
+    const urlWithQueryParams = route.path(params)
+    history.push(urlWithQueryParams)
+  }
+
+  const filterBySearch = (data) => {
+    return data.filter((data) => data[searchTarget].match(new RegExp(searchTerm, 'i')) !== null)
```
jruddell, line 83: it is a bad practice to reuse the same variable name (`data` in this case), most linters error out here, ours should be as well... I'll have to look into that. For other variable names, I usually go with `items`, `item`, `data` then get more creative after that

src/app/plugins/kubernetes/components/apiAccess/ApiAccessListPage.js
```diff
@@ -1,24 +1,8 @@
...
 const ApiAccessListPage = () => {
-  const { container } = useStyles()
-  return (
-    <div className={container}>
-      <EnpointsListPage />
-      <Divider />
-      <KubeConfigListPage />
-    </div>
-  )
+  return <ApiServicesPage />
```
jruddell, line 5: why do we need a component to just render another component? instead of rendering `ApiAccessListPage` just render what it is `<ApiServicesPage />`. then you can delete this file :)

src/app/plugins/kubernetes/components/apiAccess/ApiAccessPage.js
```diff
@@ -1,17 +1,25 @@
...
+import { FormFieldCard } from 'core/components/validatedForm/FormFieldCard'
+import TerraformListPage from './TerraformListPage'
 
 const ApiAccessPage = () => (
   <PageContainer>
-    <ApiAccessListPage />
-    {/* <Tabs>
-      <Tab value="apiAccess" label="API Access">
+    <Tabs>
+      <Tab value="api" label="API">
```
jruddell, line 13: label this tab `API Services` and update the route name to reflect this too

src/app/plugins/kubernetes/components/apiAccess/ApiDetailsPage.tsx
```diff
@@ -0,0 +1,151 @@
...
+  resmgr: ResMgr,
+}
+
+const ApiDetailsPage = ({ service }) => {
+  const [activeCardMetadata, setMetadata] = React.useState(null)
+  const [filteredMethods, setFilteredMethods] = React.useState([])
+  const classes = useStyles()
+  const methods = apiServices[service].apiMethodsMetadata
+
+  const filters: Filter[] = [
```
jruddell, line 107: constants shouldn't be defined in the body of a react functional component. Its a constant so define it outside of the component.

src/app/plugins/kubernetes/components/apiAccess/ApiRequestHelper.tsx
```diff
@@ -0,0 +1,252 @@
...
+        onChange={getParamsUpdater(paramName)}
+        required
+      />
+    )
+  }
+}
+
+const getServiceEndpoint = async (service, apiClient) => {
+  if (service === 'keystone') {
+    return apiClient.options.keystoneEndpoint
```
jruddell, line 116: this is really messy. You are returning a promise sometimes and other times you aren't. You should always have a consistent return type in your functions. Aside from that, you shouldn't be doing this work here. Each `basic<VERB>` request will automatically add the endpoint for you. Look at the implementation of `basicGet` for example.

src/app/plugins/kubernetes/components/apiAccess/ApiRequestHelper.tsx
```diff
@@ -0,0 +1,252 @@
...
+        POST: () => apiClient.basicPost(params),
+        PATCH: () => apiClient.basicPatch(params),
+        PUT: () => apiClient.basicPut(params),
+        DELETE: () => apiClient.basicDelete(params),
+      }[metadata.type]()
+
+      return response
+    } catch (e) {
+      typeof e === 'object'
+        ? showToast(e.message, MessageTypes.error)
```
jruddell, line 175: lets not show toast messages for errors anymore. Instead we should integrate the error on the page. Use a state variable for it and show it either right above or below the button that triggers the api call

src/app/plugins/kubernetes/components/apiAccess/ApiRequestHelper.tsx
```diff
@@ -0,0 +1,252 @@
...
+                </Text>
+                {metadata.params.map((paramName) =>
+                  renderParamField({ classes, paramName, params, getParamsUpdater }),
+                )}
+              </div>
+            )}
+            {methodsWithBody.includes(metadata.type) && devEnabled && (
+              <TextField className={classes.textField} id="body" label="Body" multiline rows={3} />
+            )}
+            {(devEnabled ||
```
jruddell, line 227: Anytime there is a complicated set of flags used to determine when to render something or not I opt to define that logic in a variable. This makes reading the HTML structure much cleaner and easier to follow.

src/app/core/components/Filter.tsx
knnguy: In order to update the current URL to one with query parameters from `Filter`, I need to use Route's `path` method. The path method constructs a URL with query parameters.

Screen Shot 2020-10-18 at 6 57 41 PM

src/app/plugins/kubernetes/components/apiAccess/ApiAccessListPage.js
knnguy: ApiAccessListPage is in `/openstack/index.js`. Do I remove it from there as well?

src/app/core/components/Filter.tsx
jruddell: The line before you get the route, but only use the url. Why don't you just use the current routes path method?

src/app/core/components/Filter.tsx
knnguy: Actually, I have changed the dependency of this `useEffect` to [] so that it only renders once.

In this `useEffect`, I'm pulling the filter and search params from the URL to set `Filter` state. I believe this only needs to be done on the first render. This is the case where if a user were to refresh the page, the filters and search term will be re-populated.

src/api-client/helpers.ts
```diff
@@ -75,3 +77,18 @@ export const normalizeResponse = <T>(response): T extends ResponseWithData ? T['
...
+  methodName: string,
+) => {
+  target.constructor.apiMethodsMetadata.push({
+    name: methodName,
+    params: propOr([], 'params', metadata),
+    ...metadata,
+  })
+}
+
+export const replaceTextBetweenCurlyBraces = (str, replacements) => {
```
jruddell, line 92: We have a method for doing this inside our Route class. Lets use that please

jruddell:
updated the design

Terraform Page
image

Kubeconfig Page
image

Api Services Page 
image

When clicking on a method a modal appears for you to test that API
image
image

jruddell (approved):

PR #898, opened 2020-10-31
UX-667/Added account plugin, second header, changed navbar stack update logic, new styling on avatar dropdown, moved user management into new plugin directory

kevinylee1991:
Screen Shot 2020-10-30 at 6 25 39 PM
Screen Shot 2020-10-30 at 6 33 44 PM


cypress/integration/tenants.spec.js
```diff
@@ -2,7 +2,7 @@ describe('tenants', () => {
   before(() => {
     cy.resetServerContext('dev')
     cy.login()
-    cy.visit('/ui/kubernetes/user_management#tenants')
+    cy.visit('/ui/account/user_management#tenants')
```
jruddell, line 5: can we call this `/ui/me/...`

src/app/core/containers/AuthenticatedContainer.tsx
```diff
@@ -200,11 +212,24 @@ const redirectToAppropriateStack = (ironicEnabled, kubernetesEnabled, history) =
   }
 }
 
-const determineCurrentStack = (history, features) => {
-  if (features.metalstack) {
+const determineCurrentStack = (location, features, lastStack) => {
+  // first try the pathname, then use the saved session, else default to kubernetes
+  if (location.pathname.includes('account')) {
```
jruddell, line 217: Would it be better to make the current stack a dynamic parameter in React Router and then just grab it here?

src/app/plugins/account/components/secondHeader/MyAccountHeader.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const {
+    username,
+    userDetails: { displayName },
+  } = session
+
+  return (
+    <div className={clsx(classes.root, className)}>
+      <div className={classes.upgrade}></div>
+      <div className={classes.userDetails}>
+        <div className={classes.userIcon}>{displayName[0]}</div>
```
jruddell, line 65: Yea, lets definitely make this an avatar component.

src/app/core/components/UserMenu.js
```diff
@@ -67,19 +115,22 @@ class UserMenu extends React.PureComponent {
...
       return <ChangePasswordModal onCancel={this.handleCancelChangePassword} />
     }
 
     return (
       <div className={`${classes.avatar} ${className}`}>
-        <Text color="inherit" variant="body2" className={classes.title} onClick={this.handleClick}>
-          {username} &#9662;
-        </Text>
+        <div onClick={this.handleClick} className={clsx(classes.userIcon, classes.toolbarIcon)}>
+          {displayName[0]}
```
jruddell, line 132: Is this the avatar icon? Lets make a specific component for the avatar so it can be reused in other places. Also, when getting the first character use `str.charAt(0)` instead. The reason why its better to use this is it will return an empty string instead of undefined. Keeps the data type consistent.

src/app/plugins/account/components/secondHeader/MyAccountHeader.tsx
```diff
@@ -0,0 +1,77 @@
...
+  upgrade: {
+    width: 180,
+  },
+  userDetails: {
+    paddingBottom: theme.spacing(2),
+    flexGrow: 1,
+    display: 'flex',
+    alignItems: 'flex-end',
+    justifyContent: 'space-between',
+    maxWidth: '75%',
```
jruddell, line 25: percentage widths generally will not work for all use cases. What is this supposed to be doing in the context of the page? I can help if you need some direction

src/app/plugins/account/components/secondHeader/MyAccountHeader.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { SessionState, sessionStoreKey } from 'core/session/sessionReducers'
+import { RootState } from 'app/store'
+import { prop } from 'ramda'
+import Text from 'core/elements/text'
+
+const useStyles = makeStyles<Theme>((theme: Theme) => ({
+  root: {
+    height: 96,
+    background: theme.palette.grey[900],
+    display: 'flex',
```
jruddell, line 14: I think the main issue you're having is because this is flex instead of grid. If you use columns here and let the parent define the layout, it can specify `1fr` for the `userDetails` section

src/app/plugins/account/components/userManagement/tenants/AddTenantPage.js
```diff
@@ -6,16 +6,16 @@ import WizardStep from 'core/components/wizard/WizardStep'
...
-import { mngmTenantActions } from 'k8s/components/userManagement/tenants/actions'
-import { mngmUserActions } from 'k8s/components/userManagement/users/actions'
+import UserRolesTableField from 'account/components/userManagement/tenants/UserRolesTableField'
+import { mngmTenantActions } from 'account/components/userManagement/tenants/actions'
+import { mngmUserActions } from 'account/components/userManagement/users/actions'
 import Progress from 'core/components/progress/Progress'
 import { requiredValidator } from 'core/utils/fieldValidators'
 
-const listUrl = pathJoin(k8sPrefix, 'user_management')
+const listUrl = pathJoin(userAccountPrefix, 'user_management')
```
jruddell, line 18: we shouldn't be defining urls anywhere except our `routes.tsx` file. If you find places where they're being defined like here go ahead and refactor it please

src/app/plugins/account/components/userManagement/tenants/TenantsListPage.js
```diff
@@ -1,13 +1,13 @@
...
-import { k8sPrefix } from 'app/constants'
+import { mngmTenantActions } from 'account/components/userManagement/tenants/actions'
+import { userAccountPrefix } from 'app/constants'
 import { pathJoin } from 'utils/misc'
 
 const isNotServiceTenant = (tenants) => !tenants.find((t) => t.name === 'service')
 
 export const options = {
-  addUrl: '/ui/kubernetes/user_management/tenants/add',
+  addUrl: '/ui/account/user_management/tenants/add',
```
jruddell, line 10: update these urls to be from our `routes` file

src/app/plugins/account/components/userManagement/users/UsersListPage.js
```diff
@@ -48,9 +48,9 @@ export const options = {
     { id: 'tenants', label: 'Tenants', render: pipe(arrayIfNil, pluck('name'), join(', ')) },
   ],
   addText: 'Create a new User',
-  addUrl: pathJoin(k8sPrefix, 'user_management/users/add'),
+  addUrl: pathJoin(userAccountPrefix, 'user_management/users/add'),
```
jruddell, line 51: update these urls to pull from our `routes` file

src/app/plugins/account/index.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react'
+import UserManagementIndexPage from 'account/components/userManagement/UserManagementIndexPage'
+import AddTenantPage from 'account/components/userManagement/tenants/AddTenantPage'
+import EditTenantPage from 'account/components/userManagement/tenants/EditTenantPage'
+import AddUserPage from 'account/components/userManagement/users/AddUserPage'
+import EditUserPage from 'account/components/userManagement/users/EditUserPage'
+import MyAccountHeader from 'account/components/secondHeader/MyAccountHeader'
+
+class Account extends React.PureComponent {
```
jruddell, line 9: lets go with `MyAccount` for naming purposes. Account sounds too generic. What do you think?

src/app/plugins/account/index.js
```diff
@@ -0,0 +1,76 @@
...
+    {
+      name: 'Add Tenant',
+      requiredRoles: 'admin',
+      link: { path: '/user_management/tenants/add', exact: true },
+      component: AddTenantPage,
+    },
+    {
+      name: 'Edit Tenant',
+      requiredRoles: 'admin',
+      link: { path: '/user_management/tenants/edit/:id', exact: true },
```
jruddell, line 38: Technically the `Route` class has a `toString` method on it that prints out its url with the dynamic variable in it. Maybe it would be good to use that here so we are really truly defining the route in one place

kevinylee1991: I considered this but the Route class includes the prefix as well (the /ui/my-account part), how would you feel about using the Route class and then stripping the prefix here?

src/app/plugins/account/components/secondHeader/MyAccountHeader.tsx
kevinylee1991: https://app.zeplin.io/project/5ef3c544d588919f0165e0f0/screen/5f985967463eaa452d54c387

Using this as a reference, it gets aligned with the form. But not all screens have a form, some will span the entire screen like the list pages for example. Also not sure how we would handle or if we even should handle large monitor sizes. So I really wasn't sold on any solution here. Any thoughts?

src/app/core/containers/AuthenticatedContainer.tsx
kevinylee1991: That could work... and then since we have URLs like /ui/help we could I guess either whitelist or blacklist that parameter to handle special cases?

src/app/plugins/account/components/secondHeader/MyAccountHeader.tsx
kevinylee1991: Going to wait on above comment before resolving

src/app/core/containers/AuthenticatedContainer.tsx
jruddell: we dont need to do that, just leave it as a dynamic parameter. If it matches specific ones we want to redirect somewhere, on others we just dont care what it is. we still are defining the help route so it shouldn't matter

src/app/plugins/account/components/secondHeader/MyAccountHeader.tsx
jruddell: If you make the window large percentage widths will not layout well. You can just do a max width of like 1k px or something instead. But again I would let the parent define how it should be laid out. `grid-template-columns: <sidebar width> minmax(max-content, 1100px)`

src/app/plugins/account/index.js
jruddell: Yea I would prefer that. We define the prefix here, put that in a variable and then reuse to remove. Or even better pass an option to the toString method that does it for you (callback fn or something)

src/app/core/containers/AuthenticatedContainer.tsx
kevinylee1991: Looks like this is not possible without some workarounds since the url params are not available to the AuthenticatedContainer component. Since it's so close to the base of the dom branch the match object only ever processes up to the root path / and never goes further. The issue I encountered trying to get the dynamic parameter to work is described here

https://github.com/ReactTraining/react-router/issues/5870

src/app/constants.js
```diff
@@ -30,6 +30,7 @@ export const uuidRegex = new RegExp(
 export const originUsernameRegex = new RegExp(/:\/\/(?<originUsername>(.+?)).platform9./, 'i')
 
 export const k8sPrefix = `${appUrlRoot}/kubernetes`
+export const userAccountPrefix = `${appUrlRoot}/my-account`
```
jruddell, line 33: do we need these base url prefix's? where are they needed to be used?

jruddell (approved):

PR #899, opened 2020-11-02
UX-433, UX-495/ Modified scaling masters constraints. Allow user to scale master nodes 1 at a time. Changed cluster connection and health status to show loading icon loading.

knnguy:
**Changes:**
- User can only scale master nodes one at a time

- User cannot scale master nodes while there are master nodes converging

- Master nodes can only be scaled when all master nodes in the cluster are healthy

- BareOs clusters with a single master node can only be scaled up with a virtual IP

Screen Shot 2020-11-03 at 2 27 49 PM

**Non-related changes:**

- Loading icon is now shown when cluster is converging

Screen Shot 2020-11-02 at 5 05 39 PM

- Removed top margin from cloud provider verification page

Screen Shot 2020-11-04 at 2 10 05 PM

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatus.tsx
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatus.tsx
@@ -135,6 +135,7 @@ interface IClusterStatusProps {
   cluster: IClusterSelector
   variant?: StatusVariant
   message?: string
+  iconStatus?: boolean
 }
 
 export const ClusterHealthStatus: FC<IClusterStatusProps> = ({
@@ -156,7 +157,6 @@ export const ClusterHealthStatus: FC<IClusterStatusProps> = ({
           title={fields.message}
           status={fields.status}
           variant={variant}
-          iconStatus={fields.status === 'loading' ? true : false}
           {...rest}
         >
           {variant === 'header' ? (
@@ -195,7 +195,6 @@ export const ClusterConnectionStatus: FC<IClusterStatusProps> = ({
       title={fields.message}
       status={fields.clusterStatus}
       variant={variant}
-      iconStatus={fields.clusterStatus === 'loading' ? true : false}
       {...rest}
     >
       {variant === 'header' ? (
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatus.tsx
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatus.tsx
@@ -133,7 +133,7 @@ const renderTransientStatus = ({ uuid, connectionStatus }, variant) => {
 
 interface IClusterStatusProps {
   cluster: IClusterSelector
-  variant: StatusVariant
+  variant?: StatusVariant
   message?: string
 }
 
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatus.tsx
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatus.tsx
@@ -156,6 +156,7 @@ export const ClusterHealthStatus: FC<IClusterStatusProps> = ({
           title={fields.message}
           status={fields.status}
           variant={variant}
+          iconStatus={fields.status === 'loading' ? true : false}
           {...rest}
         >
           {variant === 'header' ? (
@@ -194,6 +195,7 @@ export const ClusterConnectionStatus: FC<IClusterStatusProps> = ({
       title={fields.message}
       status={fields.clusterStatus}
       variant={variant}
+      iconStatus={fields.clusterStatus === 'loading' ? true : false}
       {...rest}
     >
       {variant === 'header' ? (
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
@@ -27,6 +27,8 @@ import { cloudProviderTypes } from '../cloudProviders/selectors'
 import ClusterDeleteDialog from './ClusterDeleteDialog'
 import DownloadKubeConfigLink from './DownloadKubeConfigLink'
 import ResourceUsageTables from '../common/ResourceUsageTables'
+import { CloudProviders } from '../cloudProviders/model'
+import { hasConvergingNodes } from './ClusterStatusUtils'
 
 const useStyles = makeStyles((theme) => ({
   links: {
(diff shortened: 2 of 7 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/ScaleMastersPage.tsx
```diff
@@ -110,27 +134,49 @@ const ScaleMasters: FunctionComponent<ScaleMasterProps> = ({
...
-      return transitionConstraint.relation === 'allow'
+      // BareOs clusters with single master node cannot scale without a virtial IP
+      if (
+        cluster.cloudProviderType === CloudProviders.BareOS &&
+        numMasters === 1 &&
+        !hasMasterVip
+      ) {
+        if (validatorRef.current) {
+          validatorRef.current.errorMessage =
+            'You cannot add master nodes to a single master cluster without a virtual IP.  You need to create a multi-master cluster with at least 2 masters before you can add more masters to the cluster.'
```
jruddell, line 172: This error message is a bit confusing. can you ask Chris to write up something better? it should say something about creating a cluster with at least one node and a VIP.

jruddell: Actually, it would be better to make a separate validator to handle this logic.

src/app/plugins/kubernetes/components/infrastructure/clusters/ScaleMastersPage.tsx
```diff
@@ -165,15 +211,21 @@ const ScaleMasters: FunctionComponent<ScaleMasterProps> = ({
...
                 ? isUnassignedNode
                 : allPass([isMaster, inCluster(cluster.uuid)])
             }
             validations={[validatorRef.current]}
+            onChange={(value) => setSelectedNode(value)}
             required
           />
+          {showWarningMessage && (
+            <div className={classes.warning}>
+              <Alert variant={'warning'} message={transitionConstraint.message} />
```
jruddell, line 226: pass `small` as a property to the alert to remove excess padding and spacing

src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.js
```diff
@@ -131,9 +133,12 @@ const renderMetaData = (_, { tags }) => {
 }
 
 const canScaleMasters = ([cluster]) =>
+  cluster.cloudProviderType === CloudProviders.BareOS &&
```
jruddell, line 136: Lets break this into smaller more composable items. For instance you can have a `isHealthyCluster` that checks the clusters health and the nodes health.

If you move this to a typescript file (convert this file or move to `helpers.tsx`) you can use optional chaining to avoid the `\\ []` (`cluster?.nodes?.length > 1`). Also equality comparisons should always be strict `===` or `!==`. In this instance you should just cast it to a boolean `!!cluster.masterVipIpv4`

src/app/plugins/kubernetes/components/infrastructure/clusters/ScaleMastersPage.tsx
```diff
@@ -111,30 +140,46 @@ const ScaleMasters: FunctionComponent<ScaleMasterProps> = ({
...
+    ) || ({} as IConstraint)
+
+  useEffect(() => {
+    setSelectedNode(null)
+  }, [params.scaleType])
+
+  const clusterAndNodeStatusValidator = useCallback(() => {
+    return customValidator(() => {
+      return (
+        cluster.status === 'ok' &&
```
jruddell, line 163: reuse the same helpers here

src/app/plugins/kubernetes/components/infrastructure/clusters/ScaleMastersPage.tsx
```diff
@@ -111,30 +140,46 @@ const ScaleMasters: FunctionComponent<ScaleMasterProps> = ({
...
       return transitionConstraint.relation === 'allow'
     }, defaultErrorMessage)
-  }, [numMasters, params, validatorRef])()
+  }, [cluster, params])()
 
-  validatorRef.current = masterNodeLengthValidator
+  const bareOsValidator = useCallback(() => {
+    return customValidator(() => {
+      // BareOs clusters with single master node cannot scale without a virtial IP
+      return cluster.cloudProviderType === CloudProviders.BareOS && (numMasters > 1 || hasMasterVip)
```
jruddell, line 180: make this a helper as well so were not writing out this logic in multiple places

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatus.tsx
```diff
@@ -133,7 +133,7 @@ const renderTransientStatus = ({ uuid, connectionStatus }, variant) => {
 
 interface IClusterStatusProps {
   cluster: IClusterSelector
-  variant: StatusVariant
+  variant?: StatusVariant
```
jruddell, line 136: why is variant optional?

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatus.tsx
```diff
@@ -156,6 +156,7 @@ export const ClusterHealthStatus: FC<IClusterStatusProps> = ({
           title={fields.message}
           status={fields.status}
           variant={variant}
+          iconStatus={fields.status === 'loading' ? true : false}
```
jruddell, line 159: I dont think we want to do this here. Look at places where we use the cluster status span, some places pass through a flag `iconStatus` for other states than `loading`. The cluster list page when rendering the status should pass `iconStatus` as it needs to

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterStatus.tsx
knnguy: I made it optional to fix a TypeScript error. It should be optional since we defined a default value for it.

Screen Shot 2020-11-03 at 2 50 04 PM

src/app/plugins/kubernetes/components/infrastructure/clusters/helpers.ts
```diff
@@ -1,12 +1,39 @@
...
 
+export const clusterIsHealthy = (cluster) =>
+  cluster.status === 'ok' && cluster.masterNodesHealthStatus === 'healthy'
+
+export const clusterNotBusy = (cluster) =>
+  cluster.taskStatus === 'success' && !hasConvergingNodes(cluster.nodes)
+
+export const isBareOsMultiMasterCluster = (cluster) =>
+  cluster.cloudProviderType === CloudProviders.BareOS &&
+  (cluster.masterNodes?.length > 1 || !!cluster.masterVipIpv4)
```
jruddell, line 21: shouldn't the logic here require a VIP since you have to have a vip to scale?

jruddell (approved):
one comment about logic

src/app/plugins/kubernetes/components/infrastructure/clusters/helpers.ts
knnguy: I believe you need a VIP to scale only if it's a single master cluster.

PR #905, opened 2020-11-12
UX-637/Ipv6

jruddell:
image


src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/validators.tsx
```diff
@@ -1,23 +1,83 @@
...
+  const servicesIPInstance = new IPValidatorCls(value)
+
+  return (
+    servicesIPInstance.isInSubnet(containersIPInstance) === false &&
+    containersIPInstance.isInSubnet(servicesIPInstance) === false
+  )
+}, 'Services CIDR must be an independent subnet of Containers CIDR')
+
+const ipv6BlockSizeValidator = customValidator((value, formValues) => {
+  const blockSize = parseInt(value)
```
kevinylee1991, line 41: is this calling parseInt on the IPv6 string?

kevinylee1991 (approved):
I don't know IPv6 stuff well but aside from understanding how it works the react part looks good to me.

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/validators.tsx
jruddell: No, this is the block size validator for when calico is selected on the following input 
image

PR #910, opened 2020-11-14
UX-144, UX-160, UX-696, UX-716/Load user details when authenticating

jruddell:
image

image

image


src/api-client/Keystone.ts
```diff
@@ -364,18 +364,21 @@ class Keystone extends ApiService {
...
-      }
       this.client.activeProjectId = projectId
       this.client.scopedToken = scopedToken
       await this.getServiceCatalog()
 
-      return { user, role, scopedToken }
+      const user = await this.getUser(_user.id)
+      return {
+        user: {
+          ...user,
```
kevinylee1991, line 374: Unfortunately the clarity UI uses the username property in a lot of places. Could we keep user.username in here?

jruddell: It is there, when loading user details it returns the username

kevinylee1991 (approved):

PR #911, opened 2020-11-16
UX-695/Modified Navbar: Changed Pods, Deployments, Services to Workloads and…

knnguy:
… removed Namespaces

Changes Made:

- "Pods, Deployments, Services" in the NavBar has been changed to "Workloads"
- "Namespaces" has been removed from the Navbar and is now under "Workloads"
- Updated the URL for pods, services, deployments, and namespaces (Ex. /ui/kubernetes/workloads#pods)

Screen Shot 2020-11-16 at 2 22 56 PM

Screen Shot 2020-11-16 at 2 24 10 PM


src/app/plugins/kubernetes/components/namespaces/AddNamespacePage.js
```diff
@@ -38,7 +38,7 @@ export const AddNamespaceForm = ({ onComplete }) => {
 export const options = {
   cacheKey: ActionDataKeys.Namespaces, // TODO I think this is namespaces
   FormComponent: AddNamespaceForm,
-  listUrl: '/ui/kubernetes/namespaces',
+  listUrl: '/ui/kubernetes/pods#namespaces',
```
jruddell, line 41: two things, we should be using the routes we've defined (routes.ts) here so its easier to refactor. Also lets update the route to reflect the name `workloads`.

for example `/ui/kubernetes/workloads#pods`

knnguy: Screen Shot 2020-11-17 at 11 17 32 AM
Screen Shot 2020-11-17 at 11 18 03 AM

src/app/core/utils/routes.ts
```diff
@@ -332,15 +332,15 @@ export const routes = {
   },
   pods: {
     list: Route.register({
-      url: `${pluginRoutePrefix}/pods#pods`,
+      url: `${pluginRoutePrefix}/workloads/#pods`,
```
jruddell, line 335: why do some of these routes have a leading `/` and some dont?

for instance
`${pluginRoutePrefix}/workloads/#pods` and `${pluginRoutePrefix}/workloads#services`

before the `#` there is a leading forward slash.

jruddell (approved):

PR #912, opened 2020-11-17
UX-717/Fixed AWS regions not showing in update cloud provider page

knnguy:
What caused the bug:

- The cloud provider type wasn't passed to the region picklist component, so it defaulted to show the DisplayName, which is not available for AWS regions

Changes Made:

- Reused some cluster form components in the update cloud provider page
- Cloud provider type is now passed into the region picklist

Screen Shot 2020-11-17 at 12 44 41 PM

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/cloudProviders/UpdateCloudProviderPage.tsx
+++ b/src/app/plugins/kubernetes/components/infrastructure/cloudProviders/UpdateCloudProviderPage.tsx
@@ -19,6 +19,7 @@ import { CloudProviders, ICloudProvidersSelector } from './model'
 import DocumentMeta from 'core/components/DocumentMeta'
 import WizardMeta from 'core/components/wizard/WizardMeta'
 import { pick } from 'ramda'
+import { routes } from 'core/utils/routes'
 const objSwitchCaseAny: any = objSwitchCase // types on forward ref .js file dont work well.
 
 const useStyles = makeStyles((theme: Theme) => ({
@@ -148,7 +149,7 @@ export const options = {
   FormComponent: UpdateCloudProviderForm,
   updateFn: cloudProviderActions.update,
   loaderFn: cloudProviderActions.list,
-  listUrl: '/ui/kubernetes/infrastructure#cloudProviders',
+  listUrl: routes.cloudProviders.list,
   name: 'UpdateCloudProvider',
   title: 'Update Cloud Provider',
   uniqueIdentifier: 'uuid',
--- a/src/app/plugins/kubernetes/components/infrastructure/cloudProviders/UpdateCloudProviderPage.tsx
+++ b/src/app/plugins/kubernetes/components/infrastructure/cloudProviders/UpdateCloudProviderPage.tsx
@@ -149,7 +149,7 @@ export const options = {
   FormComponent: UpdateCloudProviderForm,
   updateFn: cloudProviderActions.update,
   loaderFn: cloudProviderActions.list,
-  listUrl: routes.cloudProviders.list,
+  listUrl: routes.cloudProviders.list.path(),
   name: 'UpdateCloudProvider',
   title: 'Update Cloud Provider',
   uniqueIdentifier: 'uuid',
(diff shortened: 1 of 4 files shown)


src/app/plugins/kubernetes/components/infrastructure/cloudProviders/UpdateCloudProviderPage.tsx
```diff
@@ -148,7 +149,7 @@ export const options = {
   FormComponent: UpdateCloudProviderForm,
   updateFn: cloudProviderActions.update,
   loaderFn: cloudProviderActions.list,
-  listUrl: '/ui/kubernetes/infrastructure#cloudProviders',
+  listUrl: routes.cloudProviders.list,
```
jruddell, line 152: did you test this? I don't think this will do what you expect. 

`routes.cloudProviders.list` is a route class object. I think instead you want to use the `path` method to expose this routes url. `routes.cloudProviders.list.path()`.

knnguy: Yep, my mistake. I forgot to call the path method

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/ssh-key-picklist.tsx
```diff
@@ -1,16 +1,19 @@
...
     <PicklistField
       DropdownComponent={dropdownComponent}
-      disabled={!(values.cloudProviderId && values.region)}
+      disabled={!(values.cloudProviderId && (values.region || values.cloudProviderRegionId))}
       id="sshKey"
       label="SSH Key"
       cloudProviderId={values.cloudProviderId}
-      cloudProviderRegionId={values.region}
-      info="Select an AWS SSH key to be associated with the nodes. This key can be used to access the nodes for debugging or other purposes."
+      cloudProviderRegionId={values.region ? values.region : values.cloudProviderRegionId}
```
jruddell, line 15: is the reason why we have different keys here (`region` and `cloudProviderRegionId`) because the field is used in more than one place? trying to understand why we store the value under both of these keys

knnguy: So when I created the `cloudProviderRegionField` for the AWS and Azure cluster creation forms, the field id is `regions`.  The field id used to be `cloudProviderId`, but I changed it to `regions` because I didn't see why we had to name it differently like that. The API accepts `region`, not `cloudProviderId`. 

So when I used this sshKeyField in the update cloud provider page, I saw that Kevin used `cloudProviderId` in `wizardContext` so that's why I added in `(values.region || values.cloudProviderRegionId)`.

When we use the `loadCloudProviderRegionDetails`, it accepts `cloudProviderRegionId` as a param. That's why in wizardContext, Kevin used `cloudProviderRegionId`. 

I guess I could change it to regions in the wizardContext and take out the `(values.region || values.cloudProviderRegionId)` in `sshKeyField`

knnguy: So when I created the `cloudProviderRegionField` for the AWS and Azure cluster creation forms, the field id is `regions`.  The field id used to be `cloudProviderId`, but I changed it to `regions` because I didn't see why we had to name it differently like that. The API accepts `region`, not `cloudProviderId`. 

So when I used this sshKeyField in the update cloud provider page, I saw that Kevin used `cloudProviderId` in `wizardContext` so that's why I added in `(values.region || values.cloudProviderRegionId)`.

When we use the `loadCloudProviderRegionDetails`, it accepts `cloudProviderRegionId` as a param. That's why in wizardContext, Kevin used `cloudProviderRegionId`. 

I guess I could change it to regions in the wizardContext and take out the `(values.region || values.cloudProviderRegionId)` in `sshKeyField`

jruddell: right. If you are refactoring the code to change something, we should be consistent and change it everywhere. The way I see it right now your change to use `region` was a partial refactor. We shouldn't be writing code to handle different keys like this because it is messy and your code will become unmanageable. Lets refactor the rest to use region so we can be explicit here.

jruddell (approved):

PR #914, opened 2020-11-18
UX-431/Added Node Authorization

knnguy:
### **Added an Authorization option in the toolbar and added a State filter dropdown**

Screen Shot 2020-11-18 at 7 17 02 PM

### **Node state filter dropdown options**

Screen Shot 2020-11-18 at 7 18 28 PM

### **New Advanced tab in the Onboard New Node page**

Screen Shot 2020-11-30 at 5 19 54 PM

### **Collapsible Info Cards**

Screen Shot 2020-11-20 at 7 33 12 PM

Screen Shot 2020-11-18 at 7 19 33 PM

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/nodes/NodesListPage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/nodes/NodesListPage.js
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react'
+import React from 'react'
 import { localizeRoles } from 'api-client/ResMgr'
 import { pathStrOr } from 'utils/fp'
 import ExternalLink from 'core/components/ExternalLink'
@@ -28,9 +28,9 @@ import { routes } from 'core/utils/routes'
 import { ToolbarActionIcon } from 'core/components/listTable/ListTableBatchActions'
 import { ActionDataKeys } from 'k8s/DataKeys'
 import ResourceUsageTables from '../common/ResourceUsageTables'
-import ApiClient from 'api-client/'
-import { loadResMgrHosts } from '../common/actions'
 import NodesStatePicklist from './nodes-state-picklist'
+import NodeAuthDialog from './NodeAuthDialog'
+import { NodeState } from './model'
 
 const useStyles = makeStyles((theme) => ({
   title: {
@@ -54,25 +54,14 @@ const usePrefParams = createUsePrefParamsHook('Nodes', listTablePrefs)
 const ListPage = ({ ListContainer }) => {
   return () => {
     const { params, getParamsUpdater } = usePrefParams(defaultParams)
-    const [data, loading, reload] = useDataLoader(loadNodes, params)
-    const [hosts] = useDataLoader(loadResMgrHosts)
-
-    const isUnauthorizedNode = useCallback(
-      (node) => {
-        const host = hosts.find((host) => host.id === node.uuid)
-        return (
-          host.roles.length === 0 || (host.roles.length === 1 && host.roles.includes('pf9-support'))
-        )
-      },
-      [hosts],
-    )
+    const [nodes, loading, reload] = useDataLoader(loadNodes, params)
 
     const filterByCluster = (node) => {
       if (params.clusterId === allKey) {
         return true
       } else if (params.clusterId === '') {
         return !node.clusterUuid
-      } else if (params.clusterId) {
+      } else {
         return node.clusterUuid === params.clusterId
       }
     }
@@ -80,15 +69,14 @@ const ListPage = ({ ListContainer }) => {
     const filterByState = (node) => {
       if (params.state === allKey) {
         return true
-      } else if (params.state === 'authorized') {
-        return !isUnauthorizedNode(node)
-      } else if (params.state === 'unauthorized') {
-        return isUnauthorizedNode(node)
+      } else if (params.state === NodeState.Authorized) {
+        return node.isAuthorized
+      } else {
+        return !node.isAuthorized
       }
     }
 
-    // Filter nodes based on cluster
-    const filteredNodes = data.filter((node) => {
+    const filteredNodes = nodes.filter((node) => {
       return filterByCluster(node) && filterByState(node)
     })
 
@@ -107,7 +95,7 @@ const ListPage = ({ ListContainer }) => {
               selectFirst={false}
               showNone
             />
-            <NodesStatePicklist onChange={getParamsUpdater('state')} selectedFirst={false} />
+            <NodesStatePicklist onChange={getParamsUpdater('state')} value={params.state} />
           </>
         }
         {...pick(listTablePrefs, params)}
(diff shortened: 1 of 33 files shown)


src/app/plugins/kubernetes/components/infrastructure/nodes/NodesListPage.js
```diff
@@ -44,22 +47,49 @@ const useStyles = makeStyles((theme) => ({
...
     const [data, loading, reload] = useDataLoader(loadNodes, params)
-    // Filter nodes based on cluster
-    const filteredNodes = data.filter((node) => {
+    const [hosts] = useDataLoader(loadResMgrHosts)
+
+    const isUnauthorizedNode = useCallback(
+      (node) => {
+        const host = hosts.find((host) => host.id === node.uuid)
+        return (
+          host.roles.length === 0 || (host.roles.length === 1 && host.roles.includes('pf9-support'))
```
jruddell, line 64: Can you explain the logic here? Did backend give a requirement on the length of the roles a host has?

Also lets convert this to typescript and use optional chaining here (`path?.to?.something`). since this is data loaded from the API it could potentially change in the future and also `Array.find` will return undefined if there are no matches. better to be error proof :)

src/app/plugins/kubernetes/components/infrastructure/nodes/nodes-state-picklist.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import PicklistDefault from 'core/components/Picklist'
+const Picklist: any = PicklistDefault // types on forward ref .js file dont work well.
+
+const options = [
+  { label: 'Authorized', value: 'authorized' },
```
jruddell, line 6: lets make the value use an enum here and reuse the same enum when checking the value of what is selected

src/app/plugins/kubernetes/components/infrastructure/nodes/actions.ts
```diff
@@ -6,19 +6,33 @@ import createContextUpdater from 'core/helpers/createContextUpdater'
...
-    const [rawNodes] = await Promise.all([
+    const [rawNodes, , hosts] = await Promise.all([
       qbert.getNodes(),
       // Fetch dependent caches
       loadServiceCatalog(),
       loadResMgrHosts(),
     ])
-    return rawNodes
+
+    const unauthorizedHosts: Resmgr[] = hosts.filter(
```
jruddell, line 24: you dont need two variables here, just chain a filter and then a map. `hosts.filter(...).map(...)`

src/app/plugins/kubernetes/components/infrastructure/nodes/selectors.ts
```diff
@@ -30,6 +30,7 @@ export const nodesSelector = createSelector(
       const usage = calculateNodeUsages([combinedHostsObj[node.uuid]]) // expects array
       return {
         ...node,
+        isAuthorized: node.isAuthorized === undefined ? true : false,
```
jruddell, line 33: Is there not a piece of data from the API that tells us if it is authorized? assigning true based on it not existing feels error prone

knnguy: There's nothing in a node that would tell us if it's authorized or not. All nodes from the nodes API are authorized. However, on hosts, you can tell if it's authorized if there is a role in the `roles` field. 

In nodes/actions.ts, I got the unauthorized `hosts` and made them into type `Nodes` so that they can be dispatched along with the `rawNodes` to the store. I added the `isAuthorized` field so we can tell in the selector if it's authorized or not.

Screen Shot 2020-11-30 at 10 58 18 AM

Do you suggest a different a way to do this?

src/app/plugins/kubernetes/components/infrastructure/nodes/actions.ts
```diff
@@ -6,19 +6,32 @@ import createContextUpdater from 'core/helpers/createContextUpdater'
...
+    const [rawNodes, , hosts] = await Promise.all([
       qbert.getNodes(),
       // Fetch dependent caches
       loadServiceCatalog(),
       loadResMgrHosts(),
     ])
-    return rawNodes
+
+    const unauthorizedNodes: Node[] = hosts
+      .filter((host: Resmgr) => rawNodes.find((node) => node.uuid === host.id) === undefined)
```
kevinylee1991, line 25: I think it may be better to check for resmgr hosts that do not have the pf9-kube role, I believe resmgr may be more of the 'source of truth' for the nodes. There may also be nodes that do not have the pf9-kube role but have other roles (like hypervisor role in OpenStack) that are not technically unauthorized, they just are not authorized for kubernetes purposes.

src/app/plugins/kubernetes/components/infrastructure/nodes/selectors.ts
kevinylee1991: Perhaps mapping rawNodes to add isAuthorized true in loadNodes would be better to consolidate the assignment in one spot. Or assign isAuthorized for all the nodes in the selector instead.

src/app/plugins/openstack/components/hosts/actions.tsx
```diff
@@ -2,6 +2,6 @@ import ApiClient from 'api-client/ApiClient'
 
 const { keystone } = ApiClient.getInstance()
 
-export const getDownloadLinks = () => (
-  keystone.getDownloadLinks()
-)
+export const getDownloadLinks = async () => {
```
kevinylee1991, line 5: could revert to the original (or if prettier changed this I guess we can keep it this way so it doesn't change again later)

kevinylee1991 (commented):
Few comments but I think it looks good. Once the isAuthorized thing is resolved I can go ahead and approve

src/app/plugins/kubernetes/components/infrastructure/nodes/selectors.ts
jruddell: > you can tell if it's authorized if there is a role in the roles field.

can you show me where this logic is happening? this is what I was referring to as what I was expecting to see. right now when you map over the hosts and filter it by a raw node you set the `isAuthorized` flag to false. isn't that where you should be comparing the roles field to determine if its authorized or not?

knnguy: Yep, I have changed the logic for this. I now determine if a host is unauthorized or not based on its roles, not whether or not it matches up with a node from Qbert.

Screen Shot 2020-12-01 at 10 22 19 AM

src/app/plugins/kubernetes/components/infrastructure/nodes/actions.ts
```diff
@@ -11,6 +11,8 @@ import { Node } from 'api-client/qbert.model'
 
 const { qbert, resMgr } = ApiClient.getInstance()
 
+const isUnauthorizedHost = (host) => !host.roles.includes('pf9-kube')
```
jruddell, line 14: move this to a helpers or utility file please

jruddell (approved):

PR #916, opened 2020-12-01
UX-599/Added 'View Container Logs' option to Pods List Table

knnguy:
- Added a links column to the pods list table

- Added 'View Container Logs' link to the links column

Screen Shot 2020-12-01 at 10 52 06 AM

When "View Container Logs' is clicked, a new window will open showing the logs

Screen Shot 2020-11-30 at 4 29 05 PM

### DIFF
--- a/src/app/plugins/kubernetes/components/pods/PodsListPage.js
+++ b/src/app/plugins/kubernetes/components/pods/PodsListPage.js
@@ -82,12 +82,7 @@ const PodLinks = ({ pod }) => {
     <div className={classes.links}>
       <ExternalLink url={pod.dashboardUrl}>dashboard</ExternalLink>
       {pod.logs && (
-        <SimpleLink
-          src={pod.logs || ''}
-          target="_blank"
-          rel="noopener"
-          onClick={openLogsWindow(pod)}
-        >
+        <SimpleLink src={pod.logs} target="_blank" rel="noopener" onClick={openLogsWindow(pod)}>
           View Container Logs
         </SimpleLink>
       )}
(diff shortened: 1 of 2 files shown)


src/app/plugins/kubernetes/components/pods/PodsListPage.js
```diff
@@ -99,6 +122,13 @@ export const options = {
       render: (value) => <DateAndTime value={value} />,
     },
   ],
+  batchActions: [
+    {
```
jruddell, line 126: lets not make this an action, the link in the table row is sufficient. Actions should be more reserved for mutation type actions (edit, delete... etc).

src/app/plugins/kubernetes/components/pods/selectors.ts
```diff
@@ -37,6 +44,10 @@ export const podsSelector = createSelector(
           namespace: pod?.metadata?.namespace, // pathStr('metadata.namespace', pod),
           labels: pod?.metadata?.labels, // pathStr('metadata.labels', pod),
           clusterName: cluster?.name,
+          logs: `${qbertUrl}/clusters/${cluster?.uuid}/k8sapi/api/v1/namespaces/${pod?.metadata?.namespace}/pods/${pod?.metadata?.name}/log`.replace(
```
jruddell, line 47: make a helper function to generate this url that you call here.

src/app/plugins/kubernetes/components/pods/selectors.ts
```diff
@@ -13,9 +13,22 @@ import { FluffySelector, MatchLabelsClass } from 'api-client/qbert.model'
 
 const k8sDocUrl = 'namespaces/kube-system/qbertservices/https:kubernetes-dashboard:443/proxy/#'
 
+const getLogsUrl = (pod, cluster, rawServiceCatalog) => {
+  const qbertUrl =
```
jruddell, line 17: what happens when we cannot create this url? when the qbert url or the pod isn't there?

jruddell (approved):

PR #917, opened 2020-12-02
UX-636/Made username and password field for KubeConfig download form bigger

knnguy:
- The username and password input field is now bigger and fills up the rest of the form

- When a user hits enter, the download kubeconfig button will get triggered. (This was already implemented)

Screen Shot 2020-12-01 at 6 07 11 PM


jruddell (approved):
make sure to link the jira ticket in the title

PR #924, opened 2020-12-03
UX-665/Added SSO Groups, started on SSO config page

kevinylee1991:
Screen Shot 2020-12-03 at 2 49 31 PM
Screen Shot 2020-12-03 at 2 49 52 PM
Screen Shot 2020-12-03 at 2 50 06 PM
Screen Shot 2020-12-03 at 2 50 56 PM
Screen Shot 2020-12-03 at 2 51 43 PM


src/app/core/elements/input.tsx
```diff
@@ -87,7 +87,8 @@ const useStyles = makeStyles<Theme, { variant: string }>((theme: Theme) => ({
     },
     '& input': {
       ...theme.typography.inputPlaceholder,
-      minHeight: 54,
+      fontSize: 14,
```
jruddell, line 90: the fontsize should be defined inside `inputPlaceholder`

src/app/core/components/wizard/Wizard.js
```diff
@@ -204,6 +214,8 @@ Wizard.propTypes = {
   startingStep: PropTypes.number,
   hideBack: PropTypes.bool,
   hideAllButtons: PropTypes.bool,
+  showButtonsIfProp: PropTypes.string,
```
jruddell, line 217: this seems pretty specific of a use case to make a new prop on the wizard. what are you trying to do here? can we find a better way to address this?

src/app/plugins/account/components/ssoManagement/groups/EditGroupPage.tsx
```diff
@@ -0,0 +1,238 @@
...
+    !roleAssignments
+
+  const finishedInitialLoad = !(
+    loadingGroups ||
+    loadingGroupMappings ||
+    // loadingRoleAssignments ||
+    !roleAssignments
+  )
+
+  const submitForm = async (params) => {
```
jruddell, line 126: isn't this the same as the create page submit? maybe this logic should exist in an action. This is pretty specific / complicated flow of updates and mappings. lets move it to one place (an action) and call it in these forms

jruddell: Remember to think about your architecture and how to make it scale. Hardcoding and rewriting things like this make your life harder in the long run

src/app/plugins/account/components/ssoManagement/sso/UploadSamlMetadataLink.tsx
```diff
@@ -0,0 +1,101 @@
...
+    reader.onabort = () => console.log('file reading was aborted')
+    reader.onerror = () => console.log('file reading has failed')
+    reader.onload = () => {
+      const text = reader.result
+      setWizardContext({ metadata: text, metadataFileName: file.name })
+      handleClose()
+    }
+    reader.readAsText(file)
+  }, [])
+  const { getRootProps, getInputProps } = useDropzone({ onDrop, accept: '.xml', maxFiles: 1 })
```
jruddell, line 68: We should probably make a drag and drop component that we can use in other places in the future

src/app/plugins/account/components/userManagement/groups/UserGroupsListPage.js
```diff
@@ -1,20 +1,20 @@
-import createCRUDComponents from 'core/helpers/createCRUDComponents'
-import { ActionDataKeys } from 'k8s/DataKeys'
+// import createCRUDComponents from 'core/helpers/createCRUDComponents'
```
jruddell, line 1: should this file just be deleted?

src/app/core/components/wizard/Wizard.js
kevinylee1991: Regarding that showButtonsIfProp thing, it was for showing/hiding the wizard buttons based on whether a wizardContext property was true/false. It's not currently being used anywhere since SSO is hidden so maybe it can just be removed right now.

The best solution I can think of is perhaps to take a function that has wizardContext passed in that should evaluate true/false for showing the wizard buttons. Then the hideAllButtons prop on the wizard could be deprecated too

src/app/plugins/account/components/ssoManagement/groups/EditGroupPage.tsx
kevinylee1991: it's really close... yeah I should be able to merge the similar parts and set a flag to distinguish to 1 different section

jruddell (approved):

src/app/plugins/account/components/ssoManagement/groups/GroupsListPage.tsx
```diff
@@ -9,12 +11,16 @@ export const options = {
     { id: 'samlAttributesString', label: 'SAML Attributes Mapped' },
   ],
   cacheKey: ActionDataKeys.ManagementGroups,
-  // editUrl: '/ui/kubernetes/infrastructure/groups/edit',
+  customEditUrlFn: (_, id) => routes.sso.editGroup.path({ id }),
```
jruddell, line 14: change this to just use `editUrl` once you have merged latest master in with Xan's change

src/app/plugins/account/components/userManagement/users/actions.js
```diff
@@ -153,5 +153,6 @@ export const mngmUserRoleAssignmentsLoader = createContextLoader(
   {
     uniqueIdentifier: ['user.id', 'role.id'],
     indexBy: 'userId',
+    cache: false,
```
jruddell, line 156: remove

jruddell (approved):

PR #925, opened 2020-12-04
Fixed cluster review table bug where checkbox field values were alway…

knnguy:
…s showing up as enabled (true)


src/app/plugins/kubernetes/components/infrastructure/clusters/aws/create-templates/custom.tsx
```diff
@@ -64,6 +64,8 @@ export const initialContext = {
   tags: [],
   appCatalogEnabled: false,
   networkStack: 'ipv4',
+  privileged: false,
```
jruddell, line 67: lets set privileged to true as the default value

jruddell (approved):

PR #931, opened 2020-12-08
UX-676 & UX-745/Added groups as option for rbac role bindings and fixed rbac bugs from refactor

kevinylee1991:
Screen Shot 2020-12-07 at 7 44 05 PM
Screen Shot 2020-12-07 at 7 44 37 PM
Screen Shot 2020-12-07 at 7 44 51 PM


src/app/plugins/account/components/userManagement/groups/actions.js
```diff
@@ -10,7 +10,7 @@ export const mngmGroupActions = createCRUDActions(ActionDataKeys.ManagementGroup
     const [groups] = await Promise.all([
       keystone.getGroups(),
       // Fetch dependent caches
-      mngmGroupMappingActions.list(),
+      mngmGroupMappingActions,
```
jruddell, line 13: is this an action? if so rename it to be singular.

xantorres: We should rename this to `mngmGroupMappingLoader` and call it with `()`, otherwise this won't do anything

src/app/core/components/validatedForm/withFormContext.js
```diff
@@ -49,7 +49,9 @@ const ValidatedFormInput = ({
   const getCurrentFieldValue = getFieldValue(id)
   const validateCurrentField = validateField(id)
   const currentInitialValue = initialValue !== undefined ? initialValue : initialValues[id]
-  const currentValue = isNil(value) ? values[id] : value
+  const currentValue = value
+  // Is below line needed? This causes desync between useParams and the picklist components
+  // const currentValue = isNil(value) ? values[id] : value
```
xantorres, line 54: What this does is to try to pick the value from the ValidatedForm values if the value is not provided by the controlled component.
If you want to override the value provided by the ValidatedForm just initialise the input value to an empty string or 0. 
But it all depends on how you want to be handling the Picklist value, either from the picklist itself or by the `ValidatedForm`.

In any way please don't change this as it is being used in all the forms so it might cause unexpected issues.

src/app/plugins/kubernetes/components/common/ClusterPicklist.js
```diff
@@ -33,7 +34,7 @@ const ClusterPicklist = forwardRef(
 
     // Select the first cluster as soon as clusters are loaded
     useEffect(() => {
-      if (!isEmpty(options) && selectFirst) {
+      if (!isEmpty(options) && selectFirst && !value) {
```
xantorres, line 37: I don't think this is necessary

xantorres (changes requested):

src/app/core/components/validatedForm/withFormContext.js
kevinylee1991: thanks I'll give this a try

kevinylee1991: Cool I didn't know about adding the property in the useParams initialization. That did the trick for the desync issue I was having.

src/app/plugins/kubernetes/components/common/ClusterPicklist.js
kevinylee1991: For the rbac wizards, when selecting another cluster, the clusters cache gets updated and ends up causing the clusters list to repopulate. Causes this useEffect to run again. Is there any negative effect to including the !value?

xantorres: I understand, in that case I guess it makes sense, lets keep it 
Maybe we should check other similar `Picklists` as they are following the same approach and we could potentially run into the same issue in the future.

jruddell (approved):

PR #932, opened 2020-12-08
Fixed Segment and Drift bug during login

knnguy:
A bug occurs when you login, logout, then login again. When `createSegmentScript()` and `createDriftScript() ` runs for the 2nd time, it returns `undefined` since the segment script already exists. When you try to add `undefined` to the DOM body, an exception occurs.

### DIFF
--- a/src/app/core/containers/AppContainer.tsx
+++ b/src/app/core/containers/AppContainer.tsx
@@ -108,18 +108,12 @@ const getUserDetails = async (activeTenant, isSsoToken) => {
 
   // Segment tracking
   if (!analyticsOff) {
-    const segmentScript = createSegmentScript()
-    if (segmentScript) {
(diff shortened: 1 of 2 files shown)


src/app/core/containers/AppContainer.tsx
```diff
@@ -108,13 +108,20 @@ const getUserDetails = async (activeTenant, isSsoToken) => {
 
   // Segment tracking
   if (!analyticsOff) {
-    DocumentMeta.addElementToDomBody(createSegmentScript())
+    const segmentScript = createSegmentScript()
+    if (segmentScript) {
+      DocumentMeta.addElementToDomBody(segmentScript)
```
jruddell, line 113: you should put this check inside DocumentMeta, this way you dont have to constantly check before calling

jruddell (approved):

PR #935, opened 2020-12-09
UX-680/Multiversion support on both create and upgrade

jruddell:
Only one upgrade option
image

multiple upgrade options
image

cluster create
image


src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterUpgradeDialog.tsx
```diff
@@ -0,0 +1,89 @@
...
+import Text from 'core/elements/text'
+import { FormControlLabel, Radio, RadioGroup } from '@material-ui/core'
+
+enum UpgradeTypes {
+  Minor = 'minor',
+  Patch = 'patch',
+}
+
+function getUpgradeTarget(cluster) {
+  if ((cluster.canMinorUpgrade && cluster.canPatchUpgrade) || cluster.canMinorUpgrade) {
```
kevinylee1991, line 14: The conditions seems redundant, seems the first one can be removed?

knnguy, line 14: Isn't the `cluster.canPatchUpgrade ` here redundant?

kevinylee1991 (approved):

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterUpgradeDialog.tsx
jruddell: so in both cases of 2 possible upgrades or if there is only one possible upgrade of type `minor` i want to default to a minor upgrade.

knnguy: So you would only need `if (cluster.canMinorUpgrade)`. That's what I was trying to say

jruddell: sorry misread what you were saying. I was thinking it was a bit more explicit but, it probably reads better to do it that way

src/api-client/ApiClient.ts
```diff
@@ -175,88 +175,93 @@ class ApiClient {
...
     url,
+    version,
     endpoint = undefined,
     config = undefined,
     options: { clsName, mthdName },
   }: IRawRequestGetParams) => {
-    if (endpoint === undefined) {
-      endpoint = await this.apiServices[clsName].getApiEndpoint()
-    }
+    endpoint = await this.getEndpoint({ endpoint, version, clsName })
```
xantorres, line 195: A decorator would have done the trick in a more elegant way

xantorres (approved):

PR #948, opened 2020-12-18
UX-668/Created user profile page

knnguy:


knnguy:
I noticed during testing that sometimes when I change my own password multiple times in a row, I get this 404 error that says the token is invalid.

So after the password is updated, I call `keystone.authenticate` to get the` unscopedToken`. Then, I call `keystone.changeProjectScope` with the 'unscopedToken' to get the `scopedToken`. Sometimes for this call, I would get the 404 error. I checked the networks tab and the `unscopedToken` that is being sent is exactly the same as the one we received earlier for the authentication.

Screen Shot 2020-12-17 at 6 05 00 PM
Screen Shot 2020-12-17 at 6 06 02 PM

knnguy:
Another problem that I've noticed is that when the user changes their username, a new field under `preferences` in the redux store is created with the new username. 

As you can see in the screenshot below, my I changed my username from `kntest@kim.commmm` to `kntest` and a new field for `kntest` was created under `preferences` in the redux store. The current region defaulted to `ironic`. I can't seem to find where this user preferences gets set when a user update is triggered.

Screen Shot 2020-12-17 at 9 24 09 PM

It doesn't look like the `updatePrefs` from `preferencesActions` allow us to modify the username for a preference. Not sure if I should try to copy the old user preferences over to the new username.

Now I believe this worked in the Edit User page in master because username under `session` was not updated to reflect the new username. Only the username under `userDetails` in the redux store was updated

src/app/plugins/account/components/userManagement/users/actions.js
```diff
@@ -83,22 +128,36 @@ export const mngmUserActions = createCRUDActions(ActionDataKeys.ManagementUsers,
...
     // Perform the api calls to update the user and the tenant/role assignments
-    const updatedUserPromise = keystone.updateUser(userId, {
+    const updatedUser = await keystone.updateUser(userId, {
       name: username,
       email: username,
       displayname,
       password: password || undefined,
     })
+
+    // Update session if updating active user
```
xantorres, line 150: I don't think this is the right place to put this kind of logic. Api related actions such as `updateFn` should just be responsible of updating the cache and calling the right api endpoints to update the data in the server, which in this case is complex enough already.

If we want to run other side effects they should be called from the view. Moreover, from there you will have the advantage of being able to use handy hooks like `useScopedPreferences`. Maybe a good solution would be to create a separate function and call it from view, although the problem is that hooks cannot be used, but you could fetch the required data in the view at pass it to the `authenticateUser` function.

src/app/plugins/account/components/userManagement/users/actions.js
```diff
@@ -118,20 +177,17 @@ export const mngmUserActions = createCRUDActions(ActionDataKeys.ManagementUsers,
...
+        console.warn(err.message)
+        return emptyArr
+      },
+    )(null)
+
     mngmTenantActions.invalidateCache()
     // Refresh the user/roles cache
-    await mngmUserRoleAssignmentsLoader({ userId }, true)
-
+    const currentRoleAssignments = await mngmUserRoleAssignmentsLoader({ userId }, true)
```
xantorres, line 190: This variable is not used anywhere, we can just remove it.

xantorres:
> Another problem that I've noticed is that when the user changes their username, a new field under `preferences` in the redux store is created with the new username so all their old preferences are gone.
> 
> As you can see in the screenshot below, my I changed my username from `kntest@kim.commmm` to `kntest` and a new field for `kntest` was created under `preferences` in the redux store. The current region defaulted to `ironic`. I can't seem to find where this user preferences gets set when a user update is triggered.
> 
> Screen Shot 2020-12-17 at 9 24 09 PM
> 
> It doesn't look like the `updatePrefs` from `preferencesActions` allow us to modify the username for a preference. Not sure if I should try to copy the old user preferences over to the new username.

If you want to do that you should create a new specific reducer in `preferencesReducers`

src/app/plugins/account/components/userManagement/users/actions.js
jruddell: I would prefer this is not in the view but rather an action. Multiple places can call an action potentially, maybe we should have some form of a workflow action function that we call so it can be common and not specifically inside the update function? This would be useful for some of the specific things we do in other views (like AppContainer for setting up the session). @xantorres thoughts?

knnguy: You're saying I shoudn't call `updateSession` and `authenticateUser` in `updateFn` right? 

The problem is that right after we call `keystone.updateUser`, we need to call `authenticateUser` right after or else the subsequent calls to the api will fail with a 401 error. I was told this is because the token gets invalidated after you update the user's password .

knnguy:
Regarding the user preferences problem in the previous comment, my solution is to create

> > Another problem that I've noticed is that when the user changes their username, a new field under `preferences` in the redux store is created with the new username so all their old preferences are gone.
> > As you can see in the screenshot below, my I changed my username from `kntest@kim.commmm` to `kntest` and a new field for `kntest` was created under `preferences` in the redux store. The current region defaulted to `ironic`. I can't seem to find where this user preferences gets set when a user update is triggered.
> > Screen Shot 2020-12-17 at 9 24 09 PM
> > It doesn't look like the `updatePrefs` from `preferencesActions` allow us to modify the username for a preference. Not sure if I should try to copy the old user preferences over to the new username.
> 
> If you want to do that you should create a new specific reducer in `preferencesReducers`, and return a new state slice from there.

So my solution is to transfer the old user's preferences to the new user. I created a new reducer in `preferencesReducers `that basically replaces the entire preferences for the user.

src/app/plugins/account/components/userManagement/users/actions.js
xantorres: Sounds good to me, but I would not call it "action" as it could lead to confusion. That's why I said about using a regular function and calling that from the view.

jruddell: Right. I think it would be better to include this function call inside the action though. I don't want the view calling this. I'm thinking similar to how we already do cluster creation, the action calls a helper function to create the cluster. This is a good pattern because it keeps the logic around a change like this tied to the actual action of doing that thing.

src/app/core/session/preferencesReducers.ts
```diff
@@ -18,6 +18,10 @@ const {
     updatePrefs: (state, { payload }: PayloadAction<PreferencesState>) => {
       return over(lensPath([payload.username, payload.key]), mergeLeft(payload.prefs), state)
     },
+    updateAllPrefs: (state, { payload }: PayloadAction<PreferencesState>) => {
+      const { username, ...prefs } = payload
```
jruddell, line 22: what is this line doing? doesn't look like its being used

src/app/plugins/account/components/userManagement/users/actions.js
```diff
@@ -37,11 +37,20 @@ export const loadCredentials = createContextLoader(
...
+      userDetails: { ...activeUser, username, name: username, email: username, displayName },
+    }),
+  )
+}
+
+const updateUserPrefs = (newUsername, oldUserPrefs) => {
+  store.dispatch(
+    preferencesActions.updateAllPrefs({
+      username: newUsername,
+      ...oldUserPrefs,
```
jruddell, line 53: does the `oldUserPrefs` have the key `username` on it? if so wont that override the newUsername you're trying to set here? I think it would be better to do your spread operation first, then specify the keys you want to overwrite

src/app/plugins/account/components/userManagement/users/actions.js
xantorres: If you put this function inside the action you are giving too much responsibility to the action and therefore obfuscating the code and making things harder to follow. The logic we are following now is that the views are the "controllers" of the logic flow, and they give orders about what has to be done and how. The rest are just utilities with the least amount of side effects as possible that deal with just one action at a time.

jruddell: If you have to replicate the same kind of action from more than one view, you are rewriting the workflow of code. Actions are much more appropriate for logic like this.

PR #952, opened 2020-12-22
Switch SSH user for target DU from centos to tcbuilder

ParagDoke:


deploy.bash
```diff
@@ -29,8 +31,8 @@ pushd ${WORKSPACE}/pf9-ui-plugin/build
     for du in ${TARGET_DUS}; do
         echo "INFO: Starting to update DU ${du}"
         # TODO: Need a dedicated login to target DU that can modify files owned by root
-        scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no /tmp/ui.tgz centos@${du}:/tmp/ui.tgz
-        ssh -t -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no centos@${du} "cd /opt/pf9/www/public; sudo tar xvzpf /tmp/ui.tgz && \rm /tmp/ui.tgz"
+        scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no /tmp/ui.tgz ${SSH_USER}@${du}:/tmp/ui.tgz
+        ssh -t -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ${SSH_USER}@${du} "cd /opt/pf9/www/public; tar xvzpf /tmp/ui.tgz && \rm /tmp/ui.tgz"
```
deeptijoshi, line 35: can we keep sudo, in case ssh user is set to a different one?

deeptijoshi (approved):

deploy.bash
ParagDoke: Ansible post playbook sets owner to tcbuilder via https://github.com/platform9/pf9-post-playbooks/blob/atherton/ui-dev.platform9.horse/post-playbook.yml#L21. So we should be able to avoid sudo.

jruddell (approved):

PR #957, opened 2021-01-05
UX-772/Added hostagent download links for DDUs

knnguy:


src/app/core/components/DownloadHostAgentWalkthrough.tsx
```diff
@@ -21,11 +26,22 @@ enum OsDownloadLabel {
...
-enum osDownloadLinkName {
+enum OsDownloadLinkName {
   Linux = 'rpm_install',
   Ubuntu = 'deb_install',
 }
 
+const deccoDownloadOptions = [
+  {
+    label: OsDownloadLabel[OsOptions.Linux],
+    link: 'https://pmkft-1584219278-18238.platform9.io/clarity/platform9-install-redhat.sh',
```
jruddell, line 37: this is a hardcoded URL. Please think about this more before writing out a url like this. Its a product that will run on many different origins, this part should be dynamic

knnguy: Oops. Sorry, I misunderstood what you said about the links in Slack

src/app/core/components/DownloadHostAgentWalkthrough.tsx
```diff
@@ -85,23 +101,31 @@ const DownloadHostAgentWalkthrough = ({ osOptions }): JSX.Element => {
   const classes = useStyles({})
   const [anchorEl, setAnchorEl] = useState(null)
   const [downloadOptions, setDownloadOptions] = useState([])
+  const session = useSelector<RootState, SessionState>(prop(sessionStoreKey))
+  const { features } = session
+  const isDecco = pathStrOr(false, 'experimental.kplane', features)
```
jruddell, line 106: Lets make this (isDecco) a helper as its used in more than one file

knnguy: I put it in a new helper file in `app/core/utils/helpers.ts`. Not sure if this is the best place to put it.

jruddell (approved):
can you also update the other places where we check to see if the environment is a decco one to use this helper?

PR #961, opened 2021-01-08
UX-749, UX-781/Added missing routes for Segment and replaced string paths in list pages with routes

knnguy:


src/app/core/utils/routes.ts
```diff
@@ -344,82 +351,95 @@ export const routes = {
...
+    root: Route.register({
       url: `${pluginRoutePrefix}/storage_classes`,
-      name: 'Storage:List',
       defaultParams: {
         plugin: AppPlugins.Kubernetes,
       },
+      name: 'StorageClasses:Root',
+    }),
+    list: Route.register({
+      url: `${pluginRoutePrefix}/storage_classes##storage`,
```
jruddell, line 431: looks like theres an extra # in the url

src/app/core/utils/routes.ts
```diff
@@ -344,82 +351,95 @@ export const routes = {
...
       defaultParams: {
         plugin: AppPlugins.Kubernetes,
       },
+      name: 'Namespaces:Add',
     }),
   },
-  storage: {
-    list: Route.register({
+  storageClasses: {
+    root: Route.register({
```
jruddell, line 423: do we need a separate route definition for the page without the hash?

src/app/core/utils/routes.ts
```diff
@@ -344,82 +351,95 @@ export const routes = {
...
-      name: 'Namespaces:Add',
       defaultParams: {
         plugin: AppPlugins.Kubernetes,
       },
+      name: 'Namespaces:Add',
     }),
   },
-  storage: {
-    list: Route.register({
+  storageClasses: {
```
jruddell, line 422: is this not used anywhere? changing the name here would lead me to expect that there should be changes in other files to handle this name change.

knnguy: No, `routes.storage` is not used anywhere, but since the name change is not very necessary, I can change it back just to be safe.

jruddell: ok please make a ticket to go though and ensure were using these routes correctly, the list pages usually are defining urls, looks like storage wasn't updated to use this

src/app/core/utils/routes.ts
knnguy: According to the ticket, yes. Also, I just checked with Chris.

https://pmkft-1584219278-18238.platform9.io/ui/kubernetes/alarms

https://pmkft-1584219278-18238.platform9.io/ui/kubernetes/alarms#alarms

These two should have the same name, Monitoring:Alarms:List

knnguy: This might be error prone though because if a tab is no longer the first tab for a page, we need to remember to update the page name for the urls without the hash

jruddell (approved):

PR #962, opened 2021-01-13
UX-692/Allow users to see the statuses of the validation tests performed during step 1 of cloud provider creation

knnguy:
NOTE:
- The `POST` request to `/cloudProviders` already does the validations so no need for any extra API calls to validate
- `TestsDialog` is a reusable component that can show all the tests performed during cloud provider creation
-  The close button in the `TestsDialog` will only show up when there are no longer any tests that are loading (tests have either failed or completed successfully)
- Closing the `TestsDialog` box will send the user to the second step if all the tests passed
- The error message from the API response cannot be shown due to the current architecture of the UI code. In the code, there was a comment saying that Xan is/will be working on this.

Loading

Screen Shot 2021-01-14 at 11 42 21 AM

Screen Shot 2021-01-14 at 11 43 31 AM

### DIFF
--- a/src/app/core/themes/relianceLight.ts
+++ b/src/app/core/themes/relianceLight.ts
@@ -268,15 +268,6 @@ export const typography = {
     lineHeight: 'normal',
     letterSpacing: 'normal',
   },
-  body2Italicized: {
-    fontFamily: 'Eina04',
-    fontSize: '14px',
-    fontWeight: 'normal',
-    fontStretch: 'normal',
-    fontStyle: 'italic',
-    lineHeight: 'normal',
-    letterSpacing: 'normal',
-  },
   sidenav: {
     fontFamily: 'Eina04',
     fontSize: '13px',
--- a/src/app/core/themes/relianceLight.ts
+++ b/src/app/core/themes/relianceLight.ts
@@ -268,6 +268,15 @@ export const typography = {
     lineHeight: 'normal',
     letterSpacing: 'normal',
   },
+  body2Italicized: {
+    fontFamily: 'Eina04',
+    fontSize: '14px',
+    fontWeight: 'normal',
+    fontStretch: 'normal',
+    fontStyle: 'italic',
+    lineHeight: 'normal',
+    letterSpacing: 'normal',
+  },
   sidenav: {
     fontFamily: 'Eina04',
     fontSize: '13px',
--- a/src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AddCloudProviderCredentialStep.tsx
+++ b/src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AddCloudProviderCredentialStep.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useCallback, useMemo, useRef, useState } from 'react'
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { makeStyles } from '@material-ui/styles'
 import { Theme } from '@material-ui/core'
 import ValidatedForm from 'core/components/validatedForm/ValidatedForm'
@@ -10,7 +10,8 @@ import { awsPrerequisitesLink, azurePrerequisitesLink } from 'k8s/links'
 import { objSwitchCase } from 'utils/fp'
 import { cloudProviderActions } from './actions'
 import Text from 'core/elements/text'
-import { CloudProviders } from './model'
+import { CloudProviders, CloudProvidersFriendlyName } from './model'
+import TestsDialog, { TestStatus } from './tests-dialog'
 const objSwitchCaseAny: any = objSwitchCase // types on forward ref .js file dont work well.
 
 const useStyles = makeStyles((theme: Theme) => ({
@@ -31,6 +32,7 @@ interface Props {
   wizardContext: any
   setWizardContext: any
   onNext: any
+  handleNext: any
   title: string
   setSubmitting: any
 }
@@ -47,6 +49,11 @@ const links = {
     </ExternalLink>
   ),
 }
+
+const requiredTests = (provider) =>
+  new Map<string, TestStatus | null>([
+    [`${CloudProvidersFriendlyName[provider]} account access`, null],
+  ])
 
 const formCpBody = (wizardContext) => {
   if (wizardContext.provider === CloudProviders.Aws) {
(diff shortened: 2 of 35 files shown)


src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AddCloudProviderCredentialStep.tsx
```diff
@@ -73,47 +80,90 @@ const AddCloudProviderCredentialStep = ({
...
+
     setSubmitting(false)
-    return true
+    updateTestStatus(`${provider} account access`, TestStatus.Success)
+    setVerified(true)
+    return false
   }, [wizardContext])
 
   useEffect(() => {
+    // Reset the tests whenever cloud provider changes
```
kevinylee1991, line 144: Rather than resetting the tests, we could keep a dictionary of separate tests, like

{
  [CloudProviders.Aws]: testsForAws,
  [CloudProviders.Azure]: testsForAzure,
}

And then we can reference the appropriate tests based on the cloud provider type selected.

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AddCloudProviderCredentialStep.tsx
```diff
@@ -73,47 +80,90 @@ const AddCloudProviderCredentialStep = ({
...
+    // submitStep will always return false and prevent the user from
+    // proceeding to the next step because we only want to move on to
+    // the next step when the user closes the TestsDialog box
     onNext(submitStep)
   }, [submitStep])
 
+  const handleClose = () => {
+    if (verified) {
+      // Move on to the next step
+      onNext()
```
kevinylee1991, line 158: Add a comment here to explain what the onNext() and handleNext() lines do exactly since it is a bit weird. Seems like it is unbinding submitStep and then attempting to move to the next step?

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/tests-dialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+  },
+  tests: {
+    display: 'grid',
+    gridTemplateColumns: '48px 180px auto',
+    margin: theme.spacing(1, 1, 2, 1),
+  },
+  subHeader: {
+    color: theme.palette.common.black,
+  },
+  checkingText: {
```
kevinylee1991, line 23: Feels like it'd make sense to have a typography class for this. Is there any sort of typography class defined in the wireframes for this?

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/tests-dialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+}
+
+const iconsMap = new Map<string, { icon: string; classes: string }>([
+  ['loading', { icon: 'sync', classes: 'fa-spin' }],
+  ['success', { icon: 'check-circle', classes: '' }],
+  ['fail', { icon: 'times-circle', classes: '' }],
+  ['error', { icon: 'exclamation-circle', classes: '' }],
+])
+
+const getIconClass = (status) =>
```
kevinylee1991, line 86: no need to make this a function, just have it be an object and reference the value directly wherever you are using this function currently.

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/tests-dialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+  showDialog,
+  onClose,
+}: Props) => {
+  const classes = useStyles()
+  const [testElements, setTestElements] = useState([])
+  const [showCloseButton, setShowCloseButton] = useState(false)
+  const [showTestCompletionMessage, setShowTestCompletionMessage] = useState(false)
+
+  useEffect(() => {
+    const elements = []
```
kevinylee1991, line 120: This is where using an array of objects instead of Map could make things simpler. Array has some useful methods ex. Array.some() to determine if there are any loading tests, or Array.map to create the array of HTML elements.

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/tests-dialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+      if (status === TestStatus.Loading) {
+        hasLoadingTests = true
+      }
+
+      elements.push(
+        <div key={`${test}-${status}`} className={classes.tests}>
+          {status && (
+            <FontAwesomeIcon
+              solid={status !== TestStatus.Loading}
+              className={clsx(icon.classes, classes[getIconClass(status)])}
```
kevinylee1991, line 134: when you change getIconClass above, just do iconClass[status] here instead

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/tests-dialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+        </div>,
+      )
+    })
+    setTestElements(elements)
+    setShowCloseButton(!hasLoadingTests)
+    setShowTestCompletionMessage(!hasLoadingTests)
+  }, [tests])
+
+  return (
+    <Dialog open={showDialog} fullWidth maxWidth="sm">
```
kevinylee1991, line 152: General concern about the dialog -- can a user click on the outside of the dialog and end up closing it? User may end up being locked out from going to the next step. If so could look into if there's a way to disable closing the dialog by clicking outside of it.

kevinylee1991 (commented):
Some good work here, the dialog looks nice. Had a few concerns about the implementation I listed in the comments, curious to get @jruddell input on it too.

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AddCloudProviderCredentialStep.tsx
knnguy: Yeah, it is quite weird, but I don't know how else to trigger the next step after clicking the Close button.

Before, when the Next button in the form is clicked, the submitStep() function was called and that would determine if the user would move on to the next step or not. But now, we have to call submitStep() first and then move to the step only after the user closes the dialog.

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/tests-dialog.tsx
knnguy: No, a user cannot do that. But thanks for bringing it up! I didn't think about testing that.

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AddCloudProviderCredentialStep.tsx
```diff
@@ -48,6 +50,11 @@ const links = {
   ),
 }
 
+const requiredTests = (provider) =>
+  new Map<string, TestStatus | null>([
```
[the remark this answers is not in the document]
knnguy: I've made the change from a map to an array. Now that I think about it, there's not really a need for a Map. We can just reference each test by index since the order of the tests always stays the same anyways.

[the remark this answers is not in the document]
jruddell: I prefer things to be explicit and guaranteed, of the two solutions (map vs array) there are assumptions that if ever false will break the code in the array implementation (i.e. order of array is always the same). Its probably ok to keep it as an array.. but don't shy away from using Map, Set... etc.

src/app/core/themes/relianceLight.ts
```diff
@@ -268,6 +268,15 @@ export const typography = {
     lineHeight: 'normal',
     letterSpacing: 'normal',
   },
+  body2Italicized: {
```
jruddell, line 271: do we need a specific typography just for specifying italic? in the style guide there were a large number of typography types, but the majority of them were variations on color and style. You should be able to just use body2 and have a class in the component that specifies `fontStyle: 'italic'`

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AddCloudProviderCredentialStep.tsx
```diff
@@ -48,6 +51,15 @@ const links = {
   ),
 }
 
+const testsForAws = [{ name: 'AWS account access', status: null }]
```
jruddell, line 54: Is the backend not supplying a list of tests that need to run / are running? I'm not a fan of hardcoding the tests the backend is going to run as this can become outdated in subsequent backend releases. If we can push to get more granularity on what tests are running that would be good

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AddCloudProviderCredentialStep.tsx
```diff
@@ -73,47 +85,87 @@ const AddCloudProviderCredentialStep = ({
...
   const validatorRef = useRef(null)
 
   const setupValidator = (validate) => {
     validatorRef.current = { validate }
   }
 
+  const updateTestStatus = useCallback(
+    (testIndex, status: TestStatus) => {
+      const newTests = clone(tests)
+      newTests[testIndex].status = status
```
jruddell, line 106: optional chaining here to prevent property 'status' of undefined potential error

jruddell (approved):
👍

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AddCloudProviderCredentialStep.tsx
knnguy: Nope, backend isn't giving us the list of tests.

PR #964, opened 2021-01-15
UX-779/Allow admin users to enable/disable other users in the users table and automatically enable a disabled user when changing their password

knnguy:
**NOTE:**
- 'Enable' and 'Disable' actions are disabled for the active user, meaning that you cannot enable or disable yourself.

### **Added 'Enable' and 'Disable' to the actions toolbar in the users list table**
Screen Shot 2021-01-15 at 11 36 04 AM

Screen Shot 2021-01-15 at 2 50 45 PM

Screen Shot 2021-01-15 at 2 52 04 PM

Screen Shot 2021-01-15 at 2 52 23 PM

### DIFF
--- a/src/app/plugins/account/components/userManagement/users/enable-disable-user-dialog.tsx
+++ b/src/app/plugins/account/components/userManagement/users/enable-disable-user-dialog.tsx
@@ -1,0 +1,54 @@
+import React, { useCallback, useMemo } from 'react'
+import ConfirmationDialog from 'core/components/ConfirmationDialog'
+import Text from 'core/elements/text'
+import useDataUpdater from 'core/hooks/useDataUpdater'
+import { mngmUserActions, mngmUserRoleAssignmentsLoader } from './actions'
+import useDataLoader from 'core/hooks/useDataLoader'
+import { pathStr } from 'utils/fp'
+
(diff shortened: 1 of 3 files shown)


src/app/plugins/account/components/userManagement/users/enable-disable-user-dialog.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const userRoleAssignments = roleAssignments.reduce(
+      (acc, roleAssignment) => ({
+        ...acc,
+        [pathStr('scope.project.id', roleAssignment)]: pathStr('role.id', roleAssignment),
+      }),
+      {},
+    )
+
+    update({
+      id: user.id,
```
kevinylee1991, line 34: If the only thing this is changing is the enabled flag, I think you may only need to include the id and the enabled properties. Could you try it out and see?

kevinylee1991 (approved):
Looks good, just 1 comment below.

src/app/plugins/account/components/userManagement/users/enable-disable-user-dialog.tsx
knnguy: You're right. I don't need to send the username or displayname. However, I do need to include roleAssignments

PR #967, opened 2021-01-19
UX-647/Added self-service SSO configuration

kevinylee1991:
I didn't implement contextLoader in the actions here, might make sense to just not use it at all? Seems like a fast API and important to stay up-to-date. Thoughts?

Screenshots of some of the different screens below

Screen Shot 2021-01-19 at 3 27 37 PM
Screen Shot 2021-01-19 at 3 27 54 PM
Screen Shot 2021-01-19 at 3 28 20 PM
Screen Shot 2021-01-19 at 3 28 41 PM

### DIFF
--- a/src/app/constants.ts
+++ b/src/app/constants.ts
@@ -5,6 +5,7 @@ export const imageUrlRoot = `${appUrlRoot}/images`
 export const helpUrl = `${appUrlRoot}/help`
 export const loginUrl = `${appUrlRoot}/login`
 export const loginWithCookieUrl = `${appUrlRoot}/pmkft/login`
+export const loginWithSsoUrl = `${appUrlRoot}/ssologin`
 export const resetPasswordThroughEmailUrl = '/reset/password'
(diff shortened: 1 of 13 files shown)


src/app/constants.ts
```diff
@@ -5,6 +5,7 @@ export const imageUrlRoot = `${appUrlRoot}/images`
 export const helpUrl = `${appUrlRoot}/help`
 export const loginUrl = `${appUrlRoot}/login`
 export const loginWithCookieUrl = `${appUrlRoot}/pmkft/login`
+export const loginWithSsoUrl = `${appUrlRoot}/ssologin`
```
jruddell, line 8: Why do we need a unique url for sso login? shouldn't that just be a type query string param inside the url on the login page? `/login?type=sso`

jruddell (approved):

src/app/constants.ts
kevinylee1991: i'll try that out

PR #972, opened 2021-02-11
UX-639/Self-service MFA feature

kevinylee1991:
Noted a few things I wanted to get feedback on in the comments. Also waiting on joint discussion with chris/pooja to find out how we want to enable self-service users. It would require a new backend API for it to work with self-service users since they don't have permission to patch their own user in keystone.

Screen Shot 2021-02-10 at 6 09 45 PM
Screen Shot 2021-02-10 at 6 09 57 PM
Screen Shot 2021-02-10 at 6 10 24 PM
Screen Shot 2021-02-10 at 6 10 42 PM
Screen Shot 2021-02-10 at 6 11 09 PM
Screen Shot 2021-02-10 at 6 11 19 PM
Screen Shot 2021-02-10 at 6 31 59 PM


src/app/core/components/validatedForm/FormFieldCard.tsx
```diff
@@ -13,7 +13,7 @@ interface ContainerProps {
   inputsWidth?: number
   middleHeader?: JSX.Element
 }
-const defaultMaxWidth = 800
+const defaultMaxWidth = 932
```
jruddell, line 16: formfieldcard is used in a lot of views, how is this affecting those views?

kevinylee1991: Most of the wireframes actually have a width of 932 for these cards, only the wizards with the double columns (like cloud provider edit for example) have a width of 800. I felt like the 932 might make more sense as the default width, but i'm ok with whatever

src/app/plugins/account/components/user-settings/MFASettingsPage.tsx
```diff
@@ -0,0 +1,382 @@
...
+                    ></TextField>
+                    <CopyToClipboard
+                      copyText={params.mfaSecret}
+                      copyIcon={false}
+                      inline={false}
+                      triggerWithChild
+                    >
+                      {
+                        // Button component doesn't like FontAwesomeIcon as a child even though it works
+                        // @ts-ignore
```
jruddell, line 293: there is an icon button that you should be able to use instead

jruddell (approved):

PR #974, opened 2021-02-18
UX-833/Fix endpoint version replacement to be more specific to avoid …

kevinylee1991:
…replacing the wrong substring

### DIFF
--- a/src/api-client/ApiClient.ts
+++ b/src/api-client/ApiClient.ts
@@ -190,7 +190,7 @@ class ApiClient {
       endpoint = await this.apiServices[clsName].getApiEndpoint()
     }
     if (version !== undefined) {
-      return endpoint.replace('v3', version)
+      return endpoint.replace('/v3', `/${version}`)
     }
     return endpoint
   }


jruddell (approved):

src/api-client/ApiClient.ts
```diff
@@ -190,7 +190,7 @@ class ApiClient {
       endpoint = await this.apiServices[clsName].getApiEndpoint()
     }
     if (version !== undefined) {
-      return endpoint.replace('v3', version)
+      return endpoint.replace('/v3', `/${version}`)
```
roopakparikh, line 193: This will replace ALL "/v3" with ${VERSION} that is also incorrect, it should be much more intelligent and only replace the API Version (there might be other component like a tenant id starting with string v3

jruddell: It'll replace a version string with an equivalent one. This is using string templating to inject a variable into the string denoted by `${}`. We don't control what version of the api we communicate with, the backend provides us a version in the base endpoint string. However we now need to communicate with two separate versions of the API, namely `v3` and `v4`.

roopakparikh: ok understand, how about we trim and replace the string if it ends with /v3 ; I am concerned that we will inadvertently replace something else in the url (similar to the bug we are trying to fix)

jruddell: That's essentially what we are doing. We are constructing the url and when an override is passed we replace. It's only on the base endpoint. The rest of the api url is concatenated after this replace is happening

PR #1012, opened 2021-03-25
couple of bug fixes around aws cluster create

jruddell:


kevinylee1991 (approved):

src/app/core/components/wizard/WizardMeta.tsx
```diff
@@ -31,7 +31,7 @@ export default function WizardMeta<T>({
   fields,
   icon,
   className,
-  showContextPreview = false,
+  showContextPreview = localStorage.getItem('enableContextPreview') === 'true',
```
kevinylee1991, line 34: localStorage can be called by itself? This doesn't need window.localStorage?

kevinylee1991 (approved):

PR #1013, opened 2021-03-25
UX-813/Add other option to SSO provider dropdown

kevinylee1991:
Screen Shot 2021-03-25 at 2 28 44 PM


src/app/plugins/account/components/ssoManagement/sso/SsoPage.tsx
```diff
@@ -88,11 +89,10 @@ const updateSsoSettings = (data, setLoading, setDialogOpened, updateParams) => {
     }
 
     const body = {
-      provider: data.ssoProvider,
+      provider: data.ssoProvider === 'other' ? data.ssoProviderName : data.ssoProvider,
```
jruddell, line 92: update the enum `SsoProviders` to include other, and then reference that here and on line 190

jruddell (approved):

PR #1017, opened 2021-03-29
UX-830/Used different API endpoint to update users' passwords. Added a curre…

knnguy:
…ntPassword field and disabled displayname editing for self-service users

### **User Profile page for self-service users**
- Disabled email and displayname editing since they do not have permission to call `PATCH /v3/users/{user_id}`
- Added a Current Password field since it's required to make the password update
- The tenants and users table has also been removed because again, they do not have permission to get this info

Screen Shot 2021-03-29 at 1 38 01 PM

### **User Profile page for admin users**
- Same as before, just added a Current Password field

Screen Shot 2021-03-29 at 1 34 06 PM

Other Changes:
- Created to action to update the user's password. Instead of calling `PATCH /v3/users/{user_id}`, we now call `POST /v3/users/{user_id}/password`


src/app/core/containers/AuthenticatedContainer.tsx
```diff
@@ -68,6 +68,7 @@ const useStyles = makeStyles<Theme, StyleProps>((theme: Theme) => ({
     position: 'relative',
     display: 'flex',
     width: '100%',
+    height: '100vh',
```
jruddell, line 71: should this be minHeight instead of height?

src/app/plugins/account/components/user-settings/my-profile-page.tsx
```diff
@@ -113,46 +64,26 @@ const MyProfilePage = () => {
   const [oldUserPrefs, setOldUserPrefs] = useState(prefs)
   const [userUpdated, setUserUpdated] = useState(false)
   const userIdFormFieldSetter = useRef(null)
+  const [updatingPassword, setUpdatingPassword] = useState(false)
 
   const session = useSelector<RootState, SessionState>(prop(sessionStoreKey))
   const { userDetails } = session
-  const { id: userId, displayName, email } = userDetails
+  const { id: userId, displayName, email, role } = userDetails
+  const isAdmin = role === 'admin'
```
jruddell, line 72: there is a helper that checks for admin role, use that please

src/app/plugins/account/components/userManagement/users/actions.js
```diff
@@ -26,6 +26,27 @@ const authMethods = {
   [LoginMethodTypes.SSO]: async (_u, _p, _t) => keystone.authenticateSso(),
 }
 
+const reauthenticateUser = async ({ email, currentPassword }) => {
+  const state = store.getState()
```
jruddell, line 30: lets avoid side effects like this as much as possible. Can you pass the current tenant id through to this function instead?

src/app/plugins/account/components/userManagement/users/actions.js
```diff
@@ -291,3 +289,23 @@ export const mngmUserRoleAssignmentsLoader = createContextLoader(
     indexBy: 'userId',
   },
 )
+
+export const updateUserPassword = async ({ id, email, currentPassword, newPassword }) => {
+  try {
+    // User must be authenticated before making a password change request
+    await reauthenticateUser({ email, currentPassword })
```
jruddell, line 296: shouldn't the user already be authenticated at this point?

knnguy: You're right. I was trying to verify that the old password was correct first but now that I think about, it's quite redundant.

jruddell (approved):

PR #1019, opened 2021-04-02
UX-287/Added alarm snooze feature

kevinylee1991:
Some notes:

- Found that watchdog alarm has the same fingerprint across clusters, so I changed the alarm ID
- Found that the label for pf9-monitoring does not return a bool, but a string so changed a helper function appropriately
- Found that the prometheus APIs call across all clusters, changing parseClusterParams function in the cluster actions.js file to use the selector seems to make more sense. It doesn't use the selector version since it calls the list action directly, thus the params aren't being used for anything at the moment bc the cluster list action doesn't read any parameters.

Screen Shot 2021-04-01 at 6 10 20 PM


src/app/plugins/kubernetes/components/infrastructure/importedClusters/helpers.ts
```diff
@@ -0,0 +1,6 @@
+import { pathEq } from 'ramda'
+
+export const importedHasPrometheusTag = pathEq(
```
jruddell, line 3: shouldn't appbert be the source of truth for if prometheus is deployed? From my understanding appbert is supposed to work with imported clusters the same as internal clusters

kevinylee1991: I can check with backend to see if the tag is reliable -- I know for the normal clusters we are using the tags so I figured that the imported ones could use them as well.

PR #1024, opened 2021-04-09
UX-425/Added Zendesk widget

knnguy:
**NOTES:** 

- Chat option is only available to enterprise users on PMK
- Chat widget only shows/opens when you click on the Chat button in either the help page or the avatar dropdown menu

Screen Shot 2021-04-12 at 1 32 36 PM


src/app/utils/zendesk-widget.js
```diff
@@ -0,0 +1,118 @@
+import axios from 'axios'
+import { DocumentMetaCls } from 'core/components/DocumentMeta'
+
+const zendeskScriptUrl =
+  'https://static.zdassets.com/ekr/snippet.js?key=5d4d85e0-ef42-4308-8aad-70a4a0a92b0d'
+const jwtTokenEndpoint = 'https://cs.pf9.us/support/zd/widget-auth'
+const authErrorMsg = 'Error authenticating Zendesk widget: '
+
+export const showZendeskWidget = () => {
+  if (window.zE) {
```
jruddell, line 10: these if checks are ok, but generally a better pattern is to do negative checks. This way you aren't indenting your code per if statement. I.E. if there are 5 nested if statements you have a lot of indented code which is harder to read. So I prefer 

```
if (!window.zE) {
  return
}
zE('webWidget', 'show')
```

src/app/utils/zendesk-widget.js
```diff
@@ -0,0 +1,118 @@
+import axios from 'axios'
```
jruddell, line 1: can this be a typescript file?

PR #1026, opened 2021-04-14
UX-304/Monitoring overview dashboard

kevinylee1991:
Not totally finished yet, need some last few q's answered from Chris (like if we need to add a refresh button, if imported vs normal clusters need to be treated differently, heights of the boxes if there's not much data, pagination).

Also regarding the grafana icon in the zeplin, Emi just added the assets in and then asked me the following: "Are you still using the icons in font format or everything as svgs? I've added some of the other icons that were used in the comp to the icon collection screen too." I'm not exactly sure how to answer that, so I told her I'd ask John about it.

Also one more thing I wanted to confirm was if I should create a new loader & selector that combines both the normal clusters and the imported clusters together. Or if it's possible to just use a selector to pull together existing selectors and have a loader that acts like a proxy for the selector? I don't like the way I currently have it where I'm querying clusters & imported clusters separately.

Screen Shot 2021-04-13 at 6 18 19 PM
Screen Shot 2021-04-13 at 6 18 35 PM
Screen Shot 2021-04-13 at 6 18 52 PM


src/app/plugins/kubernetes/components/monitoring/ClusterAlarmButton.tsx
```diff
@@ -0,0 +1,98 @@
...
+const useStyles = makeStyles<Theme, Partial<StyleProps>>((theme: Theme) => ({
+  alarmButton: {
+    width: 180,
+    height: 40,
+    display: 'inline-flex',
+    alignItems: 'center',
+    border: ({ active, type }) => {
+      if (active) {
+        if (type === 'fatal') {
+          return `1px solid #f9dcd8`
```
jruddell, line 22: shouldnt these colors be from the theme?

jruddell (approved):

PR #1031, opened 2021-04-16
UX-762/Removed cluster upgrade availability out of cluster health status

knnguy:
**NOTES:** 

- Cluster upgrade availability is no longer part of the cluster health status
- Cluster upgrade availability is now shown as a link in the K8s Version column in the clusters list table
- Link action has changed. Instead of going to the nodes details page, it will now open up the cluster upgrade dialog.

Screen Shot 2021-04-16 at 3 15 43 PM

Since it's no longer part of cluster health, 'Upgrade Available' will no longer show up in the cluster status card. 

Screen Shot 2021-04-16 at 4 24 45 PM

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.tsx
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.tsx
@@ -19,7 +19,7 @@ import ClusterUpgradeDialog from 'k8s/components/infrastructure/clusters/Cluster
 import PrometheusAddonDialog from 'k8s/components/prometheus/PrometheusAddonDialog'
 import { ActionDataKeys } from 'k8s/DataKeys'
 import { both, omit, prop } from 'ramda'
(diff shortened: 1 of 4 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.tsx
```diff
@@ -127,6 +127,19 @@ const renderNodeLink = (_, { uuid, nodes }) => {
 
 const renderStats = (_, { usage }) => <ResourceUsageTables usage={usage} />
 
+const renderK8sVersion = (_, { uuid, version, canUpgrade }) => {
+  return (
+    <div>
+      <Text variant="body2">{version}</Text>
+      {canUpgrade && (
```
jruddell, line 134: it probably makes more sense to have this prompt the cluster upgrade modal

jruddell (approved):

PR #1036, opened 2021-04-23
UX-303/Added cluster details alarm tab

kevinylee1991:
Note: this is based off of ux-304 bc there is some code in there I am making use of, would be easier to review after 303 is pushed in

Screen Shot 2021-04-23 at 1 13 41 PM

### DIFF
--- a/src/app/core/components/graphs/StackedAreaChart.tsx
+++ b/src/app/core/components/graphs/StackedAreaChart.tsx
@@ -59,6 +59,7 @@ interface Props<T extends string, V extends string> {
   xAxis: T
   keys: Array<AreaChartType<V>>
   responsive?: boolean
+  responsiveHeight?: number
   verticalAxisLines?: boolean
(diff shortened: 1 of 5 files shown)


src/app/plugins/kubernetes/components/monitoring/ClusterAlarmButton.tsx
```diff
@@ -0,0 +1,98 @@
...
+const useStyles = makeStyles<Theme, Partial<StyleProps>>((theme: Theme) => ({
+  alarmButton: {
+    width: 180,
+    height: 40,
+    display: 'inline-flex',
+    alignItems: 'center',
+    border: ({ active, type }) => {
+      if (active) {
+        if (type === 'fatal') {
+          return `1px solid #f9dcd8`
```
jruddell, line 22: shouldn't these colors be from the theme?

jruddell (approved):

PR #1040, opened 2021-04-27
UX-812/Allow search on list tables on multiple properties

kevinylee1991:
Screen Shot 2021-04-26 at 5 27 16 PM

### DIFF
--- a/src/app/core/components/listTable/ListTable.js
+++ b/src/app/core/components/listTable/ListTable.js
@@ -318,7 +318,7 @@ class ListTable extends PureComponent {
   }
 
   handleSearch = (value) => {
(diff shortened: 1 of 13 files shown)


jruddell (approved):
This is ok, can merge for now. One thing that might be better is to just stringify the whole record and do a match off of that. Concern here is you have to explicitly specify which columns to search off of. So we can have the same issue again where a user wants to search by something and it doesn't find a match.

kevinylee1991:
One downside to stringifying the whole record is that there are some nested fields that could have a lot of random stuff that are never shown on the column, just something to keep in mind

PR #1042, opened 2021-04-27
UX-897/Modify node details page and node hardware requirements

knnguy:
**Changes made:**

- Added a 'Kubernetes Node Details' card to the Node Details Page

Screen Shot 2021-05-12 at 2 46 02 PM

- Added hardware requirements and clock drift status to the node header card
- When user hoovers over the icons in the header card, a message appears telling the user what the recommended hardware requirements are

https://user-images.githubusercontent.com/23369276/118053879-2d096e80-b33a-11eb-97b5-63c22c632a53.mov

- Error message is shown if clock drift is detected in the node

Screen Shot 2021-05-12 at 3 58 59 PM

- Modified the hardware requirements in the BareOs virtual and physical  cluster creation page

Screen Shot 2021-05-12 at 3 42 33 PM


knnguy:
Is `startKube` supposed to have a boolean value?

knnguy:
Also, I did not fully test the clock skew message display. We did not have any nodes that had a clock skew.

jruddell:
Related to the design of this, can we look at reusing what we do for the cluster create requirements? where we call out certain specs that are required? if we have these available, would be appropriate to put it in the header card. give it some more content

src/app/plugins/kubernetes/components/infrastructure/clusters/helpers.ts
```diff
@@ -327,3 +327,33 @@ const createGenericCluster = async (body, data) => {
...
+  condition?: (cluster: T) => boolean
+  render?: (value: any, item: any) => string | React.ReactNode
+}
+
+/**
+ * Gets fields for the InfoPanel component
+ *
+ * Ex. getFieldsForCard(fields, cluster)
+ */
+export function getFieldsForCard<T>(fields: Array<IDetailFields<T>>, item: T) {
```
jruddell, line 345: maybe we should move this helper to the InfoPanel component? its not really cluster specific.

src/app/plugins/kubernetes/components/infrastructure/nodes/NodeInfo.tsx
```diff
@@ -0,0 +1,185 @@
...
+  },
+  {
+    id: 'nodePoolName',
+    title: 'Node Pool Name',
+  },
+  {
+    id: 'clusterUuid',
+    title: 'Cluster Unique ID',
+  },
+  {
```
jruddell, line 80: uuid is duplicated

src/app/plugins/kubernetes/components/infrastructure/clusters/helpers.ts
knnguy: I moved this helper to the InfoPanel component. InfoPanel is still under `core/components`. Not sure what generic entities stuff you're referring to though. Is it `core/components/entities-page`?

knnguy:
> Related to the design of this, can we look at reusing what we do for the cluster create requirements? where we call out certain specs that are required? if we have these available, would be appropriate to put it in the header card. give it some more content

Are you referring to the callout fields in the WizardMeta?

Screen Shot 2021-05-04 at 11 22 26 AM

Or the requirements here?

Screen Shot 2021-05-04 at 11 24 52 AM

jruddell:
> > Related to the design of this, can we look at reusing what we do for the cluster create requirements? where we call out certain specs that are required? if we have these available, would be appropriate to put it in the header card. give it some more content
> 
> Are you referring to the resource requirements here?
> 
> Screen Shot 2021-05-04 at 11 24 52 AM
> 
> If yes, I think it would more appropriate if I do it in UX-763 or in a new ticket. In UX-763, I'm working on modifying the node resource requirements and node chooser table. However, I'm still currently waiting for the backend to include cpu info in the api response

Yes, the hardware specs is what i was referring to. I dont want us to implement a poorly designed card and then update it. lets just do it correctly now so we dont put it infront of customers. Can you figure out what we can show and then we can discuss the design

knnguy:
> > I just made a little fix. I realized that the bool values should be True/False, not Enabled/Disabled. Now that I think about it, maybe Yes/No makes more sense? Also, is `startKube` supposed to be a boolean value?
> 
> I think we decided to standardize on boolean fields to say Enabled/Disabled instead of True/False. Probably best to keep it that way

Even for fields like 'Is Master', 'API Responding', etc? I personally think it's a bit weird, but I can change it

knnguy:
> Yes, the hardware specs is what i was referring to. I dont want us to implement a poorly designed card and then update it. lets just do it correctly now so we dont put it infront of customers. Can you figure out what we can show and then we can discuss the design

In [UX-763](https://platform9.atlassian.net/jira/software/projects/UX/boards/96?selectedIssue=UX-763), I am modifying the hardware requirements. Here are some screenshots of the changes I've made so far for that ticket.

Screen Shot 2021-05-12 at 11 42 09 AM

Screen Shot 2021-05-12 at 11 45 03 AM

As you can see, we're still missing the # of CPUs. I filed a [CORE](https://platform9.atlassian.net/browse/CORE-1174?page=com.atlassian.jira.plugin.system.issuetabpanels:all-tabpanel) ticket and am waiting for that to be completed.
If you want the hardware requirements to be in the header card, I can just use the old requirements for now.

knnguy:
If a node is not attached to a cluster, should we still display the hardware requirements in the header card? If yes, should we compare the node's hardware specs to a single node cluster hardware requirement or a multi-node cluster's hardware requirement?

Current behavior is that even if a node is not attached to a cluster, we display the hardware specs in the header card and compare it against a multi-node cluster's hardware specs

jruddell (approved):

chrisj-platform9:
@knnguy Please stick with your current design of using the multi-node hardware requirements. 
```
If a node is not attached to a cluster, should we still display the hardware requirements in the header card? If yes, should we compare the node's hardware specs to a single node cluster hardware requirement or a multi-node cluster's hardware requirement?

Current behavior is that even if a node is not attached to a cluster, we display the hardware specs in the header card and compare it against a multi-node cluster's hardware specs. Not sure if that makes sense or not
```

PR #1052, opened 2021-05-06
UX-918/Added hoover help to upgrade cluster text

knnguy:
Screen Shot 2021-05-06 at 11 14 49 AM


knnguy:
The hoover help text is:
"An upgrade to newer version of Kubernetes is now available for this cluster. Click here or select the upgrade action for the cluster from infrastructure -> clusters view to see more details"

I personally think the "from infrastructure -> clusters view" is unnecessary. We could just shorten it to 
"An upgrade to newer version of Kubernetes is now available for this cluster. Click here or select the upgrade action for the cluster to see more details"

@jruddell  what do you think?

jruddell (approved):

PR #1065, opened 2021-05-20
UX-974/Track untracked actions and leave Bugsnag breadcrumbs for each path change and each action

knnguy:
Changes made:
- Tracked untracked actions for Segment
- Added breadcrumbs with type 'navigation' for each UI path change. These correspond with the tracked pages in Segment. 
- Added breadcrumbs for each action, such as creating clusters, deleting node, scaling masters, etc.
- Added customer tier and du version into the App tab in Bugsnag

Screenshots taken from Bugsnag

Screen Shot 2021-05-20 at 11 45 57 AM

Screen Shot 2021-05-20 at 2 54 17 PM


knnguy:
I'm assuming that this is the DU version in features.json. Is this correct? How about for DDUs?
Screen Shot 2021-05-20 at 1 11 46 PM

src/app/core/containers/AppContainer.tsx
```diff
@@ -143,7 +145,12 @@ const AppContainer = () => {
 
   useEffect(() => {
     const unlisten = history.listen((location) => {
-      trackPage(`${location.pathname}${location.hash}`)
+      const url = `${location.pathname}${location.hash}`
```
jruddell, line 148: wouldnt the getCurrentRoute on the Route class do the same thing?

jruddell (approved):
Really awesome stuff here, great work!

knnguy:
I made a small change where I moved `Bugsnag.addBreadcrumb` for each route into the `trackPage` function so that we can just call `Route.find` once and track the route for both Segment and Bugsnag at the same time

jruddell (approved):

PR #1067, opened 2021-05-21
UX-948/change-sku-to-Standard_A4_v2-for-Azure

hgupta2363:
Screenshot 2021-05-21 at 10 03 02 AM


src/app/plugins/kubernetes/components/infrastructure/clusters/azure/create-templates/custom.tsx
```diff
@@ -119,8 +119,8 @@ const networkOptions = [
 ]
 
 const templateOptions = [
-  { label: 'Sm - Single Node Master & Worker (Standard_A1_v2)', value: 'small' },
-  { label: 'Md - 1 Master + 3 Workers (Standard_A2_v2)', value: 'medium' },
+  { label: 'Sm - Single Node Master & Worker (Standard_A4_v2)', value: 'small' },
+  { label: 'Md - 1 Master + 3 Workers (Standard_A4_v2)', value: 'medium' },
```
jruddell, line 123: all you are doing here is just changing a string for presentation purposes. This doesn't change the template. further down the file there is a function called `handleTemplateChoice` that is what actually sets the SKU, update those strings to be the correct value please. Also, since its always the same lets remove `(Standard_A4_v2)` from the option text and instead put the SKU type in the label of the dropdown itself

jruddell (changes requested):
all you are doing here is just changing a string for presentation purposes. This doesn't change the template. further down the file there is a function called `handleTemplateChoice` that is what actually sets the SKU, update those strings to be the correct value please. Also, since its always the same lets remove `(Standard_A4_v2)` from the option text and instead put the SKU type in the label of the dropdown itself

jruddell (changes requested):
One missing piece of the feedback i left is that we use this template name as the label of the dropdown component. `Standard_A4_v2` needs to be added to `form-components/cluster-templates.tsx` under the `label` prop. This way the developers know what template is being used still

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/cluster-templates.tsx
```diff
@@ -2,7 +2,7 @@ import PicklistField from 'core/components/validatedForm/PicklistField'
 import React from 'react'
 
 const ClusterTemplatesField = ({ options, onChange }) => (
-  <PicklistField id="template" label="Cluster Template" options={options} onChange={onChange} />
+  <PicklistField id="template" label="Standard_A4_v2" options={options} onChange={onChange} />
```
jruddell, line 5: Think about this, if you are trying to use the UI and all you see is the title of a dropdown with this text. What does that mean? Obviously we should keep the `Cluster Template` title and rather like I was saying in my prior comments add `(Standard_A4_v2)` to the end so they know what template they are dealing with.

hgupta2363: oh yeah, this gave me a lesson. Read the requirements carefully

jruddell (approved):

PR #1068, opened 2021-05-21
UX-975/Change how we get the clock drift error message

knnguy:
**Description**

Backend has changed the host message structure so we need to change how we retrieve the clock drift error message.

New message structure:

```
"message": {
        "warn": [
            {
                "code": "HOST_TIMEDRIFT_DETECTED", 
                "message": "Host clock may be out of sync"
            },
            {
                "code": "CERTS_EXPIRED",
                "message": "Certificates are about to expire."
            }
        ],
        "Error":  [{
            "code": "..",
            "message": "..."
        }]
    }
```

The backend change will be in 5.3


jruddell (approved):
What version of the backend do we need to sync this change with? can you make sure it is backported if the backend change is going to be backported?

knnguy:
> What version of the backend do we need to sync this change with? can you make sure it is backported if the backend change is going to be backported?

They will be in 5.3 and they will not be backported unless someone insists on it

PR #1069, opened 2021-05-22
Only add Zendesk script if in prod env

knnguy:


src/app/core/containers/AuthenticatedContainer.tsx
```diff
@@ -374,7 +374,7 @@ const AuthenticatedContainer = () => {
 
   // Add Zendesk widget script only for Enterprise users
   useEffect(() => {
-    if (customerTier === CustomerTiers.Enterprise) {
+    if (customerTier === CustomerTiers.Enterprise && process.env.NODE_ENV === 'production') {
```
jruddell, line 377: lets make a helper for checking the environment since i believe we use this in more than one place

jruddell (approved):

PR #1075, opened 2021-06-02
UX-736, UX-978, UX-932/ Add cloud provider defaults from preference store to AWS and Azure cloud provider and cluster creation forms

knnguy:
**Changes:**
- Allow users to set cloud provider defaults like region and sshKey in the AWS and Azure cloud provider create and update wizard
- Added sshKey field in Azure cloud verification form(Can be seen in Azure cloud provider create and update wizard)
- Populate the cloud provider defaults in the AWS and Azure cluster creation wizard
- Modified the texts in the cloud verification cards
- Added user default values in the callout fields in WizardMeta

**Cloud Defaults**
- All AWS Cloud providers -> One User Default Region, SSH Key, etc (Same for Azure)

### **Create Cloud Provider Page**

Screen Shot 2021-06-02 at 1 49 04 PM

Screen Shot 2021-06-02 at 1 49 04 PM

### **Azure Update Page**

Screen Shot 2021-06-14 at 5 57 53 PM

### **Validation Tests Dialog**

Screen Shot 2021-06-02 at 1 58 03 PM

### **Cloud Defaults Page (Step 2 of cloud provider create wizard)**

Screen Shot 2021-06-02 at 1 56 39 PM

Screen Shot 2021-06-02 at 2 00 41 PM

### **Cluster Create Wizard (One-Click and Custom)** 
- The cloud provider defaults are populated in the wizard AFTER the user selects a cloud provider. This is just to maintain the order in the form. The cloud provider ID is need to populate the options in the other fields

https://user-images.githubusercontent.com/23369276/120547519-150b9480-c3a6-11eb-9adf-f263f3845e5a.mov

### **Where the cloud defaults are stored in the Redux store**
Screen Shot 2021-06-02 at 2 05 30 PM


knnguy:
### **Questions**

@jruddell 
I loaded all the user pref values  in `AuthenticatedContainer.tsx`. I basically looped through a list of all the keys in the pref store and called the API to get the values. Not sure if that is the best way or place to do that. I was thinking it's best to load everything once the user is authenticated, but the API doesn't provide us with a way to get all the keys/values at once. 

After that, updating and getting the user pref values are done through the `useScopedPreferences()`hook. The `updateUserDefaults` function in `useScopedPreferences()` calls the API to update the key/value and updates the key/value in the store

@chrisj-platform9 
1. There's currently no way for the user to delete a default value right now. Is that something we intend to add in later or is that not an option for the user?
2. How do we ensure that the user will choose the right defaults? For example, sshKey options depend on cloud region. What if the user sets a default region first and then goes back later to set a default sshKey that is not available for that region? Right now, it's possible for them to do that because each cloud default is set separately.

chrisj-platform9:
**There's currently no way for the user to delete a default value right now. Is that something we intend to add in later or is that not an option for the user?**
 - I guess we could add a 'delete' on the cloud provider edit screen next to where the default is. I think Emi had this in her design. 
 - 
**How do we ensure that the user will choose the right defaults? For example, sshKey options depend on cloud region. What if the user sets a default region first and then goes back later to set a default sshKey that is not available for that region? Right now, it's possible for them to do that because each cloud default is set separately.**
 - This is a good point. Maybe setting the Default SSH key automatically updates the default region??

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AwsCloudProviderFields.tsx
```diff
@@ -69,56 +72,69 @@ const AwsCloudProviderFields = ({
   showSubmitInCard = false,
   updateWizard = false,
   errorMessage = '',
+  showInfo = true,
+  setDefaultValueForTextfields = false,
+  defaultTextfieldValue = '**********************',
```
jruddell, line 77: whats this for?

knnguy: Chris wanted the fields to be pre-populated with the asterisks

Screen Shot 2021-06-14 at 5 56 47 PM

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/helpers.tsx
```diff
@@ -20,3 +26,84 @@ export const RegionAvailability = ({ classes, regions }) => {
...
+              <FontAwesomeIcon className={classes[iconClass]} solid>
+                {icon}
+              </FontAwesomeIcon>
+              <Text variant="body2" className={classes.spaceRight}>
+                {label}
+              </Text>
+            </div>
+            <Text
+              variant="caption1"
+              className={clsx(classes.value, value ? '' : classes['noneText'])}
```
jruddell, line 100: dont use bracket notation to access properties on an object, use dot notation. Dot notation works better with typescript `classes.noneText`

jruddell (approved):
Few minor changes, we will want to move the preference loading to these components themselves vs loading all in the authenticated container. but that can be a future update to this, the concern being maintainability of a large list of items to load on startup isn't a good model. But for now lets go ahead and merge this, great work on getting it all hooked up!

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AwsCloudProviderFields.tsx
jruddell: Shouldn't that just be a placeholder? I don't really like setting the wizard context to contain those literal strings. Would be afraid somehow that gets sent to the backend.

PR #1076, opened 2021-06-04
 UX-906/Expose uploaded metadata XML in a codemirror instead of not displaying it

kevinylee1991:
Screen Shot 2021-06-04 at 4 16 54 PM

### DIFF
--- a/src/app/constants.ts
+++ b/src/app/constants.ts
@@ -55,7 +55,7 @@ export enum CustomerTiers {
 }
 
 // Don't forget to revert ssoEnabledTiers back to normal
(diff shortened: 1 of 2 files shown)


src/app/constants.ts
```diff
@@ -54,6 +54,7 @@ export enum CustomerTiers {
   OEM = 'oem',
 }
 
+// Don't forget to revert ssoEnabledTiers back to normal
```
jruddell, line 57: do you need this comment?

jruddell (approved):

PR #1081, opened 2021-06-10
UX-924/cluster-upgrade design created

rajshekhar23:
The update cluster form is on a new page, with a few new options:

* Sequential upgrades - 1 by 1 node upgrades
* Percentage upgrades - % of nodes to be upgraded in parallel 
* Batch upgrades - advanced flow to specify batches to upgrade to

### Default view
image

### Percentage view
image

### Batch view
image

After creating a cluster and upgrading one node via a batch, this is the view once the cluster was ready for another upgrade
image


package.json
```diff
@@ -49,7 +49,8 @@
     "totp-generator": "^0.0.7",
     "url-pattern": "^1.0.3",
     "use-react-router": "^1.0.7",
-    "uuid": "^3.2.1"
+    "uuid": "^3.2.1",
+    "yarn": "^1.22.10"
```
jruddell, line 53: why is yarn added as a dependency?

src/app/core/components/validatedForm/radio-fields.tsx
```diff
@@ -1,22 +1,30 @@
 import React, { FC, PureComponent } from 'react'
-import { FormControl, FormControlLabel, FormHelperText, Radio, WithStyles } from '@material-ui/core'
```
jruddell, line 2: probably best to just remove the changes to this file since we didnt go forward with this work

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,191 @@
...
+      return b.filter((valueB) => {
+        return valueB.name
+          ? valueB.name.indexOf(valueA.name) === -1
+          : valueB.indexOf(valueA.name) === -1
+      })
+    }
+  })
+}
+
+function intersection(a, b) {
```
jruddell, line 49: we have a library called Ramda that we use for things like this. I believe they have an `intersection` method you can use

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,191 @@
...
+    if (b.length > 0) {
+      return b.filter((valueB) => {
+        return valueB.name ? valueB.name.indexOf(valueA) !== -1 : valueB.indexOf(valueA) !== -1
+      })
+    }
+  })
+}
+
+export default function TransferList({ clusterNodes }) {
+  console.log('clusterNodes', clusterNodes)
```
jruddell, line 60: console.log

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const classes = useStyles()
+  const [checked, setChecked] = React.useState([])
+  const [searchTerm, setSearchTerm] = React.useState('')
+  const [left, setLeft] = React.useState(clusterNodes)
+  const [right, setRight] = React.useState([])
+
+  const leftChecked = intersection(checked, left)
+  const rightChecked = intersection(checked, right)
+  console.log(leftChecked, rightChecked)
```
jruddell, line 69: console.log

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,191 @@
...
+    <Grid container spacing={2} justify="center" alignItems="center" className={classes.root}>
+      <Grid item>
+        <SearchBar
+          className={classes.search}
+          onSearchChange={handleSearchChange}
+          searchTerm={searchTerm}
+        />
+        {customList(left)}
+      </Grid>
+      <Grid item>
```
jruddell, line 139: we dont use grid much at all. Lets try to avoid the unnecessary elements

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,191 @@
...
+        <Grid container direction="column" alignItems="center">
+          {/* <Button
+            variant="outlined"
```
jruddell, line 142: remove commented out code

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,241 @@
...
+    ...togglersMapping,
+    [id]: field,
+  }
+}
+
+const UpgradeClusterPage = () => {
+  const [clusters] = useDataLoader(clusterActions.list)
+  const { match, history } = useReactRouter()
+  const cluster = clusters.find((x) => x.uuid === match.params.id)
+  console.log('cluster', cluster)
```
jruddell, line 69: console.log

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,241 @@
...
+                            </RadioGroup>
+                          </FormFieldCard>
+                          <FormFieldCard title="Upgrade Strategy">
+                            <AddonTogglers
+                              addons={clusterAddons}
+                              wizardContext={wizardContext}
+                              setWizardContext={setWizardContext}
+                            />
+                          </FormFieldCard>
+                          {/* <FormFieldCard title="Upgrade Strategy">
```
jruddell, line 184: remove commented out code

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/cluster-node-upgrade-strategy.tsx
```diff
@@ -0,0 +1,87 @@
...
+    sequentialStrategy: field === 'sequentialStrategy' ? true : false,
+    percentageStrategy: field === 'percentageStrategy' ? true : false,
+    listStrategy: field === 'listStrategy' ? true : false,
+  }
+}
+
+export const SequentialUpgradeField = ({ wizardContext, setWizardContext }) => {
+  return (
+    <RadioFields
+      value={wizardContext['sequentialStrategy']}
```
jruddell, line 26: use `.` notation for accessing properties instead of `['`

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/percentage-cluster-upgrade.tsx
```diff
@@ -0,0 +1,53 @@
...
+      <Text variant="body2">
+        <b>Total Worker Nodes:</b>
+        {values?.workerNodes?.length}
+      </Text>
+    </IconInfo>
+    <TextField
+      id="percentage"
+      label="Worker nodes to upgrade in parallel"
+      type="number"
+      step="1"
```
jruddell, line 47: we should have a min and max range for this input. 1 -> 100

src/stories/config.js
```diff
@@ -1,66 +1,47 @@
-import StoryRouter from 'storybook-react-router'
```
jruddell, line 1: 2 things. you shouldn't be pushing the config.js file, make sure this is removed. This is also inside stories which is the wrong location

jruddell (changes requested):
couple changes in code style and some cleanup, make sure you always review the change log before making the pull request. Also I didn't see anything about you sending this data in the API request for the backend to use. Just because the element has been added to the page and its data is tracked in the wizardContext, you still need to pass the data through in the API call. look at how we generate the payload for creating a cluster (cluster/actions.js has the create function, it calls helper methods to build the payload inside cluster/helpers.ts)

package.json
rajshekhar23: @jruddell  may be i tried multiple ways to make yarn working. globally installed yarn was not working so i tried to install that only current repository.

I will remove it quickly

src/api-client/Qbert.ts
```diff
@@ -392,16 +392,21 @@ class Qbert extends ApiService {
...
       url,
       version: 'v4',
       options: {
         clsName: this.getClassName(),
-        mthdName: 'upgradeCluster',
+        mthdName: 'upgradeClusterNodes',
       },
-    })
+    }
+    if (body) {
```
jruddell, line 405: why do you need to do this?

rajshekhar23: @jruddell to differentiate three scenarios based on body value.
if something is there in body ( batchUpgradeNodes or batchUpgradePercent) then will go inside body params.
else body wont get added into end point.

rajshekhar23: @jruddell Updated.

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/advanced-batch-upgrade.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export const AdvancedBatchUpgradeAddonField = ({ values }) => {
+  const {
+    wizardContext,
+    setWizardContext,
+  }: { wizardContext: any; setWizardContext: any } = useContext(WizardContext) as any
+  const classes = useStyles()
+  const nodeVersionToCompare = getNodeVersionToCompare(wizardContext)
+  const [upgradedNodes, toBeUpgradedNodes] = partition(
```
xantorres, line 47: please use `useMemo` here, it looks like an expensive operation

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterBatchUpgradeDialog.tsx
```diff
@@ -0,0 +1,114 @@
...
+  wizardContext,
+  handleClose,
+  confirmBatchUpgrade,
+  cluster,
+}) => {
+  const classes = useStyles({})
+  const nodeVersionToCompare = getNodeVersionToCompare(cluster)
+
+  // partion node list based upgradingTo flag ( in progress / intial )
+  const [upgradedNodes, toBeUpgradedNodes] = partition(
```
xantorres, line 62: `useMemo`

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,238 @@
...
+      if (wizardContext.sequentialClusterUpgrade) {
+        upgradeClusterNodes(wizardContext)
+        onComplete()
+      }
+
+      if (wizardContext.advancedBatchUpgrade) {
+        handleOpen()
+      }
+    },
+    [upgradeType],
```
xantorres, line 120: This dependency is not used anywhere in the callback function, so you can remove it

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,238 @@
...
+      upgradeChoice === 'minor' ? cluster.minorUpgradeRoleVersion : cluster.patchUpgradeRoleVersion
+    setWizardContext({ selectedUpgradeVersion })
+  }, [])
+
+  const confirmBatchUpgrade = useCallback(
+    (wizardContext) => {
+      upgradeClusterNodes(wizardContext)
+      onComplete()
+    },
+    [upgradeClusterNodes],
```
xantorres, line 136: You can remove this dependency here.

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,238 @@
...
+                    <ValidatedForm
+                      classes={{ root: classes.validatedFormContainer }}
+                      fullWidth
+                      initialValues={wizardContext}
+                      onSubmit={setWizardContext}
+                      triggerSubmit={onNext}
+                      elevated={false}
+                      withAddonManager
+                    >
+                      {({ setFieldValue, values }) => (
```
xantorres, line 173: Why are we even using `ValidatedForm` here if we are not setting the form values or doing any validation? You are essentially bypassing the `ValidatedForm` and just using the `Wizard` values

src/app/plugins/kubernetes/components/infrastructure/clusters/helpers.ts
```diff
@@ -364,3 +366,20 @@ const createGenericCluster = async (body, data) => {
...
+
+export const createBatchUpgradeClusterPayload = async (data) => {
+  const keysToPluck = ['batchUpgradeNodes', 'batchUpgradePercent']
+  const body = pick(keysToPluck, data)
+  if (data?.batchUpgradeNodes)
+    body.batchUpgradeNodes = pluck('uuid' as any, data?.batchUpgradeNodes)
+
+  if (data.batchUpgradePercent) body.batchUpgradePercent = data.batchUpgradePercent
+
+  Bugsnag.leaveBreadcrumb('Attempting to upgrade cluster on percentage', {
```
xantorres, line 378: Not a fan of moving the tracking stuff to the helpers file

src/app/plugins/kubernetes/components/monitoring/AlarmOverviewClusters.tsx
```diff
@@ -127,14 +127,14 @@ const AlarmOverviewClusters = () => {
           </Text>
         )}
         {filteredClusters.map((cluster) => (
-          <div className={classes.clustersList} key={cluster.uuid}>
-            <div key={cluster.uuid} className={classes.clusterRow}>
-              <Text variant="body2">{cluster.name}</Text>
+          <div className={classes.clustersList} key={cluster?.uuid}>
```
xantorres, line 130: If a `cluster` is nil something has had to go very wrong before we got here

yarn.lock
```diff
@@ -16,10 +16,10 @@
   dependencies:
     "@babel/highlight" "^7.12.13"
 
-"@babel/compat-data@^7.13.0", "@babel/compat-data@^7.13.8":
-  version "7.13.8"
-  resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.13.8.tgz#5b783b9808f15cef71547f1b691f34f8ff6003a6"
-  integrity sha512-EaI33z19T4qN3xLXsGf48M2cDqa6ei9tPZlfLdb2HC+e/cFtREiRd8hdSqDbwdLB0/+gLwqJmCYASH0z2bUdog==
+"@babel/compat-data@^7.13.11", "@babel/compat-data@^7.14.4":
```
xantorres, line 19: Why has this file been updated?

xantorres (changes requested):

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+const useStyles = makeStyles((theme) => ({
+  listWrapper: {
+    display: 'grid',
+    gridTemplateColumns: '360px auto 360px',
+    justifyContent: 'space-between',
+    minHeight: '300px',
+  },
+  search: {
+    marginBottom: '10px !important',
```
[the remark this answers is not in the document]
rajshekhar23: @xantorres  !important needed to add margin-bottom because SearchBar component is having !important for margin 0px with !important.

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,212 @@
...
+    setChecked([])
+  }
+
+  const handleCheckedLeft = () => {
+    setLeft(left.concat(rightChecked))
+    setRight(right.filter(comparer(rightChecked)))
+    setChecked([])
+  }
+
+  const renderList = (items, listName) => (
```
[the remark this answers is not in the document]
rajshekhar23: @xantorres used useCallback and wrapped complete function inside useCallback.

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+  useEffect(() => {
+    setRight(
+      rightClusterNodes.filter((value) =>
+        value.name.toLowerCase().includes(searchRight.toLowerCase()),
+      ),
+    )
+  }, [searchRight])
+
+  const handleLeftSearchChange = (search) => setSearchLeft(search)
```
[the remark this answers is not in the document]
rajshekhar23: @xantorres it is outside render body itself.

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,212 @@
...
+    if (!searchLeft) {
+      leftClusterNodes = []
+      leftClusterNodes = uniqueList(leftClusterNodes.concat(left))
+    }
+
+    setWizardContext({ batchUpgradeNodes: right })
+  }, [right, left])
+
+  useEffect(() => {
+    setLeft(
```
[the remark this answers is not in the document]
rajshekhar23: @xantorres I tried but its not working. not getting how to use useMemo here at useEffect.

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,212 @@
...
+  return function(current) {
+    return (
+      otherArray.filter(function(other) {
+        return other.name === current.name
+      }).length === 0
+    )
+  }
+}
+
+function uniqueList(list) {
```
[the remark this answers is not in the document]
rajshekhar23: @xantorres updated with ramda function uniqBy

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @xantorres  removed.

src/app/plugins/kubernetes/components/infrastructure/clusters/helpers.ts
rajshekhar23: @xantorres removed and added back to clusters/actions.js file.

src/app/plugins/kubernetes/components/monitoring/AlarmOverviewClusters.tsx
rajshekhar23: @xantorres I changed this file because i was getting some typescript error. I undo only the changes which i updated.

yarn.lock
rajshekhar23: @xantorres i installed all the packages again. May be that time it got updated. I removed changes from this file.

src/app/core/elements/transfer-list.tsx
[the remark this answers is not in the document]
xantorres: This whole `TransferList` is a render function, please declare as much stuff outside of this function as possible.

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/advanced-batch-upgrade.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+const useStyles = makeStyles((theme) => ({
+  text: {
+    marginBottom: theme.spacing(1),
+  },
+}))
+
+const AdvancedBatchUpgradeField = ({ wizardContext, setWizardContext }) => {
+  return (
+    <>
```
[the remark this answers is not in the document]
rajshekhar23: @xantorres Removed from all files wherever added.

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @xantorres removed.

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/advanced-batch-upgrade.tsx
rajshekhar23: @xantorres Updated with useMemo.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterBatchUpgradeDialog.tsx
rajshekhar23: @xantorres Updated with useMemo

src/app/core/elements/transfer-list.tsx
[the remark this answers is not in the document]
jruddell: Dont put a margin on the search box as it is. Use the parent component to define the layout. if you need help with this ill do it with you on a call

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,203 @@
...
+      newChecked.push(value)
+    } else {
+      newChecked.splice(currentIndex, 1)
+    }
+
+    setChecked(newChecked)
+  }
+
+  const handleCheckedRight = () => {
+    setRight(uniqBy(prop('uuid') as any, right.concat(leftChecked)))
```
jruddell, line 106: you shouldnt need to cast to any type here, lets hop on a call and go over the typing issues you have an ill help resolve

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+  const handleCheckedLeft = () => {
+    setLeft(uniqBy(prop('uuid') as any, left.concat(rightChecked)))
+    setRight(right.filter(comparer(rightChecked)))
+    setChecked([])
+  }
+
+  const filteredRightNodes = useMemo(() => {
+    return searchRight
+      ? right.filter((value) => value.name.toLowerCase().includes(searchRight.toLowerCase()))
```
jruddell, line 119: `value?.name?.toLowerCase()?.includes` if name isn't a string this will throw an exception.

src/app/core/elements/transfer-list.tsx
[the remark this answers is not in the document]
rajshekhar23: @jruddell applied as mentioned using parent component.

src/app/core/elements/transfer-list.tsx
rajshekhar23: @jruddell added as mentioned

src/app/core/elements/transfer-list.tsx
rajshekhar23: @jruddell I just removed 'as any' and its working fine.

src/app/core/elements/transfer-list.tsx
```diff
@@ -18,7 +18,9 @@ const useStyles = makeStyles((theme) => ({
     minHeight: '300px',
   },
   search: {
-    marginBottom: '10px !important',
+    '& .MuiOutlinedInput-adornedEnd': {
```
jruddell, line 21: what I meant by letting the parent define layout is to use something like grid and to remove the margin. If you can use css grid or flexbox to handle layout / placement (like centering with align-items and justify-content instead of a margin to do that)

src/api-client/Qbert.ts
```diff
@@ -392,16 +392,17 @@ class Qbert extends ApiService {
     })
   }
 
-  upgradeCluster = async (clusterId, type) => {
-    const url = createUrlWithQueryString(`/clusters/${clusterId}/upgrade`, { type })
-    return this.client.basicPost({
+  upgradeClusterNodes = async (clusterId, type, body = null) => {
+    const url = createUrlWithQueryString(`/clusters/${clusterId}/upgrade?type=${type}`)
+    const upgradeClusterNodesOptions = {
```
jruddell, line 397: this doesn't have any purpose as a variable when you can just pass it inline. Since thats the pattern lets stick with that

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,201 @@
...
+  },
+  listItem: {
+    borderBottom: '1px solid #ddd',
+  },
+  text: {
+    marginBottom: theme.spacing(1),
+  },
+}))
+
+function comparer(otherArray) {
```
jruddell, line 49: We generally default to using arrow syntax especially when they return functions. The name `comparer` is pretty weird it would be better to have a more explicit name about what the function is doing.

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,201 @@
...
+  return function(current) {
+    return (
+      otherArray.filter(function(other) {
+        return other.name === current.name
```
jruddell, line 53: why compare on name? name isn't a unique field for a node and can change. I would stick with uuid for your equality comparison

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,201 @@
...
+    </ListItem>
+  )
+}
+
+export default function TransferList({ clusterNodes, setWizardContext }) {
+  const classes = useStyles({})
+  const [checked, setChecked] = React.useState([])
+  const [searchLeft, setSearchLeft] = React.useState('')
+  const [searchRight, setSearchRight] = React.useState('')
+  const [left, setLeft] = React.useState(clusterNodes)
```
jruddell, line 89: Maybe we should think about the data structure here some. If there are 7k nodes managing these data sets and iterating over them all is going to be a hit to performance. Maybe instead we should use an object, key the elements by uuid?

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterBatchUpgradeDialog.tsx
```diff
@@ -0,0 +1,117 @@
...
+export const getNodeVersionToCompare = (cluster) => {
+  return cluster.minorUpgradeRoleVersion
+    ? cluster.minorUpgradeRoleVersion
+    : cluster.patchUpgradeRoleVersion
+}
+
+export const filterInProgressUpgrade = (targetVersion) => (node) =>
+  node.actualKubeRoleVersion === targetVersion
+
+export const filterInInitialUpgrade = (targetVersion) => (node) =>
```
jruddell, line 16: what is the point of these two helper methods? they do exactly the same thing

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterBatchUpgradeDialog.tsx
```diff
@@ -0,0 +1,117 @@
...
+  handleClose,
+  confirmBatchUpgrade,
+  cluster,
+}) => {
+  const classes = useStyles({})
+  const nodeVersionToCompare = getNodeVersionToCompare(cluster)
+
+  // partion node list based upgradingTo flag ( in progress / intial )
+  const [upgradedNodes, toBeUpgradedNodes] = useMemo(() => {
+    return partition(
```
jruddell, line 63: there is a bug here, i dont see you actually using the `upgradingTo` property on the cluster when comparing versions. if a user selects a version how is that handled here?

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterBatchUpgradeDialog.tsx
```diff
@@ -0,0 +1,117 @@
...
+  listNodes: {
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'start',
+    borderBottom: `1px solid ${theme.palette.grey[300]}`,
+  },
+}))
+
+const ClusterBatchUpgradeDialog = ({
+  wizardContext,
```
jruddell, line 53: why is this dialog accepting the wizard context? its a dialog, not attached or related to the wizard context. You should only pass what the component needs to use instead of giving it the whole wizard context.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.tsx
```diff
@@ -323,8 +331,8 @@ export const options = {
     {
       cond: both(isAdmin, canUpgradeCluster),
       icon: 'level-up',
-      label: 'Upgrade Cluster',
-      dialog: ClusterUpgradeDialog,
+      label: 'Upgrade Clusters',
+      routeTo: (rows) => `/ui/kubernetes/infrastructure/clusters/${rows[0].uuid}/upgrade`,
```
jruddell, line 335: we have a `routes` file that defines all our routes. please use this instead of hardcoding the string. 

for instance a few lines up from here ~line 293, thats what it is for me you have this. `editUrl: (_, id) => routes.cluster.edit.path({ id }),`. Lets reuse this pattern as it makes changing routes / refactoring much easier

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,233 @@
...
+  sequentialClusterUpgrade: false,
+  percentageClusterUpgrade: false,
+  advancedBatchUpgrade: false,
+}
+
+export const handleSetUpgradeStrategy = (field: any, id: any) => {
+  return {
+    ...togglersMapping,
+    [id]: field,
+    field,
```
jruddell, line 63: what is the purpose of putting `field` here on this object? isn't field a boolean?

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,233 @@
...
+const UpgradeClusterPage = () => {
+  const [clusters] = useDataLoader(clusterActions.list)
+  const { match, history } = useReactRouter()
+  const classes = useStyles({})
+  const cluster = clusters.find((x) => x.uuid === match.params.id)
+  const defaultState = getUpgradeTarget(cluster)
+  cluster.selectedUpgradeVersion =
+    defaultState === 'minor' ? cluster.minorUpgradeRoleVersion : cluster.patchUpgradeRoleVersion
+
+  cluster.upgradeType = defaultState
```
jruddell, line 76: why are you mutating a record like this? whats the purpose of it?

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,233 @@
...
+    handleClose()
+  }
+
+  // @ts-ignore
+  const [upgradeClusterNodes, upgradingClusterNodes] = useDataUpdater(
+    (clusterActions as any).upgradeClusterNodes,
+    onComplete,
+  )
+
+  const wizardMetaFormattedNames = {
```
jruddell, line 94: this is a constant, define it outside of the component

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,233 @@
...
+    name: 'Cluster',
+    version: 'Version',
+    masterNodes: 'Master Nodes',
+    workerNodes: 'Worker Nodes',
+    networkPlugin: 'CNI',
+  }
+
+  const wizardMetaCalloutFields = ['name', 'version', 'masterNodes', 'workerNodes', 'networkPlugin']
```
jruddell, line 102: this is a constant, define it outside of the component. Also since it's directly related to the formatted names variable it would make more sense to change this to `const wizardMetaCalloutFields = Object.keys(wizardMetaFormattedNames)`

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+  const handleSubmit = useCallback(
+    (wizardContext) => {
+      if (wizardContext.percentageClusterUpgrade) {
+        upgradeClusterNodes(wizardContext)
+        onComplete()
+      }
+
+      if (wizardContext.sequentialClusterUpgrade) {
```
jruddell, line 111: this code is bloated, both if checks for `percentageClusterUpgrade` and `sequentialClusterUpgrade` do the same thing.

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,233 @@
...
+  }
+}
+
+const UpgradeClusterPage = () => {
+  const [clusters] = useDataLoader(clusterActions.list)
+  const { match, history } = useReactRouter()
+  const classes = useStyles({})
+  const cluster = clusters.find((x) => x.uuid === match.params.id)
+  const defaultState = getUpgradeTarget(cluster)
+  cluster.selectedUpgradeVersion =
```
jruddell, line 73: why are you mutating a record like this? whats the purpose of it?

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,233 @@
...
+        handleOpen()
+      }
+    },
+    [upgradeType],
+  )
+
+  const handleChange = useCallback((event, setWizardContext) => {
+    const upgradeChoice = event.target.value
+    setUpgradeType(upgradeChoice)
+    const selectedUpgradeVersion =
```
jruddell, line 126: im seeing this code repeated multiple times. This also looks like its error prone. Just because a cluster can do a minor upgrade doesn't mean the user has chosen that option.

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,233 @@
...
+          context={cluster}
+          submitLabel="Upgrade Now"
+          hideAllButtons
+        >
+          {({ wizardContext, setWizardContext, onNext, handleNext }) => {
+            return (
+              <>
+                <WizardMeta
+                  fields={wizardContext}
+                  icon={<CloudProviderCard active type={cluster.cloudProviderType} />}
```
jruddell, line 152: a cluster's cloud provider type doesn't directly map to an icon. bareos clusters are all listed as type `local` except the icons want to know if a `local` provider is a VM or PM (virtual vs physical). You need to determine if this is one of those two types when the cloud provider is of type `local`

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -143,15 +144,12 @@ export const clusterActions = createCRUDActions(ActionDataKeys.Clusters, {
...
-
-      // Multiple fields change on cluster upgrade, best to reload the entity to get updated values
-      // Ideally the upgrade cluster call would return the updated entity
-      const updatedCluster = await qbert.getClusterDetails(cluster.uuid)
-      return adjustWith(propEq('uuid', cluster.uuid), mergeLeft(updatedCluster), prevItems)
+    upgradeClusterNodes: async (formData, prevItems) => {
+      Bugsnag.leaveBreadcrumb('Attempting to upgrade cluster', {
+        clusterId: formData.uuid,
+      })
+      const updatedCluster = createBatchUpgradeClusterPayload(formData)
```
jruddell, line 151: this is going to throw an exception when it goes through all the way, have you tested this? `createBatchUpgradeClusterPayload` is an async function that does an API call, if you dont await for it then you are going to store the pending promise as the cluster record

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/advanced-batch-upgrade.tsx
```diff
@@ -0,0 +1,78 @@
...
+)
+
+export const AdvancedBatchUpgradeAddonField = ({ values }) => {
+  const {
+    wizardContext,
+    setWizardContext,
+  }: { wizardContext: any; setWizardContext: any } = useContext(WizardContext) as any
+  const classes = useStyles()
+  const nodeVersionToCompare = getNodeVersionToCompare(wizardContext)
+  const [upgradedNodes, toBeUpgradedNodes] = useMemo(() => {
```
jruddell, line 41: this is another duplicated piece of code that also isn't taking into account and or even using the `upgradingTo` value to compare versions. Have you tested this?

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,201 @@
...
+          disableRipple
+          inputProps={{ 'aria-labelledby': labelId }}
+        />
+      </ListItemIcon>
+      <ListItemText id={labelId} primary={`${value.name}`} />
+    </ListItem>
+  )
+}
+
+export default function TransferList({ clusterNodes, setWizardContext }) {
```
jruddell, line 84: why does transferList take `setWizardContext` and `clusterNodes` as props? This is a generic component, it should just take a list of items and an updater that are generic in name as its a common reusable element

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/percentage-cluster-upgrade.tsx
```diff
@@ -0,0 +1,64 @@
...
+    info="Percentage: The specified percent of Worker node will be upgraded in parallel. 10% would upgrade 10 nodes of 100 node cluster
+       in parallel, then the next 10."
+  />
+)
+
+export const PercentageClusterUpgradeAddonField = ({ values }) => {
+  const {
+    wizardContext,
+    setWizardContext,
+  }: { wizardContext: any; setWizardContext: any } = useContext(WizardContext) as any
```
jruddell, line 30: why do you need to pull in wizardcontext here? the values should have everything on it you'd need

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/advanced-batch-upgrade.tsx
```diff
@@ -0,0 +1,78 @@
...
+    value={wizardContext?.advancedBatchUpgrade}
+    info="Advanced Batch Upgrade: The specified worked nodes will be upgraded in parallel. Worker nodes can be manually upgraded in distinct batches."
+  />
+)
+
+export const AdvancedBatchUpgradeAddonField = ({ values }) => {
+  const {
+    wizardContext,
+    setWizardContext,
+  }: { wizardContext: any; setWizardContext: any } = useContext(WizardContext) as any
```
jruddell, line 38: why do you need to pull in wizardcontext here? the values should have everything on it you'd need

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/sequential-cluster-upgrade.tsx
```diff
@@ -0,0 +1,23 @@
...
+    infoPlacement="right-end"
+    onChange={(value) =>
+      setWizardContext(handleSetUpgradeStrategy(value, 'sequentialClusterUpgrade'))
+    }
+    value={wizardContext?.sequentialClusterUpgrade}
+    info="List: The specified worker nodes will be upgraded in parallel."
+  />
+)
+
+export const SequentialClusterUpgradeAddonField = () => {
```
jruddell, line 19: dont define this component if it doesn't do anything. in the addon manager, you can specify null instead of a component that renders null

src/app/plugins/kubernetes/components/infrastructure/clusters/helpers.ts
```diff
@@ -364,3 +365,17 @@ const createGenericCluster = async (body, data) => {
   // We need to perform a GET afterwards and return that to add to the cache.
   return qbert.getClusterDetails(uuid)
 }
+
+export const createBatchUpgradeClusterPayload = async (data) => {
```
jruddell, line 369: the name of this is misleading. If it is generating the payload then thats what it should do. if its going to request from the API then it should be named in a way that you can understand thats what its doing. when i first read the name of this function (when you were using it) i thought it only generated a payload to send to the API

jruddell (changes requested):
Do another pass over the code, fix up the issues and please make sure you test this.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterBatchUpgradeDialog.tsx
rajshekhar23: @jruddell updated.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterBatchUpgradeDialog.tsx
rajshekhar23: @jruddell Updated.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterBatchUpgradeDialog.tsx
rajshekhar23: @jruddell I am setting  wizardContext as "batchUpgradeNodes" ( selected nodes to upgrade) in transfer-list and It will show same batchUpgradeNodes list from wizardContext on ClusterBatchUpgradeDialog.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClustersListPage.tsx
rajshekhar23: @jruddell Updated same as mentioned.

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @jruddell Earlier I thought set the field name to wizardContext to compare later. forgot to remove the field key from **handleSetUpgradeStrategy** function.

src/app/core/elements/transfer-list.tsx
rajshekhar23: @jruddell Updated using display: grid

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @jruddell In ClusterBatchUpgradeDialog filtering upgraded and non-upgraded nodes we need to pass upgradingTo value for the comparison which we get it from cluster object.

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @jruddell Updated added outside of the component

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @jruddell Right. Updated same

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @jruddell Right. Updated as **if (wizardContext.percentageClusterUpgrade || wizardContext.sequentialClusterUpgrade) {**

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @jruddell Removed.

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @jruddell Updated handleChange function and thats why i was added a line 
**cluster.upgradeType = defaultState**
 defaultSelection of upgradeType if only one option available to upgrade on the form.

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
rajshekhar23: @jruddell Right. Added await while calling createBatchUpgradeClusterPayload function.

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/advanced-batch-upgrade.tsx
rajshekhar23: @jruddell. Updated and tested

src/app/core/elements/transfer-list.tsx
rajshekhar23: Right @jruddell removed setWizardContext and added callback to send selected nodes values back to component.

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/sequential-cluster-upgrade.tsx
rajshekhar23: @jruddell Removed and added null.

src/app/plugins/kubernetes/components/infrastructure/clusters/helpers.ts
rajshekhar23: @jruddell Updated function name as **upgradeNodesCluster**

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/advanced-batch-upgrade.tsx
rajshekhar23: @jruddell values does not have **batchUpgradePercent** and **batchUpgradeNodes**. These values are set by **setWizardContext** callback and are getting stored in wizardContext only not in **values** prop.

Not sure why values is not getting updated once i set value using setWizardContext function.

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/percentage-cluster-upgrade.tsx
rajshekhar23: @jruddell batchUpgradePercent is set using setWizardContext which is getting added into values prop of the component.

src/api-client/Qbert.ts
rajshekhar23: @jruddell Updated.

src/app/core/elements/transfer-list.tsx
rajshekhar23: @jruddell  Updated and changed function name.

src/app/core/elements/transfer-list.tsx
rajshekhar23: @jruddell  Updated

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @jruddell Resolved and able to see VM/PM icon.

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,198 @@
...
+      }).length === 0
+    )
+  }
+}
+
+const NodeItem = ({ checked, value, handleToggle, nodeType }) => {
+  const classes = useStyles({})
+  const labelId = `transfer-list-item-${nodeType}-${value.uuid}-label`
+  return (
+    <ListItem
```
jruddell, line 60: try to use regular html elements here as material ui elements have a lot of additional stying like ripples that cause performance issues

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,144 @@
...
+  left: { [key: string]: T }
+  right: { [key: string]: T }
+}
+interface Props<T> {
+  list: T[]
+  onUpdate: (items: T[]) => void
+}
+
+enum ItemTypes {
+  LEFT = 'left',
```
jruddell, line 58: the naming convention we use for enums is 
```
enum ItemTypes {
  Left = 'left',
  Right = 'right'
}
```

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const [left, setLeft] = React.useState<CheckedItems<T>[ItemTypes.LEFT]>(
+    indexBy(prop('uuid'), list),
+  )
+  const [right, setRight] = React.useState<CheckedItems<T>[ItemTypes.RIGHT]>({})
+
+  const handleToggle = (item, itemType) => {
+    // convert input to array
+    item = Array.isArray(item) ? item : [...[], item]
+    const newChecked = { ...checked }
+    item.forEach((item) => {
```
jruddell, line 79: dont reuse the same variable name. its a param to the parent function.

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterBatchUpgradeDialog.tsx
```diff
@@ -0,0 +1,114 @@
...
+  handleClose,
+  confirmBatchUpgrade,
+  cluster,
+}) => {
+  const classes = useStyles({})
+  const nodeVersionToCompare = getNodeVersionToCompare(cluster)
```
jruddell, line 56: what about the node version a user is selecting?

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const classes = useStyles({})
+  const [checked, setChecked] = React.useState<CheckedItems<T>>(defaultCheckedState)
+  const [left, setLeft] = React.useState<CheckedItems<T>[ItemTypes.LEFT]>(
+    indexBy(prop('uuid'), list),
+  )
+  const [right, setRight] = React.useState<CheckedItems<T>[ItemTypes.RIGHT]>({})
+
+  const handleToggle = (item, itemType) => {
+    // convert input to array
+    item = Array.isArray(item) ? item : [...[], item]
```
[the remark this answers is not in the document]
rajshekhar23: @jruddell  [...[], newItem]

trying to convert single newItem into array if the type if newItem is not Array.

src/app/core/elements/transfer-list.tsx
rajshekhar23: @jruddell Updated

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterBatchUpgradeDialog.tsx
rajshekhar23: @jruddell its working by the same logic getting version based on two values as cluster.minorUpgradeRoleVersion and cluster.patchUpgradeRoleVersion

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,234 @@
...
+          {({ wizardContext, setWizardContext, onNext, handleNext }) => {
+            return (
+              <>
+                <WizardMeta
+                  fields={wizardContext}
+                  icon={
+                    <CloudProviderCard
+                      active
+                      type={
+                        clusterType.includes('virtual')
```
jruddell, line 146: this logic should be in a helper function

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,234 @@
...
+                        {cluster.upgradingTo ? (
+                          <FormFieldCard title="Batch Upgrade in progress">
+                            <Text className={classes.text} variant={'body2'}>
+                              cluster upgrade in progress. The selected worker nodes will be
+                              upgraded to: {cluster.upgradingTo}
+                            </Text>
+                          </FormFieldCard>
+                        ) : (
+                          <FormFieldCard title="Select the version to Upgrade the Cluster">
+                            <RadioGroup
```
jruddell, line 181: if im not mistaken you can supply an ID to the RadioGroup component and use it like the other elements on the wizard. I'm not sure why you are managing this field separately.

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -0,0 +1,234 @@
...
+const UpgradeClusterPage = () => {
+  const [clusters] = useDataLoader(clusterActions.list)
+  const { match, history } = useReactRouter()
+  const classes = useStyles({})
+  const cluster = clusters.find((x) => x.uuid === match.params.id)
+  const defaultState = getUpgradeTarget(cluster)
+  const clusterType = cluster.nodes[0]?.combined?.resmgr?.info?.cpu_info['virtual/physical']
+  cluster.upgradeType = defaultState
+  const [upgradeType, setUpgradeType] = useState(defaultState)
+  const [showModal, setModal] = useState(false)
```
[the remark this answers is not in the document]
rajshekhar23: @jruddell Updated variable for Modal enabling as '**isClusterBatchUpgradeDialogEnabled**'

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/advanced-batch-upgrade.tsx
```diff
@@ -0,0 +1,77 @@
...
+    info="Advanced Batch Upgrade: The specified worked nodes will be upgraded in parallel. Worker nodes can be manually upgraded in distinct batches."
+  />
+)
+
+export const AdvancedBatchUpgradeAddonField = ({ values }) => {
+  const { setWizardContext }: { wizardContext: any; setWizardContext: any } = useContext(
+    WizardContext,
+  ) as any
+  const classes = useStyles()
+  const nodeVersionToCompare = getNodeVersionToCompare(values)
```
jruddell, line 36: all of this version comparison stuff is going to be error prone. Why not just set the current version to be used in the wizard? that current version can be `cluster.upgradingTo`, the version the user selects... etc. And then these components are just consuming the `currentUpgradeVersion` (or whatever you'd like to call it)

src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterBatchUpgradeDialog.tsx
jruddell: until a user chooses a different option?

rajshekhar23: @jruddell Updated.

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @jruddell Added in helper file function called **fetchIconByClusterType**

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @jruddell Updated as discussed using radio-fields.tsx file.

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/advanced-batch-upgrade.tsx
rajshekhar23: @jruddell  Updated

src/app/core/elements/transfer-list.tsx
```diff
@@ -72,17 +72,21 @@ export default function TransferList<T extends { uuid: string }>({ list, onUpdat
...
     const newChecked = { ...checked }
-    items.forEach((item) => {
-      if (newChecked[itemType][item.uuid]) {
-        delete newChecked[itemType][item.uuid]
-      } else {
-        newChecked[itemType][item.uuid] = true
-      }
-    })
+    if (items.length === 0) {
+      Object.keys(newChecked[itemType]).forEach((key) => delete newChecked[itemType][key])
```
jruddell, line 80: what is the purpose of this?

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -80,13 +81,17 @@ const UpgradeClusterPage = () => {
   const classes = useStyles({})
   const cluster = clusters.find((x) => x.uuid === match.params.id)
   const defaultState = getUpgradeTarget(cluster)
-  const clusterType = cluster.nodes[0]?.combined?.resmgr?.info?.cpu_info['virtual/physical']
-  cluster.upgradeType = defaultState
-  const [upgradeType, setUpgradeType] = useState(defaultState)
-  const [showModal, setModal] = useState(false)
+  const clusterType = cluster?.nodes[0]?.combined?.resmgr?.info?.cpu_info['virtual/physical']
+  if (cluster?.upgradingTo)
+    cluster.upgradeType = { type: defaultState, version: cluster?.upgradingTo }
```
jruddell, line 86: why are you mutating the cluster object again? I've gone over this already, don't mutate objects like this.

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -169,35 +182,22 @@ const UpgradeClusterPage = () => {
...
-                                  control={<Radio color="primary" />}
-                                  label={`Patch - ${cluster.patchUpgradeRoleVersion}`}
-                                />
-                              )}
-                            </RadioGroup>
+                            <RadioFields
+                              formControlLabelClasses={{ label: classes.label }}
+                              id="upgradeType"
+                              options={possibleVersions}
+                              value={wizardContext?.upgradeType}
```
jruddell, line 198: the upgrade type you are defining on the cluster is not what this component expects. it should be the string version. using an object as the value can be problematic because you then are relying on a referrential check of an object. if its cloned but still the same "value" they will not appear checked. you should instead flatten out the options here to be both stored on the wizard.  `onChange({ upgradeVersion: value, upgradeType: UpgradeTypes.Patch})`

src/app/plugins/kubernetes/components/infrastructure/clusters/helpers.ts
```diff
@@ -14,6 +14,14 @@ import { CalicoDetectionTypes } from './form-components/calico-network-fields'
 import { ClusterCreateTypes } from './model'
 import { awsClusterTracking, azureClusterTracking, bareOSClusterTracking } from './tracking'
 
+export const fetchIconByClusterType = (clusterType, cluster) => {
+  return clusterType?.includes('virtual')
```
jruddell, line 18: the point of moving this to a function was to clean up the ternary hell you have currently. Please change this to use  if statement checks for virtual and physical

src/app/core/elements/transfer-list.tsx
rajshekhar23: @jruddell onSelectAll checkbox event if return value is empty, then all the checked keys of [left|right] side need to be deleted.

src/app/plugins/kubernetes/components/infrastructure/clusters/helpers.ts
rajshekhar23: @jruddell  updated

src/app/core/elements/transfer-list.tsx
jruddell: so then just make it an empty object? `newChecked[itemType] = {}`

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @jruddell to flatten the object, RadioFields component is not returning object on handleChange. it is just value.

src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
rajshekhar23: @jruddell Updated

src/app/core/elements/transfer-list.tsx
rajshekhar23: @jruddell Updated

src/app/core/components/validatedForm/ValidatedForm.js
```diff
@@ -255,7 +255,11 @@ class ValidatedForm extends PureComponent {
           {withAddonManager ? (
             <ClusterAddonManager>
               {contents}
-              <AddonDetailCards values={values} />
+              <AddonDetailCards
```
xantorres, line 258: This looks like a bad design to me, why is ValidatedForm even aware of ClusterAddonManager, and why does it have to explicitly pass anything to it? 
I think it should be the other way around, ClusterAddonManager rendering a ValidatedForm, and using a `withFormContext` HoC or a `useContext` hook to avoid polluting this already complicated component with external knowledge or responsibilities.

src/app/core/elements/transfer-list.tsx
```diff
@@ -0,0 +1,148 @@
...
+        }
+      })
+    }
+    setChecked(newChecked)
+  }
+
+  const handleMoveCheckedFromOneSide = (from, to) => {
+    const itemsBySide = {
+      right: { ...right },
+      left: { ...left },
```
xantorres, line 96: why not just `const itemsBySide = { right, left }`?

src/app/core/components/validatedForm/ValidatedForm.js
jruddell: hmm, the addon manager was meant to be a more simple form component mapping to rendering a detail card with more form fields. this is required to be inside the validated form as its rendering form elements. When we discussed this last I think the conclusion was it made the most sense to have this be attached via validated form.

jruddell (approved):

PR #1084, opened 2021-06-11
UX-967/New user onboarding

knnguy:
**New User Onboarding Requirements:**
- Must be on a DDU
- For free-tier users only

**What will trigger the onboarding flow?**
- If `isOnboarded === false` in the preference store
- If `isOnboarded` is not defined or not set in the preference store AND the user has 0 clusters

**How is isOnboarded flag stored in the user preference store?**
```
{
  featureFlags: { isOnboarded: false }
} 
```

- When the app first loads, if the `isOnboarded` flag is not defined or not set for existing users, I will set it to true if the number of clusters > 0, false if number of clusters === 0.  _**(Not sure if this logic is right. What if an existing user just doesn't have any clusters)**_

- Once the user has successfully created a cluster through the onboarding flow (step 3), I will set the `isOnboarded` flag to true

Screen Shot 2021-06-09 at 3 59 32 PM

Screen Shot 2021-06-09 at 4 21 25 PM

Screen Shot 2021-06-09 at 4 21 38 PM

Screen Shot 2021-06-09 at 4 30 44 PM

Screen Shot 2021-06-11 at 3 01 59 PM

Screen Shot 2021-06-09 at 4 26 25 PM


src/app/plugins/kubernetes/components/onboarding/onboarding-page.tsx
```diff
@@ -0,0 +1,251 @@
...
+    }
+  }, [history, clusterId, clusterChoice])
+
+  const handleDeploymentCardClick = (setWizardContext, handleNext, setActiveStep) => (
+    type: ClusterChoice,
+  ) => {
+    setClusterChoice(type)
+    setWizardContext({ clusterChoice: type })
+    if (type === 'bareOs') {
+      setActiveStep('step3', 2) // Step num is 2 because the step count starts at 0
```
jruddell, line 116: im not a fan of how this reads. Can you add more explicit names for these steps? maybe use an enum, but make this understandable, what step is step3?

jruddell (approved):

PR #1088, opened 2021-06-16
UX-504/suspend + resume with a new watchdog service

jruddell:
Setup a watchdog service that runs at a 1s interval, you can register a handler to the watchdog to run either once or on a frequency

Re-arranged the avatar and inbox icons, avatar should be all the way to the right 
image

_________
 
Moved session expiration check to run on the watchdog as well as a new system health check, where we load the health data from suspend resume service (sentinel) and register a notification when in an unhealthy state
image
 
_________
 
new dropdown when clicking on the inbox
image
 
_________
 
Clicking on the top element in the dropdown (management plane health), takes you to a status page for the DDU's health
image


knnguy (approved):

src/app/plugins/account/components/system-status/index.tsx
```diff
@@ -0,0 +1,67 @@
+// Libs
```
xantorres, line 1: Not a fan of naming files as `index.tsx` it makes it very hard to know what the file actually is when having multiple tabs opened in the IDE

src/app/plugins/account/components/system-status/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { ClientState, clientStoreKey } from 'core/client/clientReducers'
+
+// Components
+import ServiceCard from './service-card'
+import Summary from './summary'
+
+export default function AccountStatus() {
+  const classes = useStyles()
+  const { systemStatus } = useSelector<RootState, ClientState>(prop(clientStoreKey))
+  const elems = Object.entries(systemStatus?.serviceDetails).map(([name, details]) => (
```
xantorres, line 19: if `serviceDetails` is nil `Object.entries` will most likely break the app

src/app/core/watchdog/session-timeout.ts
```diff
@@ -0,0 +1,14 @@
+import { MessageTypes } from 'core/components/notifications/model'
+import { sessionActions } from 'core/session/sessionReducers'
+import moment from 'moment'
+
+const sessionTimeoutCheck = (session, showToast) => (dispatch) => {
+  // Check if session has expired
+  const { expiresAt } = session
+  const sessionExpired = moment().isAfter(expiresAt)
+  if (sessionExpired) {
+    dispatch(sessionActions.destroySession())
```
xantorres, line 10: Should we clear the cache too? `dispatch(cacheActions.clearCache())`

src/app/core/watchdog/index.ts
```diff
@@ -0,0 +1,56 @@
...
+          handlerObj.handler(this.dispatch)
+          handlerObj.currentTime = 0
+        }
+      })
+    }, this.frequency)
+  }
+
+  public destroy() {
+    clearInterval(this.timer)
+    this.timer = undefined
```
xantorres, line 45: Not 100% sure here but `Watchdog.handlers = []` sounds like a good idea, just to be safe

src/app/core/watchdog/session-timeout.ts
xantorres: Probably we should destroy the Watchdog here too

src/app/bootstrap.js
```diff
@@ -33,6 +34,9 @@ if (config.apiHost === undefined) {
 ApiClient.init({ keystoneEndpoint: `${config.apiHost}/keystone` })
 const { preferenceStore } = ApiClient.getInstance()
 
+const watchdog = new Watchdog(1000, store.dispatch)
```
xantorres, line 37: This would probably be better (safer) as a singleton, like what we do with the `ApiClient`

src/app/core/containers/AppContainer.tsx
```diff
@@ -236,6 +238,7 @@ const AppContainer = () => {
     }
 
     const initialize = async () => {
+      Watchdog.register({ handler: systemHealthCheck, frequency: 1000 * 60 * 10 })
```
xantorres, line 241: Should we "unregister" this when ending the session?

xantorres (approved):

src/app/core/watchdog/session-timeout.ts
jruddell: Watchdog destroy on session expiration doesn't make sense. It's supposed to run during the login page as well

jruddell: maybe its better for destroy session to clear cache than each place call these individually? I had just done a carry over of this function into the watchdog.

src/app/plugins/account/components/system-status/index.tsx
jruddell: i generally just close tabs and use cmd + p to open a file im looking for. or tabbing through open files will show the folder structure so its pretty easy to see what files are what.
image

src/app/core/containers/AppContainer.tsx
jruddell: no, this isn't tied to a session. it is an unauthenticated route for the health of the DDU

PR #1095, opened 2021-06-24
UX-962/Move monitoring option to edit cluster page

knnguy:
### **Description:**
User can no longer toggle monitoring for each cluster in the clusters list table. The monitoring option is now moved to the edit cluster page.

Screen Shot 2021-06-23 at 5 06 05 PM

Screen Shot 2021-06-23 at 5 06 21 PM


knnguy (commented):
In the `toggleMonitoring` function that I took from `PrometheusAddOnDialog`, if monitoring is enabled, we do `localStorage.setItem(onboardingMonitoringSetup, 'true')`. If monitoring is disabled, then shouldn't we set it to false?

jruddell:
> In the `toggleMonitoring` function that I took from `PrometheusAddOnDialog`, if monitoring is enabled, we do `localStorage.setItem(onboardingMonitoringSetup, 'true')`. If monitoring is disabled, then shouldn't we set it to false?

You can remove the code where we set that now. I believe its left over from a previous onboarding setup that we dont use anymore

jruddell (approved):

PR #1101, opened 2021-06-30
Multiple fixes for RBAC Profile Creation

xantorres:


src/app/plugins/kubernetes/components/rbac/profiles/create/api-permissions-dialog.tsx
```diff
@@ -0,0 +1,276 @@
...
+    backgroundColor: theme.palette.blue['200'],
+  },
+}))
+
+// The modal is technically inside the row, so clicking anything inside
+// the modal window will cause the table row to be toggled.
+const stopPropagation = (e) => e.stopPropagation()
+
+const arr = []
+const PermissionGroup = ({ verbs, apiGroups = arr, resources = arr, resourceNames }: ApiRules) => {
```
kevinylee1991, line 135: if apiGroups and resources are both undefined, wouldn't they both be set to the same pointer here if they default to arr? Perhaps just apiGroups = [] and resources = []?

kevinylee1991 (approved):
Seems good to me, just check the one comment

src/app/plugins/kubernetes/components/rbac/profiles/create/api-permissions-dialog.tsx
xantorres: It doesn't matter, as we are using FP, all the pointers will be resetted when adding new elements to each variable. I just did this to prevent instancing of new arrays every time this small component is created.

xantorres: Changed to `[]` to avoid confusion

PR #1104, opened 2021-06-30
UX-986/updated nodes list page to span entire screen

rajshekhar23:
Screenshot 2021-07-01 at 12 57 22 AM

### DIFF
--- a/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterDetailsPage.js
+++ b/src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterDetailsPage.js
@@ -133,6 +133,10 @@ const useStyles = makeStyles((theme) => ({
   },
   tabContainer: {
     paddingTop: theme.spacing(2),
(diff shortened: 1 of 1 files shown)


src/app/plugins/kubernetes/components/infrastructure/clusters/ClusterDetailsPage.js
```diff
@@ -133,7 +133,6 @@ const useStyles = makeStyles((theme) => ({
   },
   tabContainer: {
     paddingTop: theme.spacing(2),
-    maxWidth: 1234,
```
jruddell, line 136: this max width was needed for a different screen, I believe the details one, for the card layout. Can you apply it on that one?

rajshekhar23: @jruddell Overiden max-width property by adding one more class with max-width: none.

jruddell (approved):

PR #1105, opened 2021-06-30
UX-1012/Fix onboarding wizard bugs

knnguy:
### **Fixes:**

1. After cluster creation step (step 3), the user will no longer be redirected to the dashboard page. After step 3 is complete, the user will be taken to the add co-worker step.

2. The notifications icon in the toolbar will be disabled (not clickable) when the user is in the middle of the onboarding wizard.

3. The kubeRoleVersion error will not appear anymore during cluster creation

4. Creating cloud provider with invalid credentials will not break the app anymore.

5. Once creating a cloud provider finishes, it will always take the user to the cluster creation step.

6. When initially loading the app, don't load the dashboard or the onboarding wizard until we know exactly which one to load.


src/app/core/containers/AuthenticatedContainer.tsx
```diff
@@ -349,6 +364,22 @@ const AuthenticatedContainer = () => {
     showNavBar: showNavBar,
   })
 
+  const [{ featureFlags = {} as any }, , , updateUserDefaults] = useScopedPreferences('defaults')
```
jruddell, line 367: I'm not really liking how we have to destructure a number of items to get to the updater. maybe we should think about changing the return type of `useScopedPreferences` to be an object with named params that you can destructure

src/app/core/containers/AuthenticatedContainer.tsx
```diff
@@ -349,6 +364,22 @@ const AuthenticatedContainer = () => {
...
+    currentStack === AppPlugins.Kubernetes &&
+    isDecco(features) &&
+    customerTier === CustomerTiers.Freedom
+  const showOnboarding =
+    isTargetEnv &&
+    ((featureFlags.isOnboarded === undefined && clusters?.length === 0) ||
+      (featureFlags.isOnboarded !== undefined && !featureFlags.isOnboarded))
+
+  useEffect(() => {
+    if (!isTargetEnv || (featureFlags.isOnboarded !== undefined && clusters?.length === 0)) return
```
jruddell, line 379: shouldn't we be checking for users? onboarding is only complete after a user is added or they click the skip button. If they click skip we should be setting onboarded as complete

jruddell (approved):

PR #1106, opened 2021-06-30
Added more columns and polling to rbac profiles list page. Also fixed clusterRef for profile creation request body.

kevinylee1991:
Screen Shot 2021-06-30 at 4 00 13 PM

### DIFF
--- a/src/app/core/components/PollingData.tsx
+++ b/src/app/core/components/PollingData.tsx
@@ -36,7 +36,7 @@ function PollingData<T>({
     const ts = new Date().valueOf()
     setLastFetchTs(ts)
     setLastIntervalTs(ts)
-    if (pollingCondition && !pollingCondition()) {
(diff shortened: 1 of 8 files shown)


src/app/core/components/PollingData.tsx
```diff
@@ -27,16 +28,22 @@ function PollingData<T>({
...
   textColor = 'textSecondary',
+  pollingCondition,
 }: Props<T>) {
   const [lastIntervalTs, setLastIntervalTs] = useState(new Date().valueOf())
   const [lastFetchTs, setLastFetchTs] = useState(new Date().valueOf())
   const reload = useCallback(() => {
     const ts = new Date().valueOf()
     setLastFetchTs(ts)
     setLastIntervalTs(ts)
+    if (pollingCondition && !pollingCondition()) {
```
jruddell, line 39: probably better to just have the check be a typeof for 'function'

jruddell (approved):

PR #1112, opened 2021-07-06
UX-294/Created new design for password reset page.

rajshekhar23:
Screenshot 2021-07-06 at 6 33 01 PM


jruddell (changes requested):
Instead of duplicating the exact same design on a different page, can you make a more generic component that provides this layout and you just need to pass the input fields / form as children to the component

src/app/core/public/ResetPasswordPage.tsx
```diff
@@ -209,72 +285,65 @@ const ResetPasswordPage: React.FC = () => {
...
-      <Grid container justify="center" className={classes.root}>
-        <Grid item md={5} lg={4}>
-          <Paper className={classes.paper}>
-            <img src="/ui/images/logo-color.png" className={classes.img} />
-            <Text variant="subtitle1" align="center">
-              Password Reset
-            </Text>
-            <ValidatedForm elevated={false} onSubmit={handleFormSubmit}>
-              {({ values }) => (
+    <>
```
jruddell, line 288: dont use fragments with no purpose please.

src/app/core/public/ResetPasswordPage.tsx
```diff
@@ -23,49 +23,119 @@ import {
...
 import FontAwesomeIcon from 'core/components/FontAwesomeIcon'
+import clsx from 'clsx'
+import FormPageContainer from 'core/containers/form-page-container'
 
-const useStyles = makeStyles((theme: Theme) => ({
-  root: {
-    padding: theme.spacing(8),
-    overflow: 'auto',
+const styles = (theme) => ({
+  page: {
```
jruddell, line 35: these styles need to be removed since they are for the form page container

src/app/core/public/ResetPasswordPage.tsx
```diff
@@ -209,72 +285,65 @@ const ResetPasswordPage: React.FC = () => {
...
+            </>
+          )}
+        </ValidatedForm>
+      </FormPageContainer>
+    </>
   )
 }
 
-export default ResetPasswordPage
+export default withStyles(styles as any)(ResetPasswordPage)
```
jruddell, line 349: for styling just use `useStyles` as a hook, when you wrap this with `withStyles` it can mess up typescript typings when trying to render this component.

jruddell (changes requested):
few more minor changes

jruddell (approved):

PR #1114, opened 2021-07-07
UX-1004/Resolved CSS ugliness on PodListPage page

rajshekhar23:
Screenshot 2021-07-07 at 11 52 20 AM


src/app/core/components/listTable/ListTable.js
```diff
@@ -61,6 +61,9 @@ const styles = (theme) => ({
     '&.Mui-selected': {
       backgroundColor: theme.palette.blue[100],
     },
+    '& .MuiTableCell-body': {
+      whiteSpace: 'nowrap',
```
jruddell, line 65: what happens when the screen gets small? shouldn't we use ellipsis here?

rajshekhar23: @jruddell resolved using text-overflow property: ellipsis.

src/app/plugins/kubernetes/components/pods/renderLabels.js
```diff
@@ -15,7 +15,10 @@ const useStyles = makeStyles((theme) => ({
     gridTemplateColumns: 'max-content 1fr',
     gridGap: theme.spacing(2),
     '& > span': {
-      justifySelf: 'flex-start',
```
jruddell, line 18: renderLabels is used in a number of places, have you looked at how each piece looks with these changes? For instance there is a modal that is at least 300 px wide that uses this. in that instance it wouldn't make sense to limit to 120px

rajshekhar23: @jruddell Resolved.!!

jruddell (approved):

PR #1117, opened 2021-07-08
Added rbac profile binding management wizard

kevinylee1991:
Screen Shot 2021-07-07 at 8 31 44 PM

### DIFF
--- a/src/app/plugins/kubernetes/components/rbac/profiles/deploy/deploy-rbac-profile-page.tsx
+++ b/src/app/plugins/kubernetes/components/rbac/profiles/deploy/deploy-rbac-profile-page.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo } from 'react'
+import React, { useCallback, useMemo } from 'react'
 import FormWrapper from 'core/components/FormWrapper'
 import useDataLoader from 'core/hooks/useDataLoader'
 import { routes } from 'core/utils/routes'
@@ -19,6 +19,7 @@ import { clusterActions } from 'k8s/components/infrastructure/clusters/actions'
 import { cloudProviderTypes } from 'k8s/components/infrastructure/cloudProviders/selectors'
 import { capitalizeString } from 'utils/misc'
 import useDataUpdater from 'core/hooks/useDataUpdater'
+import { customValidator, requiredValidator } from 'core/utils/fieldValidators'
 
 const useStyles = makeStyles((theme: Theme) => ({
   wizardMeta: {
@@ -65,6 +66,8 @@ const profileCalloutFields = [
   'clusterRoleBindings',
 ]
 
+const renderActiveProfile = (hasBinding) => <span>{hasBinding ? 'Yes' : 'No'}</span>
+
 const columns = [
   { id: 'name', label: 'Name' },
   { id: 'kubeRoleVersion', label: 'Version' },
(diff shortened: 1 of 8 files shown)


src/app/plugins/kubernetes/components/rbac/profiles/deploy/deploy-rbac-profile-page.tsx
```diff
@@ -113,15 +121,39 @@ const renderProfileCalloutFields = () => (fields) => {
...
   const profileName = match?.params?.name
   const [profiles, profilesLoading] = useDataLoader(rbacProfileActions.list)
   // When wireframes are finalized for how to manage active profiles,
   // add below back in and display that info in the clusters list
-  const [, , refreshProfileBindings] = useDataLoader(rbacProfileBindingsActions.list)
   const [clusters, clustersLoading] = useDataLoader(clusterActions.list)
+  const mappedClusters = useMemo(() => {
+    if (clusters.length && profiles.length) {
+      return clusters.map((cluster) => {
+        const clusterProfile = profiles.find((profile) =>
```
jruddell, line 146: going back to having more performant code, this is going through every cluster, then through every profile, then through each binding on that profile to map to a cluster. instead build a datastructure that stores a lookup to the profile by cluster uuid.

src/app/plugins/kubernetes/components/rbac/profiles/selectors.ts
```diff
@@ -5,6 +5,34 @@ import DataKeys from 'k8s/DataKeys'
 import getDataSelector from 'core/utils/getDataSelector'
 import { determineProfileResources } from './helpers'
 
+export const rbacProfileBindingsSelector = createSelector(
+  [getDataSelector(DataKeys.RbacProfileBindings)],
+  (profileBindings) => {
+    return profileBindings.map((binding) => {
```
jruddell, line 11: Array.map always returns a new array, instead wrap it with a check that returns the `emptyArr` variable so the reference check stays the same. `return arrayIfEmpty(profileBindings.map(...))`

src/app/plugins/kubernetes/components/rbac/profiles/selectors.ts
```diff
@@ -5,6 +5,34 @@ import DataKeys from 'k8s/DataKeys'
...
+  defaultParams = {
+    orderBy: 'name',
+    orderDirection: 'asc',
+  },
+) => {
+  return createSelector(
+    [rbacProfileBindingsSelector, (_, params) => mergeLeft(params, defaultParams)],
+    (rbacProfileBindings, params) => {
+      const { orderBy, orderDirection } = params
+      return pipe(createSorter({ orderBy, orderDirection }))(rbacProfileBindings)
```
jruddell, line 31: `return pipe(createSorter({ orderBy, orderDirection }), arrayIfEmpty)(rbacProfileBindings)`

src/app/plugins/kubernetes/components/rbac/profiles/selectors.ts
```diff
@@ -18,12 +46,15 @@ const determineProfileAction = (profile) => {
...
   [
     getDataSelector(DataKeys.RbacProfiles),
-    getDataSelector(DataKeys.RbacProfileBindings),
+    rbacProfileBindingsSelector,
     getDataSelector(DataKeys.Clusters),
   ],
   (rbacProfiles, profileBindings, clusters) => {
     return rbacProfiles.map((profile) => {
-      const matchingBindings = profileBindings.filter((binding) => {
+      const nonDryrunBindings = profileBindings.filter((binding) => {
```
jruddell, line 54: you can apply both pieces of logic in one filter here. 

```
const matchingBindings = profileBindings.filter((binding) => {
  const isDryRun = !!binding.spec.dryRun
  const bindingMatches = binding.spec.profileRef.split('default/')[1] === profile.metadata.name
  return !isDryRun && bindingMatches
})
```

jruddell (changes requested):
few optimizations

jruddell (approved):

PR #1118, opened 2021-07-08
UX-984/enable-multiple-selection-of-worker-nodes-to-scale

hgupta2363:
able to select multiple nodes to scale

Screenshot 2021-07-08 at 4 08 41 PM


src/app/plugins/kubernetes/components/infrastructure/clusters/bareos/ClusterHostChooser.tsx
```diff
@@ -210,7 +210,7 @@ const ClusterHostChooser: React.ComponentType<Props> = forwardRef<HTMLElement, P
       hasError,
       errorMessage,
       pollForNodes = false,
-      selection = 'single',
+      selection,
```
jruddell, line 213: dont change this, if you pass a selection then it will override a default value. But this component is used in many places and other instances most likely rely on there being a default value of type `single` selection

jruddell (commented):
Did you test that this works and the nodes are scaled as a batch?

hgupta2363:
> Did you test that this works and the nodes are scaled as a batch?

yeah that is working like you can select multiple worker nodes and after clicking on add worker those selected nodes will be added in that cluster

jruddell (approved):

PR #1121, opened 2021-07-13
UX-963, UX-658, UX-964, UX-965, UX-966/ Modified Edit Cluster page. Allow users to edit cluster addons managed by the addon manager

knnguy:
### **Clusters List Page Action Bar**
- Modified action bar order
Screen Shot 2021-07-19 at 2 17 14 PM

### **Edit Cluster Page Changes**
- Modified it to look like the cluster create page
- Put tags in a separate card
- The following fields are shown but uneditable:
    - Kubernetes Version 
    - IP  Sessions 
    - Privileged Containers 
    - Make masters Master + Worker 
    - Kubevirt
    - Network Add-on Operator 
- The following fields are editable:
   - Name 

### **Edit AWS Cluster Page**
- Added the following addons that are managed by the addon manager:
   - Autoscaler (can only be installed/uninstalled)
   - CoreDns
   - Kubernetes Dashboard 
   - Metrics Server 
Screen Shot 2021-07-19 at 1 57 30 PM

### **Edit Azure Cluster Page**
- Added the following addons that are managed by the addon manager:
   - Autoscaler (can set min/max workers)
   - CoreDns
   - Kubernetes Dashboard
   - Metrics Server
Screen Shot 2021-07-19 at 1 55 41 PM

### **Edit BareOs Cluster Page**
- Added the following addons that are managed by the addon manager:
   - MetalLb
   - CoreDns
   - Kubernetes Dashboard
   - Metrics Server 
Screen Shot 2021-07-19 at 1 53 57 PM

### **CoreDns Override Params** 
Screen Shot 2021-07-19 at 2 03 53 PM

### **Azure Autoscaling Params** 
Screen Shot 2021-07-19 at 2 08 45 PM

### **Cluster Addons section only shows if Kubernetes version is 1.20 or later**
Screen Shot 2021-07-21 at 10 06 22 AM


src/app/plugins/kubernetes/components/infrastructure/clusters/clusterAddons/helpers.ts
```diff
@@ -0,0 +1,220 @@
...
+]
+
+// Maps Addon Manager API addon param names to UI field names
+export const addonParamsApiNameMap = {
+    // CoreDns
+    dnsMemoryLimit: 'dnsMemoryLimit',
+    dnsDomain: 'dnsDomain',
+    base64EncAdditionalDnsConfig: 'base64EncAdditionalDnsConfig',
+    // MetalLb
+    MetallbIpRange: 'metallbCidr',
```
jruddell, line 46: its a little strange to have the casing of this field start with a capital letter and the others dont. Is there a reason for this?

src/app/plugins/kubernetes/components/infrastructure/clusters/clusterAddons/helpers.ts
```diff
@@ -0,0 +1,220 @@
...
+  }, [])(addonType)
+
+export const getCoreDnsRequestBody = ({
+  clusterId,
+  dnsMemoryLimit,
+  dnsDomain,
+  additionalDnsConfig,
+}) => {
+  return {
+  apiVersion: 'sunpike.platform9.com/v1alpha2',
```
jruddell, line 77: there is a lot of repetition here related to creating these request bodies. Can we make a method that defines the overall structure of this payload and then let each of these methods pass more specifics that are different?

src/app/plugins/kubernetes/components/infrastructure/clusters/clusterAddons/helpers.ts
knnguy: Not sure, but that's how it is in the docs. 

https://platform9.atlassian.net/wiki/spaces/eng/pages/2370371762/Supported+ClusterAddon+objects

Screen Shot 2021-07-20 at 10 32 42 AM

src/api-client/Qbert.ts
```diff
@@ -73,6 +75,16 @@ const normalizeImportedClusters = (apiResponse: GCluster<ImportedCluster>) => {
   return normalizedClusters
 }
 
+const normalizeClusterAddons = (apiResponse: GCluster<ClusterAddon>) => {
+  const addons = apiResponse.items
+  const normalizeClusterAddons = addons.map((addon) => ({
+    ...addon,
+    name: addon.metadata?.name,
```
jruddell, line 82: what is the need for uplifting the name and the id on each addon here? why not just do it in the selector? (I'm thinking the extra loop over all items is unnecessary)

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/coreDns.tsx
```diff
@@ -0,0 +1,61 @@
...
+    label="CoreDNS"
+    infoPlacement="right-end"
+    value={wizardContext[coreDnsFieldId]}
+    onChange={(value) => setWizardContext({ [coreDnsFieldId]: value })}
+    info="CoreDNS"
+  />
+)
+
+export const CoreDnsAddonFields = () => {
+  const { params = {}, updateParams }: { params: any; updateParams: any } = useContext(
```
jruddell, line 23: Is this addon only used in edit cluster?

src/app/plugins/kubernetes/components/infrastructure/clusters/EditClusterPage.tsx
```diff
@@ -0,0 +1,421 @@
...
+    },
+    [update, initialValues, toggleMonitoring, updateClusterAddons, params],
+  )
+
+  const updatingAddons = creatingAddon || updatingAddon || deletingAddon
+  const loadingSomething =
+    loadingClusters || loadingClusterAddons || updatingCluster || updatingAddons
+
+  return (
+    <EditClusterContext.Provider value={{ params, updateParams }}>
```
jruddell, line 367: why do we need an edit cluster context? I feel like using a wizard here would be appropriate, and would also simplify the different places where you need to check for values in both the wizard and the edit cluster contexts.

jruddell (changes requested):
Few things to clean up

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/coreDns.tsx
knnguy: Yes. It's automatically installed at cluster creation.

src/app/plugins/kubernetes/components/infrastructure/clusters/EditClusterPage.tsx
knnguy: I did try taking out `useParams` and use `wizardContext` instead,  but it doesn't work because I can't use `setWizardContext` outside of the `render() `method. 

In the code, after the existing cluster addons are loaded, I loop through the addons in a `useEffect` hook and add their params to the form using `updateParams`. I can't do that with `setWizardContext`. I can set the form value, but I can't get it in `wizardContext`

src/api-client/Qbert.ts
knnguy: The `id` is used as the `uniqueIdentifier` in `clusterAddonActions`

mayuresh108:
@knnguy Maybe this is already fixed but we don't want to show these addon override params for older clusters like 1.17, 1.18 etc. 
The presence of `ClusterAddon` object for this cluster should confirm that the underlying pf9-kube version support addon mgmt.

Also the additionalDnsConfig parameter is optional, do we want to add that in the default value so it becomes clear.

knnguy:
> @knnguy Maybe this is already fixed but we don't want to show these addon override params for older clusters like 1.17, 1.18 etc.
> The presence of `ClusterAddon` object for this cluster should confirm that the underlying pf9-kube version support addon mgmt.
> 
> Also the additionalDnsConfig parameter is optional, do we want to add that in the default value so it becomes clear.

@mayuresh108 From what I'm told, users can only edit the cluster addons if the kubernetes version is 1.20 or later. The Cluster Addons part in the edit cluster page will not appear.

src/app/plugins/kubernetes/components/infrastructure/clusters/clusterAddons/helpers.ts
jruddell: Ok then leave it as is

src/app/plugins/kubernetes/components/infrastructure/clusters/EditClusterPage.tsx
xantorres: I might be missing something but what about using the ValidatedForm context for everything? The values should have been added to the ValidatedForm context automatically anyways when adding components that are using the `withFormContext` HoC

src/app/plugins/kubernetes/components/infrastructure/clusters/EditClusterPage.tsx
```diff
@@ -346,68 +333,98 @@ const EditClusterPage = () => {
...
+              message={'Submitting form...'}
+              backUrl={routes.cluster.list.path()}
+            >
+              <Wizard
+                context={initialValues}
+                submitLabel="Update Cluster"
+                onComplete={handleSubmit}
+              >
+                {({ wizardContext, setWizardContext, onNext }) => (
+                  <WizardStep stepId="updateCluster">
```
xantorres, line 363: I don't get why we have to use the Wizard even when there's only one step

jruddell (approved):

PR #1124, opened 2021-07-16
Added drift analysis to deployment workflow and drift analytics tab

kevinylee1991:


src/app/plugins/kubernetes/components/rbac/profiles/actions.ts
```diff
@@ -115,3 +115,36 @@ export const patchRbacProfile = async (name, body) => {
   const response = await qbert.patchRbacProfile(name, body)
   return response
 }
+
+export const getDriftAnalysis = async ({ cluster, profileName }) => {
+  const clusterId = cluster[0].uuid
```
jruddell, line 120: if `cluster` is an array it should be plural. Can we just send the one cluster though instead of an array with one item in it?

src/app/plugins/kubernetes/components/rbac/profiles/deploy/deploy-rbac-profile-page.tsx
```diff
@@ -99,6 +88,12 @@ const renderProfileCalloutFields = () => (fields) => {
         <Text variant="body2">Profile Name:</Text>
         <Text variant="caption1">{fields.name}</Text>
       </div>
+      {cluster?.length && cluster[0] && (
```
jruddell, line 91: you can consolidate this with a single check `!!cluster?.[0] && (`

src/app/plugins/kubernetes/components/rbac/profiles/deploy/deploy-rbac-profile-page.tsx
```diff
@@ -121,51 +116,77 @@ const renderProfileCalloutFields = () => (fields) => {
...
-  requiredValidator,
-  customValidator((clusters) => {
-    if (!clusters.length) {
-      return false
+const DeployNowButton = ({ wizardContext }) => {
+  const classes = useStyles({})
+  const [showDialog, setShowDialog] = useState(false)
+  const [showError, setShowError] = useState(false)
+  const clickedButton = () => {
+    if (!!wizardContext?.cluster?.[0] && !wizardContext.cluster[0].hasBinding) {
```
jruddell, line 124: I feel like it would be easier to just store the cluster record on the wizardContext instead of an array with one item in it, unless I'm missing something. Can you explain the reasoning for this data model?

src/app/plugins/kubernetes/components/rbac/profiles/deploy/dry-run-dialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+  handleClose,
+  onFinish,
+  setWizardContext,
+}: Props) => {
+  const classes = useStyles({})
+  // @ts-ignore not using lastIntervalTs but this is what triggers the poll
+  const [lastIntervalTs, setLastIntervalTs] = useState(new Date().valueOf())
+  const [lastFetchTs, setLastFetchTs] = useState(new Date().valueOf())
+
+  const getUpdate = useCallback(async () => {
```
jruddell, line 50: I wrote a component that will poll at a given interval. why not just use that?

src/app/plugins/kubernetes/components/rbac/profiles/drift/cluster-role-analysis.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  const clusterRoleAnalysis = clusterRoles.map((clusterRole) => {
+    const rules = parseRoleRules(clusterRole.newValue.rules)
+    return {
+      ...clusterRole,
+      rbac: rules,
+      numApis: numApiResources(rules),
+    }
+  })
+  console.log(clusterRoleAnalysis)
```
jruddell, line 111: remove console.log

src/app/plugins/kubernetes/components/rbac/profiles/drift/drift-analysis-result.tsx
```diff
@@ -0,0 +1,116 @@
...
+  },
+}))
+
+const DriftSummary = ({ wizardContext, createFn }) => {
+  const classes = useStyles({})
+
+  const deployCluster = useCallback(() => {
+    return createFn({
+      cluster: wizardContext.cluster,
+      profileName: wizardContext.profile[0].name,
```
jruddell, line 52: if profile is an array please make it plural.

src/app/plugins/kubernetes/components/rbac/profiles/drift/drift-analysis.tsx
```diff
@@ -0,0 +1,118 @@
...
+  verticallyCenter: {
+    display: 'flex',
+    alignItems: 'center',
+  },
+  icon: {
+    marginRight: theme.spacing(1),
+  },
+}))
+
+const validJsonOrEmpty = (string) => {
```
jruddell, line 57: try to avoid using `string` as a variable name. Its technically not a protected word but some IDE's might color code that differently because it thinks its a type.

src/app/plugins/kubernetes/components/rbac/profiles/drift/drift-analysis.tsx
```diff
@@ -0,0 +1,118 @@
...
+  try {
+    const parsedValue = JSON.parse(string)
```
jruddell, line 59: just return it `return JSON.parse(inputValue)`

jruddell (changes requested):
There is a lot of duplicated code here. for instance `const opIcons = {` is defined in 4 separate files. A number of things are identical between these files. If there are going to be differences then maybe we have a helper file that defines some of the shared items, like an icon mapping, custom validators... etc.

src/app/plugins/kubernetes/components/rbac/profiles/deploy/deploy-rbac-profile-page.tsx
kevinylee1991: ListTableField component stores the values in an array -- implementation is similar to the ListTable

src/app/plugins/kubernetes/components/rbac/profiles/drift/cluster-role-analysis-dialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+  },
+}))
+
+interface Props {
+  rbac: any
+  clusterRole: any
+  onClose: () => void
+}
+
+const renderGroupResources = (group, rbac) => {
```
xantorres, line 70: Redeclare this as a react component if you want to use hooks here (such as `useStyles`)

src/app/plugins/kubernetes/components/rbac/profiles/drift/role-analysis-dialog.tsx
```diff
@@ -0,0 +1,119 @@
...
+  },
+}))
+
+interface Props {
+  rbac: any
+  role: any
+  onClose: () => void
+}
+
+const renderGroupResources = (group, rbac) => {
```
xantorres, line 67: This should be redeclared as a component

src/app/plugins/kubernetes/components/rbac/profiles/drift/role-analysis-dialog.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const classes = useStyles({})
+  const resources = keys(rbac[group])
+  return (
+    <div className={classes.resourcesList}>
```
xantorres, line 71: Add a `key` prop here

src/app/plugins/kubernetes/components/rbac/profiles/deploy/deploy-rbac-profile-page.tsx
```diff
@@ -66,31 +79,7 @@ const profileCalloutFields = [
...
-    id: 'enableProfileAgent',
-    label: 'Profile Agent',
-    render: (enabled) => (enabled ? 'Yes' : 'No'),
-  },
-]
-
-const profileAgentInstalled = (cluster) => !!cluster.enableProfileAgent
-
-const renderProfileCalloutFields = () => (fields) => {
+const renderProfileCalloutFields = (cluster) => (fields) => {
```
xantorres, line 82: Maybe `moize` here to prevent creating a new function every time

src/app/plugins/kubernetes/components/rbac/profiles/drift/cluster-role-analysis.tsx
```diff
@@ -0,0 +1,164 @@
...
+  { id: 'op', label: '', render: renderOp },
+  { id: 'name', label: 'Name' },
+  { id: 'numApis', label: 'APIs' },
+  { id: 'rbac', label: '', render: renderInspectButton },
+]
+
+const ClusterRoleAnalysis = ({ clusterRoles }) => {
+  const classes = useStyles({})
+
+  const clusterRoleAnalysis = clusterRoles.map((clusterRole) => {
```
xantorres, line 103: `useMemo` here

src/app/plugins/kubernetes/components/rbac/profiles/deploy/dry-run-dialog.tsx
kevinylee1991: Looks like using PollingData works more easily for collections, but has some difficulty with calling a custom API and custom processing

jruddell: Its just a callback function, you can define anything you want there. meaning `<PollingData reload={anyCallbackFunctionIWant}`

jruddell (approved):

PR #1125, opened 2021-07-20
UX-1073/add-one-worker-node-in-small-template-and-change-one-click-for-small-…

hgupta2363:
Screenshot 2021-07-20 at 12 29 01 PM
…template


jruddell (changes requested):
update the flavor and were good to go :)

jruddell (approved):

PR #1127, opened 2021-07-20
UX-988/Profiles - RBAC - Edit Profile

xantorres:


src/api-client/helpers.ts
```diff
@@ -72,10 +72,15 @@ interface ResponseWithData {
...
-export const normalizeResponse = <T>(response): T extends ResponseWithData ? T['data'] : T => {
+export const normalizeResponse = <T>(
+  response,
+  fixNestedData = true,
+): T extends ResponseWithData ? T['data'] : T => {
   const data = response && response.hasOwnProperty('data') ? response.data : response
   // Fix nested data.data issue
-  return (data && data.hasOwnProperty('data') ? data.data : data) || emptyArr
+  return (
+    (fixNestedData ? (data && data.hasOwnProperty('data') ? data.data : data) : data) || emptyArr
```
jruddell, line 82: nesting ternaries is something I try to avoid as its harder to read. probably better to just check for this flag separately

```
if (!fixNestedData) return data || emptyArr
return (data && data.hasOwnProperty('data') ? data.data : data) || emptyArr
```

src/app/plugins/kubernetes/components/rbac/profiles/actions.ts
```diff
@@ -49,16 +54,33 @@ export const rbacProfileActions = createCRUDActions(ActionDataKeys.RbacProfiles,
...
+        name: string
+        namespace: string
+        details: IRbacProfileDetails
+      },
+      prevItems,
+    ) => {
+      Bugsnag.leaveBreadcrumb('Attempting to update rbac profile details')
+      trackEvent('Update RBAC Profile details', details)
+      const response = await qbert.putRbacProfileDetails(name, namespace, details)
+      console.log(response)
```
jruddell, line 80: console.log

src/app/plugins/kubernetes/components/rbac/profiles/edit/add-role-dialog.tsx
```diff
@@ -0,0 +1,341 @@
...
+const stopPropagation = (e) => e.stopPropagation()
+
+interface IApiPermissionsDialogProps {
+  onClose: () => void
+  onSubmit: (role: IRbacRole) => void
+  open: boolean
+  clusterId: string
+}
+
+const getRuleUniqueKey = (rule: IRbacRoleRule) => {
```
jruddell, line 115: probably better to have this as a helper so its not redefined

jruddell (approved):
theres a bit of duplicated code in each view file. it probably makes sense to move those to a helper file and reuse

src/app/plugins/kubernetes/components/rbac/profiles/edit/add-role-dialog.tsx
xantorres: I will do code cleanup in the next PR

PR #1131, opened 2021-07-26
UX-1089/Add tenant name to OVA and PF9 CLI walkthrough

knnguy:
### **Changes:**

- Changed tenant value to tenant name instead of tenant id in the OVA walkthrough and PF9 CLI walkthrough

### **Onboard a node page**
Screen Shot 2021-07-26 at 12 17 24 PM

Screen Shot 2021-07-26 at 12 17 30 PM

### **Onboarding flow**
Screen Shot 2021-07-26 at 12 22 05 PM

Screen Shot 2021-07-26 at 12 22 22 PM


jruddell (approved):
Only feedback here is it seems like this is repeated in multiple places, if its identical we should probably merge the rendering portion of this code to a single place to make it easier to change in the future.

jruddell (approved):

PR #1132, opened 2021-07-27
UX-842/disable-onboarding-node-for-airgepped-DU

hgupta2363:
test it by making airgep=true manually
Screenshot 2021-07-27 at 12 22 07 PM


jruddell (approved):

jruddell (changes requested):
Hold on why are you changing CRUDListContainer? this is used in many places from what I understand. We dont want to disable this for every page that uses this. What you should do is update the commands like the ticket calls out when we give them commands to run

PR #1135, opened 2021-07-27
UX-958/Issues on cluster table UI resolved.

rajshekhar23:
Screenshot 2021-07-27 at 8 56 13 PM


src/app/core/components/listTable/ListTableHead.js
```diff
@@ -41,6 +41,7 @@ const styles = (theme) => ({
   },
   cellLabel: {
     color: [theme.palette.text.primary, '!important'],
+    textAlign: 'center',
```
jruddell, line 44: why are we changing this to be center? this isn't what has been designed from our designer.

jruddell (changes requested):
I wouldn't recommend pulling in a task thats a few sprints out and working on it in the current sprint... We have a backlog in the current sprint you can pull from and when that is empty tasks from the next sprint. This one hasn't been reviewed or planned as of yet, center aligning the table headers is not something we should do.

jruddell (approved):

PR #1147, opened 2021-08-03
UX-1052/load logos from features.json

jruddell:
#### Added a new component to handle loading a bad image and showing a fallback in that case. 
image (40)
image (41)

#### Added typings support for PreferencesState. 
* It wasn't defined correctly before as it was more the action payload than the actual state.
* made the user defaults pull out a template of our current defaults 

image
image
image


src/app/core/watchdog/maybe-set-theme-from-features.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const urlBase = process.env.NODE_ENV !== 'production' ? store.appConfig.apiHost : ''
+
+const maybeSetThemeFromFeatures = (dispatch) => {
+  // TODO: change this to use features: maybeSetThemeFromFeatures = (features) => (dispatch)
+  // TODO: with features, we can check for a custom logo, and maybe set the theme once backend supports it
+
+  // For now we set a new logo and favicon, let the view render the logo with a fallback
+  const newFavicon = `${urlBase}${clarityImages}/favicon-v2.png`
+  const newLogo = `${urlBase}${clarityImages}/logo.png`
```
kevinylee1991, line 13: The default logo for the clarity UI is different from the new UI. It seems like we might end up pulling the white logo over the colored one when no custom logo is set?

Screen Shot 2021-08-02 at 5 57 08 PM
versus
Screen Shot 2021-08-02 at 5 57 21 PM

kevinylee1991 (approved):
Left comment regarding the default logo that would be pulled, but otherwise looks good to me. If we don't want to pull that white default logo, we may need that features.json property that you commented about

knnguy (approved):

src/app/core/watchdog/maybe-set-theme-from-features.ts
jruddell: I was testing this on traditional DU's that had the logo and ones that didn't. Never encountered this, can you show me an example of one that I can test on @kevinylee1991

jruddell: We've added a custom flag for these override logos on the old system that we can use to honor loading these. Marking as resolved

PR #1155, opened 2021-08-05
UX-1072,UX-1117/Modified Services list table and created service overview page

knnguy:
### **Modified the services list table**

- Added ports and age columns
- Changed the look of the tags and selectors columns
- Removed internal and external endpoints columns
- Moved the status column next to name
- When user clicks on service name, they get directed to the service overview page

Screen Shot 2021-08-05 at 3 31 35 PM

### ****Added service overview page

**Service Example #1**

Screen Shot 2021-08-13 at 11 06 44 AM

**Service Example #2**

Screen Shot 2021-08-13 at 11 36 11 AM
Screen Shot 2021-08-13 at 11 08 08 AM


knnguy:
@jruddell 

Question. The only reason I'm calling `kubeServiceDetailsLoader` in `service-overview.tsx` is to get the field `kind`. Besides the `apiVersion` and `kind` field, the data is exactly the same as what we get when we get the list of services.
```
  const [[serviceDetails] = {}, loadingServiceDetails] = useDataLoader(kubeServiceDetailsLoader, {
    clusterId,
    namespace,
    service: name,
  })
```
When we get the list of services, the `kind` field is included, but we take it out. 

```
  getClusterKubeServices = async (clusterId) => {
    const url = `/clusters/${clusterId}/k8sapi/api/v1/services`
    const data = await this.client.basicGet<GetClusterKubeServices>({
      url,
      options: {
        clsName: this.getClassName(),
        mthdName: 'getClusterKubeServices',
      },
    })
    return data.items.map(this.convertResource<GetClusterKubeServicesItem>(clusterId))
  }
```

I can put it back in, but the when we get the list of services, `kind === Services`. We want to display `kind === Service,` singular. There's a library called pluralize that can singularize a word. Not sure if I should do that or just make the extra API call. What do you think?

knnguy:
A couple of concerns: 

1. The cpu and memory usage percentage will always be shown as 'N/A' in the utilization card for services. This is because cpu and memory usage percentage comes from nodes. A service can expose multiple pods that are on different nodes, right? So to me, it doesn't make sense to come up with a total cpu and memory usage percentage value for a service. Let me know if this wrong. @chrisj-platform9 

2. In pods/actions.js, I am calling metrics server api for every single pod. It seems like metrics server only works for older clusters right now. Not sure if I should be checking the version of cluster the pod belongs to before calling it or should I just leave it as is? Is it ok to just have a bunch of failed requests?

src/app/plugins/kubernetes/components/pods/helpers.ts
```diff
@@ -0,0 +1,89 @@
+import { isEmpty } from 'ramda'
+import { pathStr } from 'utils/fp'
+
+export const calculateTotalCpuUsage = (pods) => {
```
jruddell, line 4: probably would be good to memoize on these helpers as they are fairly heavy in computation

src/app/plugins/kubernetes/components/pods/helpers.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (cpuUsage && valueNotFound) {
+      valueNotFound = false
+    }
+    // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+    return acc + parseFloat(cpuUsage || 0)
+  }, 0)
+  return valueNotFound ? 'N/A' : `${totalCpuUsage}m`
+}
+
+export const calculateTotalMemoryUsage = (pods) => {
```
jruddell, line 18: we should write this more performantly. Instead of doing a reduce over all pods in each helper, just do it once and calculate the different bits in a single pass. this is what we do for usage stats on a cluster I believe (combinedHosts.ts)

src/app/plugins/kubernetes/components/services/service-overview.tsx
```diff
@@ -0,0 +1,355 @@
...
+  columns: [
+    { id: 'name', label: 'Name' },
+    { id: 'created', label: 'Age', render: calculateAge },
+  ],
+  name: 'Ingresses',
+  title: 'Ingresses',
+}
+
+const renderTable = (tableOptions, data, classes) => {
+  const { List }: any = createCRUDComponents(tableOptions)
```
jruddell, line 236: I dont think we should be using a create crud like this. Why not just render the `<ListTable`?

jruddell (approved):
Looks great, theres a few things we should clean up but overall this is awesome

src/app/plugins/kubernetes/components/pods/helpers.ts
```diff
@@ -1,89 +1,138 @@
...
+  let totalCpuRequests
+  let totalMemRequests
+
+  pods.forEach((pod) => {
+    const {
+      usage: { cpu: cpuUsage, memory: memUsage },
+      limits: { cpu: cpuLimits, memory: memLimits },
+      requests: { cpu: cpuRequests, memory: memRequests },
+    } = getPodResourceStats(pod)
+    totalCpuUsage = cpuUsage ? parseFloat(cpuUsage) + parseFloat(totalCpuUsage || 0) : totalCpuUsage
```
jruddell, line 48: make a simple helper for this please

`const add = (value, total) => value ? parseFloat(value) + parseFloat(total || 0) : total`

src/app/plugins/kubernetes/components/ingresses/actions.ts
```diff
@@ -14,16 +10,8 @@ export const ingressActions = createCRUDActions(DataKeys.Ingresses, {
...
-        return someAsync(
-          pluck<any, string>('name', namespaces).map(async (namespace) =>
-            qbert.getIngresses(clusterId, namespace),
-          ),
-        ).then(flatten)
-      }
-      const ingresses = await qbert.getIngresses(clusterId, namespace)
-      return ingresses
+      const allIngresses = await qbert.getIngresses(clusterId)
+      return allIngresses.find((ingress) => ingress.namespace === namespace) || []
```
jruddell, line 14: do you need to filter the data? our selector should already be applying the namespace filter. probably better to let all of the data write to the store

PR #1158, opened 2021-08-09
UX-951/better errors

jruddell:
#### Cluster Errors Tab
image

#### Platform9 Events Tab
image

#### Notification Detail Modal
image

image


src/app/core/components/notifications/notification-details.tsx
```diff
@@ -0,0 +1,130 @@
...
+        </SimpleLink>
+      </header>
+      <div className={classes.hr} />
+      {nextViewType !== NotificationViewTypes.Parsed && (
+        <article>
+          <DisplayKeyValues keyValuePairs={upperValues} />
+          <div className={classes.hr} />
+          <DisplayKeyValues keyValuePairs={lowerValues} />
+          <CopyToClipboard
+            header="Api Resonse"
```
knnguy, line 105: Typo

knnguy (approved):

src/api-client/ApiClient.ts
```diff
@@ -139,41 +179,68 @@ class ApiClient {
...
+          config: formatResponseConfigPayload(responseConfig),
+          data: responseData,
+          status: responseStatus,
+          statusText: responseStatusText,
+        }
 
         // If data contains user password, remove it
-        if (cfg?.data?.includes('{"methods":["password"]')) {
-          const dataObj = JSON.parse(cfg.data)
+        if (responseConfig?.data?.includes('{"methods":["password"]')) {
```
kevinylee1991, line 213: Just a note, there may be other sensitive info fields ex. with VMs with cloudinit, passwords may be sent through there, possibly other API calls as well that I can't think of off the top of my head. Perhaps sensitive keys may need to be stored in the api method metadata?

Also I can't tell, is the responseConfig data the request body that was sent or is that something else?

src/app/core/components/notifications/NotificationsPage.tsx
```diff
@@ -1,58 +1,224 @@
...
+            notificationType={ErrorNotificationType.Platform9Event}
+          />
+        </Tab>
+      </Tabs>
+    </>
   )
 }
 
 export default NotificationsPage
+
```
kevinylee1991, line 205: still need this?

src/api-client/Appbert.ts
```diff
@@ -14,7 +14,7 @@ class Appbert extends ApiService {
     return this.client.keystone.getServiceEndpoint('appbert', 'admin')
   }
 
-  @trackApiMethodMetadata({ url: '/clusters', type: 'GET' })
+  @trackApiMethodMetadata({ url: '/clusters', type: 'GET', error: { isClusterError: false } })
```
kevinylee1991, line 17: Preferable to have isClusterError: false declared or leave empty? I noticed some in qbert (ex. qbert's /clusters) were left alone

kevinylee1991 (approved):
looks good

src/api-client/Appbert.ts
jruddell: It doesn't make much of a difference, false and undefined are treated the same in a falsey check

src/api-client/ApiClient.ts
jruddell: If you can draw up an example we can think about extending the model to include something like sensitive keys. We can also just have this be global and not specific to a single api method's metadata

src/app/core/components/notifications/NotificationsPage.tsx
jruddell: nope :)

PR #1163, opened 2021-08-10
UX-1122, UX-1110/Add cluster and namespace defaults to cluster and namespace dropdowns in the resources list pages

knnguy:
### **Added cluster and namespace defaults to the cluster and namespace picklists in the following list pages:**
- Pods list page
- Deployments list page
- Services list page
- Storage classes list page
- Roles list page
- Cluster Roles list page
- Cluster Role Bindings list page
- Role bindings list page
- Alarms list page
- Rules list page
- Deployed apps list page

### **Hid the set default buttons and the defaults callout fields in the cloud provider verification forms for non decco du**
Screen Shot 2021-08-09 at 4 41 36 PM
Screen Shot 2021-08-09 at 4 50 36 PM

### **Disabled the set default action in the clusters list table for non decco du**
Screen Shot 2021-08-09 at 4 59 04 PM


src/app/core/utils/helpers.ts
```diff
@@ -1,8 +1,14 @@
 import { AppPlugins, CustomerTiers } from 'app/constants'
 import { pathStrOr } from 'utils/fp'
+import { prop } from 'ramda'
 
 export const isDecco = (features) => pathStrOr(false, 'experimental.kplane', features)
 
+export const isDeccoEnv = (_, store) => {
+  const { features } = prop('session', store)
```
jruddell, line 8: you should probably use `sessionStoreKey` here

xantorres (approved):

PR #1164, opened 2021-08-10
Fix create and edit rbac profile issues

xantorres:
Fixes 2 and ½ first pages of issues described in the 5.3 staging UI issues document


jruddell (approved):

src/app/plugins/kubernetes/components/rbac/profiles/create/review-form.tsx
```diff
@@ -0,0 +1,167 @@
...
+  )
+  const [roles] = useDataLoader(roleActions.list, defaultParams)
+  const [clusterRoles] = useDataLoader(clusterRoleActions.list, defaultParams)
+  const [roleBindings] = useDataLoader(roleBindingActions.list, defaultParams)
+  const [clusterRoleBindings] = useDataLoader(clusterRoleBindingActions.list, defaultParams)
+  const filteredRoles = useMemo(() => roles.filter(({ id }) => !!selectedRoles[id]), [
+    roles,
+    selectedRoles,
+  ])
+  const filteredClusterRoles = useMemo(
```
kevinylee1991, line 71: If we kept the full cluster role objects etc. in the wizardContext when we select them, we could possibly just use that data, and using the dataLoader and filtering them down again may not be necessary

kevinylee1991 (approved):
Just 1 comment, otherwise lgtm

src/app/plugins/kubernetes/components/rbac/profiles/create/review-form.tsx
xantorres: It doesn't matter, the dataLoader will pull data from the cache anyways, so it will be instant

PR #1166, opened 2021-08-12
UX-913/Add-OS-versions-in-AWS-Custom-Template

Nalinmujoo:
Add OS Version in AWS Custom Template:
Screenshot 2021-08-11 at 11 32 11 AM

Error: When we are creating the cluster it's showing the following error
Screenshot 2021-08-12 at 11 13 28 AM


src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/operating-system.tsx
```diff
@@ -2,8 +2,9 @@ import React from 'react'
 import PicklistField from 'core/components/validatedForm/PicklistField'
 
 const operatingSystemOptions = [
-  { label: 'Ubuntu', value: 'ubuntu' },
-  { label: 'CentOS', value: 'centos' },
+  { label: 'Ubuntu 18.04', value: 'ubuntu18' },
```
jruddell, line 5: I dont think we can hardcode ubuntu supported versions, this will most likely need to be around an API call. And does the backend support these values?

PR #1183, opened 2021-08-23
Added yaml editor for edit rbac profile roles

xantorres:
image

image


jruddell (approved):
Looks good, lets get this in. In the future we should look at moving the duplicated code to a more generic system

PR #1187, opened 2021-08-25
UX-1064/remove-reader-role-and-upadate-member-role-to-Self-Service-User-role

hgupta2363:
remove reader role using filter function
show dispalyName property for all roles
Screenshot 2021-08-25 at 12 33 14 PM


src/app/plugins/account/components/userManagement/common/RolesPicklist.js
```diff
@@ -11,7 +11,10 @@ import { mngmRoleActions } from 'account/components/userManagement/roles/actions
 const RolesPicklist = forwardRef(
   ({ allRoles, id, loading, onChange, value, selectFirst, disabled, ...rest }, ref) => {
     const [roles, rolesLoading] = useDataLoader(mngmRoleActions.list, { allRoles })
-    const options = useMemo(() => projectAs({ label: 'name', value: 'id' }, roles), [roles])
+    const filteredRole = roles.filter((role) => role.name !== 'reader')
```
jruddell, line 14: if the variable is an array it should be pluralized. `const filteredRoles = `

jruddell (approved):

PR #1188, opened 2021-08-26
UX-1085, UX-1136/Deployments Overview Page

knnguy:
Screen Shot 2021-08-26 at 12 18 26 PM

Screen Shot 2021-08-26 at 12 41 49 PM

Screen Shot 2021-08-26 at 12 17 32 PM

Screen Shot 2021-08-26 at 12 15 55 PM


src/app/core/utils/routes.ts
```diff
@@ -605,6 +605,13 @@ export const routes = {
       },
       name: 'Deployments:Add',
     }),
+    overview: Route.register({
+      url: `${pluginRoutePrefix}/workloads/deployments/:id/overview/`,
```
jruddell, line 609: generally i dont think we leave a trailing slash at the end of our urls `/:id/overview`

src/app/plugins/kubernetes/components/deployments/selectors.ts
```diff
@@ -0,0 +1,105 @@
...
+    servicesByClusterIdAndNamespaceSelector,
+    clusters,
+    importedClusters,
+    qbertEndpoint,
+  ): IDeploymentSelector[] => {
+    return arrayIfEmpty(
+      rawDeployments
+        .map((rawDeployment) => {
+          const { clusterId } = rawDeployment
+          const allClusters = [...clusters, ...importedClusters]
```
jruddell, line 40: you dont need to create this const every time you map a deployment, just do it once outside of the return. And better yet convert it to an object with the uuid as the key to just simply lookup a cluster by its uuid. we have an all clusters selector, you could make one that presents it in an object format for all clusters and then use that here.

src/app/plugins/kubernetes/components/ingresses/helpers.ts
```diff
@@ -0,0 +1,38 @@
+import { head, isEmpty, pluck, flatten } from 'ramda'
+import { ensureArray } from 'utils/fp'
+
+export const getResourceIngresses = (services = [], ingresses = []) => {
```
jruddell, line 4: `moize` these helpers please

src/app/plugins/kubernetes/components/ingresses/helpers.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (isEmpty(services) || isEmpty(ingresses)) return []
+  const servicesList = ensureArray(services)
+  const serviceNames = pluck('name', servicesList)
```
jruddell, line 7: lets use objects / maps more frequently for these kind of lookup items. this way we dont have to do a `find(find(includes()))`

src/app/plugins/kubernetes/components/services/helpers.ts
```diff
@@ -0,0 +1,25 @@
...
+  )
+}
+
+const labelsMatchSelector = (labels = {}, selector = {}) => {
+  if (isEmpty(labels) || isEmpty(selector)) {
+    return false
+  }
+  for (const [selectorKey, selectorValue] of Object.entries(selector)) {
+    if (
+      !Object.entries(labels).find(
```
jruddell, line 17: why do you need to convert this to an array and then search it? An object gives you a direct lookup that is O(1), so shouldn't this suffice?
```
const labelsMatchSelector = (labels = {}, selector = {}) => {
  if (isEmpty(labels) || isEmpty(selector)) {
    return false
  }
  for (const [selectorKey, selectorValue] of Object.entries(selector)) {
    if (labels.hasOwnProperty(selectorKey) && labels[selectorKey] === selectorValue) {
      continue
    } else {
      return false
    }
  }
  return true
} 
```

src/app/plugins/kubernetes/components/services/helpers.ts
```diff
@@ -0,0 +1,25 @@
+import { isEmpty } from 'ramda'
+import { getPodTemplate } from '../config-maps/helpers'
+
+export const getServicesForResource = (resource = {}, services = []) => {
```
jruddell, line 4: `moize` these as well please

src/app/utils/misc.js
```diff
@@ -62,6 +62,16 @@ export const secondsToString = (seconds) => {
   return results.join(', ')
 }
 
+/**
+ * Given a timestamp, returns the years, months, days, hours and minutes in a human readable format
+ * @param creationTimeStamp
+ * @returns {string}
+ */
+export const calculateAge = (creationTimestamp) => {
```
jruddell, line 70: `moize`

jruddell (changes requested):
Few places we can optimize, lets get this addressed and merged so we can work off of this for storage classes (sync with @rajshekhar23 on this) and for pods work.

src/app/plugins/kubernetes/components/common/persistent-volume-claims/helpers.tsx
```diff
@@ -0,0 +1,94 @@
+import Text from 'core/elements/text'
+import ClusterStatusSpan from 'k8s/components/infrastructure/clusters/ClusterStatus'
+import { IClusterStatus } from 'k8s/components/infrastructure/clusters/model'
+import { complement, filter, flatten, isEmpty, isNil, map, pipe, pluck } from 'ramda'
+import React from 'react'
+import { arrayIfEmpty } from 'utils/fp'
+
+export const getPvcs = (pods = [], allPersistentVolumeClaims = []) => {
```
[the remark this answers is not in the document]
knnguy: Turns out I didn't use the function at all. Decided to map pvs to pods instead of pods to pvcs

jruddell (approved):

PR #1196, opened 2021-09-02
Surface pod error status

knnguy:
### **Pods List Table**
Screen Shot 2021-09-02 at 3 59 12 PM

### **Deployments Overview (and Service Overview as well)**
Screen Shot 2021-09-02 at 3 49 07 PM

Screen Shot 2021-09-02 at 3 55 27 PM

Screen Shot 2021-09-02 at 3 57 32 PM


jruddell (approved):

src/app/plugins/kubernetes/components/pods/selectors.ts
```diff
@@ -17,6 +17,66 @@ import { clientStoreKey } from 'core/client/clientReducers'
...
+        message: stat?.state?.waiting?.message || '',
+        isError:
+          stat?.state?.waiting?.reason !== 'ContainerCreating' ||
+          stat?.state?.waiting?.reason !== 'PodInitializing',
+      }
+      if (!podStatus?.isError && waitingStatus.isError) {
+        podStatus = waitingStatus
+      }
+      return waitingStatus
+    } else {
```
jruddell, line 64: technically you dont need the final else. since the default behavior would suffice

PR #1199, opened 2021-09-03
UX-1085,UX-1150/Add deployemnt and service events page

knnguy:
Screen Shot 2021-09-02 at 4 31 37 PM


src/app/plugins/kubernetes/components/deployments/events.tsx
```diff
@@ -0,0 +1,45 @@
...
+    namespace: deployment?.namespace,
+  })
+
+  const ingresses = useMemo(
+    () => getResourceIngresses(deployment?.deploymentServices, allIngresses),
+    [allIngresses, deployment],
+  )
+
+  const pvcs = useMemo(() => {
+    const podNames = pluck<any, any>('name', deployment?.deploymentPods || [])
```
jruddell, line 25: convert this to an object so you can just do a lookup instead of an includes.

jruddell (approved):
small optimization and we can get this in! nice work

PR #1201, opened 2021-09-04
UX-1083/Updated Kubevirt New VM wizard

kevinylee1991:
Screen Shot 2021-09-03 at 6 14 31 PM
Screen Shot 2021-09-03 at 6 14 39 PM
Screen Shot 2021-09-03 at 6 14 48 PM
Screen Shot 2021-09-03 at 6 15 10 PM
Screen Shot 2021-09-03 at 6 15 18 PM
Screen Shot 2021-09-03 at 6 15 24 PM
Screen Shot 2021-09-03 at 6 15 34 PM
Screen Shot 2021-09-03 at 6 15 47 PM
Screen Shot 2021-09-03 at 6 16 26 PM
Screen Shot 2021-09-03 at 6 16 37 PM
Screen Shot 2021-09-03 at 6 16 57 PM
Screen Shot 2021-09-03 at 6 21 24 PM
Screen Shot 2021-09-03 at 6 21 35 PM
Screen Shot 2021-09-03 at 6 21 42 PM
Screen Shot 2021-09-03 at 6 21 49 PM
Screen Shot 2021-09-03 at 6 22 05 PM
Screen Shot 2021-09-03 at 6 22 11 PM


src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
```diff
@@ -0,0 +1,233 @@
+import React, { useEffect } from 'react'
```
jruddell, line 1: this page has been reworked, what changes did you apply to this page? we moved it to typescript so you're re-adding the same page as javascript now

src/app/plugins/kubernetes/components/virtual-machines/actions.ts
```diff
@@ -107,5 +120,110 @@ export const virtualMachineActions = createCRUDActions(DataKeys.VirtualMachines,
...
+    return dataVolumes
+  },
+  {
+    entityName: 'DataVolume',
+    uniqueIdentifier: 'metadata.uid',
+    indexBy: ['clusterId', 'namespace'],
+  },
+)
+
+export const pvcsLoader = createContextLoader(
```
jruddell, line 214: can you work with kim and raj on this? there's already been work done around loading pvcs. We should consolidate under one action so we dont have conflicts

src/app/plugins/kubernetes/components/virtual-machines/add/basic-step.tsx
```diff
@@ -0,0 +1,393 @@
...
+                />
+                {wizardContext.numVms &&
+                  [...Array(wizardContext.numVms - 1)].map((_, idx) => (
+                    <TextField
+                      id={idx}
+                      label="Name"
+                      onChange={(value) =>
+                        setWizardContext({
+                          vmNames: [
+                            ...wizardContext.vmNames.slice(0, idx),
```
jruddell, line 341: why not just use splice to insert at an index `[...wizardContext.vmNames].splice(idx, 0, value)`

src/app/plugins/kubernetes/components/virtual-machines/add/helpers.ts
```diff
@@ -0,0 +1,263 @@
...
+      dataVolume: {
+        name: disk.selectedDataVolumes[0].name,
+      },
+    }
+  }
+  return {}
+}
+
+const getDataVolumeObject = (disk) => {
+  if (disk?.sourceType === 'httpUrl') {
```
jruddell, line 31: this is a bit sloppy. Can you use a more functional way to do this? I think it'll clean it up a lot and reduce the if else madness

src/app/plugins/kubernetes/components/virtual-machines/add/helpers.ts
```diff
@@ -0,0 +1,263 @@
...
+    }
+  } else if (disk.sourceType === 'clone') {
+    return null
+  } else if (disk.sourceType === 'volume') {
+    return null
+  }
+  return null
+}
+
+export const getVolumeJson = (wizardContext) => {
```
jruddell, line 87: should memoize this function and also pull out the specific fields you need from wizard context as individual params so the memoize can be effective

src/app/plugins/kubernetes/components/virtual-machines/add/storage-step.tsx
```diff
@@ -0,0 +1,365 @@
...
+]
+
+const pvcColumns = [
+  { id: 'name', label: 'Name' },
+  { id: 'spec.resources.requests.storage', label: 'Size' },
+]
+
+export const StorageStep = ({ wizardContext, setWizardContext, onNext }) => {
+  const classes = useStyles({})
+  const [dataVolumes, loading] = useDataLoader(dataVolumesLoader, {
```
jruddell, line 124: since you're already adding as any at the end of this data loader, you should also use typings here `const [dataVolumes, loading]: IUseDataLoader<DataVolumeTypeHere> = useDataLoader(..., ...) as any`

src/app/plugins/kubernetes/components/virtual-machines/add/storage-step.tsx
```diff
@@ -0,0 +1,365 @@
...
+          ...wizardContext.storageDisks.slice(0, idx),
+          {
+            ...wizardContext.storageDisks[idx],
+            ...value,
+          },
+          ...wizardContext.storageDisks.slice(idx + 1),
+        ],
+      })
+    },
+    [wizardContext, setWizardContext],
```
jruddell, line 164: instead of depending on the whole wizard context, just reference a specific value inside it, this will make these memoizers much more effective.

src/app/plugins/kubernetes/components/virtual-machines/add/storage-step.tsx
```diff
@@ -0,0 +1,365 @@
...
+    })
+  }, [wizardContext, setWizardContext])
+
+  const updateStorageDisk = useCallback(
+    (rowId, value) => {
+      const idx = wizardContext.storageDisks.findIndex((disk) => disk.rowId === rowId)
+      if (idx < 0) {
+        return
+      }
+      setWizardContext({
```
jruddell, line 153: another place to just use splice instead. 
```
setWizardContext({
  storageDisks: [ ...wizardContext.storageDisks ].splice(idx, 0, {
    ...wizardContext.storageDisks[idx],
    ...value,
  })
})
```

jruddell (changes requested):
Few things you need to address, also, why are we adding new icon files? I think we've standardized on using fontawesome icons (Image loading is more overhead that we can avoid). There should be an equivalent icon in fa that you see on the designs.

src/app/plugins/kubernetes/components/storage/AddStorageClassPage.js
kevinylee1991: ah okay, i'll check on this I didn't notice a conflict when I pulled in master

src/app/plugins/kubernetes/components/virtual-machines/actions.ts
kevinylee1991: oh I actually have this finished when I was checking out of the merged code was working. I will push those fixes I caught right now to this PR

anmolsachan:
132077839-c7dccc7f-6866-409f-a666-d2cb88cba7b5

As someone who saw this for the first time, when I saw the column Age(which makes complete sense from Admin/Devs perspective), I was expecting the humanize form like 5 hours, 4 minutes, 3seconds, and below that the actual timestamp could be mentioned.  Or there could be 2 separate columns for Timestamp and Age. Just a suggestion.

Age

5 hours, 4 minutes, 3seconds
Created At: 09/03/2021, 5:47 PM

jruddell:
> As someone who saw this for the first time, when I saw the column Age(which makes complete sense from Admin/Devs perspective), I was expecting the humanize form like 5 hours, 4 minutes, 3seconds, and below that the actual timestamp could be mentioned. Or there could be 2 separate columns for Timestamp and Age. Just a suggestion.
> 
> Age
> 
> 5 hours, 4 minutes, 3seconds
> Created At: 09/03/2021, 5:47 PM

I think it would be better to surface that information underneath a tooltip on hover vs persistent. This date format is used in many places and that may not fit well on those other views.

anmolsachan:
> > As someone who saw this for the first time, when I saw the column Age(which makes complete sense from Admin/Devs perspective), I was expecting the humanize form like 5 hours, 4 minutes, 3seconds, and below that the actual timestamp could be mentioned. Or there could be 2 separate columns for Timestamp and Age. Just a suggestion.
> > Age
> > 5 hours, 4 minutes, 3seconds
> > Created At: 09/03/2021, 5:47 PM
> 
> I think it would be better to surface that information underneath a tooltip on hover vs persistent. This date format is used in many places and that may not fit well on those other views.

Makes sense. Thanks for the info. :smiley:

jruddell:
> > > As someone who saw this for the first time, when I saw the column Age(which makes complete sense from Admin/Devs perspective), I was expecting the humanize form like 5 hours, 4 minutes, 3seconds, and below that the actual timestamp could be mentioned. Or there could be 2 separate columns for Timestamp and Age. Just a suggestion.
> > > Age
> > > 5 hours, 4 minutes, 3seconds
> > > Created At: 09/03/2021, 5:47 PM
> > 
> > 
> > I think it would be better to surface that information underneath a tooltip on hover vs persistent. This date format is used in many places and that may not fit well on those other views.
> 
> Makes sense. Thanks for the info. 😃

Yep! Thanks for the recommendation! definitely a nice thing to surface around these dates. We have moment so we could use `fromNow()` as well

src/app/plugins/kubernetes/components/virtual-machines/add/basic-step.tsx
kevinylee1991: splice edits the original object but let me know if you like the implementation better

kevinylee1991:
Screen Shot 2021-09-09 at 6 31 57 PM
Screen Shot 2021-09-09 at 6 32 19 PM
Screen Shot 2021-09-09 at 6 44 32 PM

src/api-client/Qbert.ts
```diff
@@ -1062,37 +1099,127 @@ class Qbert extends ApiService {
...
+        clsName: this.getClassName(),
+        mthdName: 'getNetworks',
+      },
+    })
+    return data.items.map(this.convertResource(clusterId))
+  }
+
+  getDataVolumes = async (clusterId, namespace) => {
+    const url = `/clusters/${clusterId}/k8sapi/apis/cdi.kubevirt.io/v1beta1/namespaces/${namespace}/datavolumes`
+    const data = await this.client.basicGet<any>({
```
jruddell, line 1213: instead of any types, can you use the autogenerator to create types based on the return of these api calls? follow the convention of putting a model in a `model` file where the component is please

src/app/plugins/kubernetes/components/virtual-machines/add/add.tsx
```diff
@@ -0,0 +1,165 @@
+import React, { useCallback, useMemo } from 'react'
```
jruddell, line 1: rename this to `index.tsx`

src/app/plugins/kubernetes/components/virtual-machines/add/storage-step.tsx
```diff
@@ -0,0 +1,376 @@
...
+  }, [wizardContext.storageDisks, setWizardContext])
+
+  const updateStorageDisk = useCallback(
+    (rowId, value) => {
+      const idx = wizardContext.storageDisks.findIndex((disk) => disk.rowId === rowId)
+      if (idx < 0) {
+        return
+      }
+      const storageDisks = wizardContext.storageDisks
+      storageDisks.splice(idx, 1, {
```
jruddell, line 163: `splice` mutates the original array. make sure you are making a copy of storage disks before calling splice `const storageDisks = [...wizardContext.storageDisks]`

src/app/plugins/kubernetes/components/virtual-machines/selectors.ts
```diff
@@ -9,6 +9,52 @@ import { filterIf } from 'utils/fp'
 import { allKey } from 'app/constants'
 import createSorter from 'core/helpers/createSorter'
 
+export const virtualMachineInstancesSelector = createSelector(
+  [
+    getDataSelector<DataKeys.VirtualMachineInstances>(DataKeys.VirtualMachineInstances, [
+      'clusterId',
+    ]),
+    clustersSelector,
```
jruddell, line 17: you can use the all cluster selector here, or better yet use the `defaultAllClustersSelector` variable i've exported

src/app/plugins/kubernetes/components/virtual-machines/virtual-machine-power-operation-dialog.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <Text variant="subtitle1">{title}</Text>
+      </DialogTitle>
+      <Progress loading={poweringVirtualMachine} renderContentOnMount maxHeight={60}>
+        <DialogContent>
+          <Text variant="body1" component="div">
+            Please confirm you want to {params?.operation} this VM.
+          </Text>
+        </DialogContent>
+        <DialogActions>
+          <Button variant="outlined" color="secondary" onClick={onClose}>
```
jruddell, line 38: dont use material-ui for elements that we have our own for. We will be getting rid of material ui in the near future.

src/app/plugins/kubernetes/components/virtual-machines/vmi-power-operation-dialog.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { useCallback } from 'react'
+import Text from 'core/elements/text'
+import { Dialog, DialogContent, DialogActions, Button, DialogTitle } from '@material-ui/core'
+import Progress from 'core/components/progress/Progress'
+import useDataUpdater from 'core/hooks/useDataUpdater'
+import { virtualMachineInstanceActions } from './actions'
+
+const capitalizeFirstLetter = (string = '') => {
```
jruddell, line 8: we have a helper method `capitalizeString` that does this I believe

src/app/plugins/kubernetes/components/virtual-machines/vmi-power-operation-dialog.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <Text variant="subtitle1">{title}</Text>
+      </DialogTitle>
+      <Progress loading={poweringVmi} renderContentOnMount maxHeight={60}>
+        <DialogContent>
+          <Text variant="body1" component="div">
+            Please confirm you want to {params?.operation} this Virtual Machine Instance.
+          </Text>
+        </DialogContent>
+        <DialogActions>
+          <Button variant="outlined" color="secondary" onClick={onClose}>
```
jruddell, line 38: `/core/elements/button`

jruddell (approved):
some minor fixes, after that lets get this merged

PR #1208, opened 2021-09-09
UX-1101/Storage details

jruddell:
### Overview Page
image

### Yaml Page
I also addressed the styling so this should look the same everywhere this generic entity page is used
image

Todo:
- [ ] need to add events page
- [ ] need to fix transient states so there is feedback to the user.
- [ ] need to convert to yaml instead of rendering JSON
- [ ] need to validate the yaml


rajshekhar23 (approved):
For me everything looks good !

src/api-client/Qbert.ts
```diff
@@ -854,6 +872,25 @@ class Qbert extends ApiService {
     return data.items.map(this.convertResource(clusterId))
   }
 
+  @trackApiMethodMetadata({
+    url: '/clusters/:clusterId/k8sapi/apis/storage.k8s.io/v1/storageclasses/:name',
+    type: 'PUT',
+    params: ['clusterId'],
```
knnguy, line 878: Missing storage class name as a param here

src/api-client/Qbert.ts
```diff
@@ -854,6 +872,25 @@ class Qbert extends ApiService {
...
+    error: { isClusterError: true, k8sResource: 'StorageClass' },
+  })
+  updateClusterStorageClasses = async (clusterId, name, body) => {
```
knnguy, line 881: Shouldn't this be singular? updateClusterStorageClass

src/app/plugins/kubernetes/components/services/overview.tsx
```diff
@@ -164,7 +166,12 @@ const ServiceOverview = ({ service }) => {
   return (
     <EntityOverviewPage headerCards={headerCards}>
       {/* Selectors Table */}
-      <Text className={classes.title} variant="subtitle2" component="h3">
+      <LabelsAndAnnotationsTable
+        title="Selectors"
+        emptyMessage="No selectors"
+        labelRows={selectorRows}
+      />
+      {/* <Text className={classes.title} variant="subtitle2" component="h3">
```
knnguy, line 174: We don't need this anymore, right?

knnguy (approved):

src/api-client/Qbert.ts
jruddell: Isn't it already singular?

PR #1214, opened 2021-09-10
Fix how loading works in the pods, deployments, and services details page

knnguy:
**Changes:**
- Modified the entity routes
- Added `clusterId` to the urls to load the resources faster 
- Added progress component to the data tables
- Removed the progress component from the entity index pages. We will have each individual page render its own progress component.

https://user-images.githubusercontent.com/23369276/132912966-adc36f1c-f623-4c95-9b99-0a8af8d0790a.mov


src/app/core/utils/routes.ts
```diff
@@ -659,35 +614,12 @@ export const routes = {
...
-    yaml: Route.register({
-      url: `${pluginRoutePrefix}/workloads/deployments/:id/overview#yaml`,
-      defaultParams: {
-        plugin: AppPlugins.Kubernetes,
-      },
-      name: 'Deployment:YAML',
-    }),
-    events: Route.register({
-      url: `${pluginRoutePrefix}/workloads/deployment/:id/overview#events`,
+    details: Route.register<{ clusterId: string; id: string; tab: string }>({
```
jruddell, line 617: you can use an enum for the tabs to enforce specific values as well

src/app/plugins/kubernetes/components/common/entity/data-table.tsx
```diff
@@ -4,14 +4,21 @@ import { isEmpty } from 'ramda'
...
+}
+
+const useStyles = makeStyles<Theme, Partial<Props>>((theme) => ({
   title: {
     margin: theme.spacing(2, 0, 2, 0),
   },
   tableContainer: {
     backgroundColor: theme.palette.grey[100],
-    height: 'max-content',
+    height: ({ loading }) => (loading ? '100px' : 'max-content'),
```
jruddell, line 21: do this on the minHeight field so we keep px values specific to one property

jruddell: actually we already specify 80px for the min height, you can just specify that on the progress component to keep those in sync, no need to conditionally change the height

src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -246,7 +246,12 @@ export const getAllClusters = async (reload = false) => {
 export const clusterEventsActions = createCRUDActions(ActionDataKeys.Events, {
   listFn: async ({ clusterId }) => {
     Bugsnag.leaveBreadcrumb('Attempting to get cluster events', { clusterId })
-    return qbert.getClusterEvents(clusterId)
+    try {
```
jruddell, line 249:
```
try {
  return qbert.getClusterEvents(clusterId)
} catch {
  return []
}
```

src/app/plugins/kubernetes/components/pods/index.tsx
```diff
@@ -37,9 +36,10 @@ const PodOverviewPage = () => {
   const classes = useStyles()
   const { match } = useReactRouter()
   const id = match.params.id
-  const [pods, loading] = useDataLoader(podActions.list, { clusterId: allKey })
+  const clusterId = match.params.clusterId
+  const [pods, loading] = useDataLoader(podActions.list, { clusterId })
   const pod = pods.find((pod) => pod.id === id)
-  const [updatePod, updating] = useDataUpdater(podActions.update)
```
jruddell, line 42: shouldn't we keep the updating boolean and pass it as loading to the views?

jruddell (approved):
This looks awesome! A few things to tweak and then lets get this in, great work!

src/app/plugins/kubernetes/components/common/entity/data-table.tsx
```diff
@@ -33,28 +40,30 @@ const useStyles = makeStyles((theme: Theme) => ({
...
-        ) : (
-          <Text className={classes.noneText} variant="caption1">
-            {`No ${options.title}`}
-          </Text>
-        )}
-      </div>
+      <Progress loading={loading} minHeight={90}>
+        <div className={classes.tableContainer}>
+          {!isEmpty(data) && (
+            <ListTable
```
jruddell, line 52: you can pass loading to the table itself instead of wrapping in progress

src/app/plugins/kubernetes/components/common/entity/data-table.tsx
```diff
@@ -33,28 +40,30 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+              {...options}
+              data={data}
+              showCheckboxes={false}
+              showPagination={false}
+              compactTable
```
jruddell, line 57: if this loader is too big for these compact tables, one thing we could think about doing is using our inline loader when the table is compact (same one we use on inputs/buttons)

PR #1217, opened 2021-09-11
UX-1202/Data loading with required params + loading messaging feedback

jruddell:
Fixed the data loader, added a better loading system to let the data populate quickly and bring dependent data in on the side

https://user-images.githubusercontent.com/1203655/132942917-9e201236-0303-4549-9cdf-19a3c110124e.mov

Todo:
- [x] Sync any dependent data requests with the loading state cache (currently only the data loader does this) so we have these variables more properly set.
- [x] Update existing actions to follow deployments / services model, and decouple other dependent requests as needed.
- [x] Go over all views and compare changes to verify nothing is missing
- [x] Make the isDataLoading helper also check for the existence of required params so the loading doesn't sometimes flash off and then on again


src/app/plugins/kubernetes/components/app-catalog/deployed-apps/deployed-apps-list-page.tsx
```diff
@@ -26,30 +28,35 @@ const usePrefParams = createUsePrefParamsHook('Deployed Apps', listTablePrefs)
...
+    const [repositories] = useDataLoader(repositoryActions.list)
 
     const filteredDeployedApps = useMemo(
       () => deployedApps.filter((app) => app.status !== 'uninstalling'),
       [deployedApps],
     )
 
+    const { loading, message } = useSelector(
+      isDataLoadingSelector([
+        clusterActions.cacheKey,
```
knnguy, line 45: I don't see you loading clusters. I think you accidentally took out const [_, loadingClusters] = useDataLoader(clusterActions.list)?

src/app/plugins/kubernetes/components/ingresses/actions.ts
```diff
@@ -15,14 +15,9 @@ export const ingressActions = createCRUDActions(DataKeys.Ingresses, {
...
-        const ingresses = await someAsync(
-          pluck<any, string>('uuid', clusters).map(qbert.getIngresses),
-        ).then(flatten)
-        return ingresses
-      }
-      const ingresses = await qbert.getIngresses(clusterId)
-      return ingresses
+      return clusterId === allKey
+        ? someAsync(pluck<any, string>('uuid', clusters).map(qbert.getIngresses)).then(flatten)
+        : qbert.getIngresses(clusterId)
```
knnguy, line 20: Would an exception still be caught if you directly return a promise?

knnguy (approved):

src/app/plugins/kubernetes/components/ingresses/actions.ts
jruddell: yes as long as its in the try block it'll be caught

src/app/core/hooks/useDataLoader.ts
```diff
@@ -47,56 +40,10 @@ const useDataLoader = (loaderFn, params = emptyObj, options = emptyObj as any) =
...
-        }
-
-        // Return an empty array, thus preventing a blank page error
-        // that could be consequence of returning a null or undefined
-        return emptyArr
-      }
-    },
-    [refetching, memoizedParams, defaultSelector],
-  )
+  const selector = useMemo(() => selectorCreator(defaultParams), [defaultParams])
```
xantorres, line 43: const defaultSelector = useMemo(() => selectorCreator(defaultParams), [defaultParams])
const selector = useCallback(state => defaultSelector(state, memoizedParams), [defaultSelector, memoizedParams])

src/app/core/hooks/useDataLoaderWithMessage.ts
```diff
@@ -0,0 +1,60 @@
...
+const useDataLoaderWithMessage = ({
+  loaderFn,
+  params = emptyObj,
+  options = emptyObj,
+  dependentCacheKeys = [],
+}: IUseDataLoaderWithMessage) => {
+  const { requiredParams = undefined } = options
+  const memoizedParams = memoizedDep(params)
+  const cacheKeys = useMemo(() => [...dependentCacheKeys, loaderFn.cacheKey], [
+    loaderFn,
```
kevinylee1991, line 30: this could be loaderFn.cacheKey right?

src/app/core/hooks/useDataLoader.ts
```diff
@@ -125,42 +76,18 @@ const useDataLoader = (loaderFn, params = emptyObj, options = emptyObj as any) =
...
-          dispatch(cacheActions.setLoading({ cacheKey, loading: false }))
-        }
+        loaderFn(memoizedParams, refetch, loadingFeedback)
       }
     },
     [loaderFn, memoizedIndexedParams, cachedParams, loadingFeedback, data],
   )
 
   // Load the data on component mount and every time the params change
   useEffect(() => {
```
knnguy, line 86: Shouldn't requiredParams be a dependency as well?

src/app/core/helpers/list-with-dependencies.ts
```diff
@@ -0,0 +1,36 @@
+import { either, equals, find, isNil, path, pickAll, pipe, reject } from 'ramda'
+import { allKey } from 'app/constants'
+import store from 'app/store'
+import { arrayIfNil } from 'utils/fp'
+import { memoizedDep } from 'utils/misc'
+import { cacheStoreKey, loadingStoreKey, paramsStoreKey } from 'core/caching/cacheReducers'
+
+const listFnWithDependencies = (listFn, deps) => async (params, refetch) => {
```
kevinylee1991, line 8: Just want to confirm the the main benefit here.

So with this method, the list function in the action will resolve quicker (assuming there is a dependency that is taking a while to load). But we can selectively show the loading screen depending on which dependencies we want to wait for, and potentially render data that is not complete earlier. Are there other big benefits that I'm missing?

kevinylee1991 (approved):

src/app/core/helpers/list-with-dependencies.ts
jruddell: Correct, we can load any form of dependent data async, show loading on required items to load the primary piece of data... other dependencies can show affordance in the page itself (deployments table spinner in a list table column for pods for instance). The loading of everything is a lot faster. There is also direct feedback as to what is loading.

src/app/core/hooks/useDataLoaderWithMessage.ts
jruddell: it wont make much of a difference as were looking at a reference to a function or a string value to determine if we recreate the cacheKey arr.

src/app/core/hooks/useDataLoader.ts
jruddell: It doesn't matter as the `requiredParamsProvided` always update when `requiredParams` do. So they have a 1:1 correlation

PR #1219, opened 2021-09-13
UX-1152/Fix scale worker nodes bug

knnguy:


src/app/plugins/kubernetes/components/infrastructure/clusters/actions.js
```diff
@@ -95,7 +95,9 @@ export const clusterActions = createCRUDActions(ActionDataKeys.Clusters, {
     }
   },
   updateFn: async ({ clusterId, ...params }) => {
-    const updateableParams = 'name tags enableProfileAgent'.split(' ')
+    const updateableParams = 'name tags enableProfileAgent numWorkers numMinWorkers numMaxWorkers'.split(
```
jruddell, line 98: I'd rather we define this as an array, vs a string that we split to generate it into an array

jruddell (approved):

PR #1222, opened 2021-09-14
UX-1207/Add test for tenants creation

hgupta2363:
1 add test cases for tenant creation flow

2 add test cases for tenant deletion flow

3 add test cases for tenant update flow


cypress/integration/deleteTenant.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+    cy.wait(4000)
+    cy.visit('/ui/my-account/user_management#tenants')
+    cy.wait(10000)
+  })
+  it('should be able to find tenant', () => {
+    cy.search(config.tenantData.name)
+    cy.getByTestId('name').should('contain', config.tenantData.name)
+  })
+  it('should be able to delete tenant', () => {
+    cy.getByTestId('check-box-cluster-selection').click()
```
jruddell, line 18: what tenant are you trying to delete here? maybe select the row by finding the tenant name... reason being there are some tenants that if deleted will make the app not usable (`service` for instance)

hgupta2363: here I am deleting the one which is be created during tenant creation workflow and to find this I am using search bar
In line 14  I am using search function which is defined in commonutils and this function is helping me to search this particular tenant

jruddell:
make sure you reference the associated JIRA ticket to this work so the tickets are automated

jruddell (approved):

PR #1227, opened 2021-09-15
UX-1102/Add Hostpath CSI Driver to Storage Providers page

knnguy:
Screen Shot 2021-09-16 at 2 07 25 PM

Screen Shot 2021-09-16 at 2 12 47 PM


src/app/plugins/kubernetes/components/app-catalog/selectors.ts
```diff
@@ -30,3 +30,17 @@ export const makeAppsSelector = (
...
 }
+
+export const appDetailsSelector = createSelector(
+  [getDataSelector<DataKeys.AppDetails>(DataKeys.AppDetails, ['name', 'repository'])],
+  (appDetails) => {
+    return appDetails.map((appDetail) => {
+      return {
+        ...appDetail,
+        version: appDetail?.metadata?.version,
+        repository: appDetail?.repository,
```
jruddell, line 41: is this needed?

src/app/plugins/kubernetes/components/storage/storage-providers/list-page.tsx
```diff
@@ -1,40 +1,110 @@
...
+  const { history } = useReactRouter()
+  const { params, getParamsUpdater } = useParams(defaultParams)
+  const [storageApps, setStorageApps] = useState([])
+  const [clusters] = useDataLoader(clusterActions.list)
+  const [repositories, loadingRepositories] = useDataLoader(repositoryActions.list)
+  const { data: apps, reload, loading, message } = useDataLoaderWithMessage({
+    loaderFn: appActions.list,
+    dependentCacheKeys: [clusterActions.cacheKey, repositoryActions.cacheKey],
+  })
+  const anyRepositoryActions = repositoryActions as any
```
jruddell, line 42: why do you need to cast this to any type? can I help you fix the typing issue?

src/app/plugins/kubernetes/components/storage/storage-providers/list-page.tsx
```diff
@@ -1,40 +1,109 @@
...
+    setStorageApps(
+      apps.filter(
+        ({ name, repository }) =>
+          repository === pf9DefaultRepository.name && storageAppNames.has(name),
+      ),
+    )
+  }, [apps])
+
+  const handleOnClick = (app) => {
+    history.push(routes.apps.deploy.path({ name: app.name, repository: app.repository }))
```
jruddell, line 81: Confirm with Chris, but I think when the user clicks deploy on the card is when we should add the pf9 repository. I don't think we want to add the repository based on page render

knnguy: Then how do I know what storage provider apps are available? I'm assuming hard-coding is not an option. @chrisj-platform9

src/app/plugins/kubernetes/components/storage/storage-providers/list-page.tsx
knnguy: Yeah, I need help on this. It's because `repositoryActions.addClustersToRepository` is part of the custom functions in `createCrudActions` and so Typescript says the function doesn't exist. I've seen that in other places we just cast the type as any

jruddell: ah, we can leave this for now then. Were going to move this to the new action set soon and that will resolve the issue

src/app/plugins/kubernetes/components/app-catalog/deployed-apps/deploy-app-page.tsx
```diff
@@ -227,7 +234,7 @@ const DeployAppPage = () => {
           {({ wizardContext, setWizardContext, onNext }) => (
             <WizardMeta
               className={classes.deployAppForm}
-              fields={wizardContext}
+              fields={{ ...appDetail, wizardContext }}
```
jruddell, line 237: you need to extend wizardContext here too

jruddell (approved):

PR #1228, opened 2021-09-16
UX-1223/Add tests for user section

hgupta2363:
1 test cases for create new user

2 test cases for edit user

3 test cases for delete user


cypress/integration/deleteUser.spec.ts
```diff
@@ -0,0 +1,31 @@
+import 'cypress-wait-until'
+import { loadConfig } from 'cypress/support/commands'
+import { Config } from 'cypress/support/types'
+import generateTestId from 'utils/test-helpers'
+describe('delete tenant', () => {
```
jruddell, line 5: The names of your tests here say `tenant` but they are about a `user`.

cypress/integration/editUser.spec.ts
```diff
@@ -0,0 +1,38 @@
+import 'cypress-wait-until'
+import { loadConfig } from 'cypress/support/commands'
+import { Config } from 'cypress/support/types'
+import { contains } from 'cypress/types/jquery'
+describe('delete tenant', () => {
```
jruddell, line 5: tenant instead of user

jruddell (changes requested):
Fix the test names please

cypress/integration/deleteUser.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+  var config: Config
+  before(() => {
+    config = loadConfig()
+    cy.login()
+    cy.wait(4000)
+    cy.visit('/ui/my-account/user_management#users')
```
jruddell, line 11: if its possible to use our routes for this that would be preferred

cypress/integration/deleteUser.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+  })
+  it('should be able to find user', () => {
+    cy.search(config.loginCredentials.username)
+    cy.getByTestId('username').should('contain', config.loginCredentials.username)
+  })
+  it('should be able to delete user', () => {
+    cy.getByTestId('check-box-cluster-selection').click()
+    cy.getByTestId('delete').should('be.visible')
+    cy.getByTestId('delete').click()
+    cy.getByTestId('delete-title')
```
jruddell, line 22: I don't like `delete-title`, can you make this be something more specific as to what it is deleting?

src/app/plugins/account/components/userManagement/users/delete-user-dialog.tsx
```diff
@@ -63,7 +64,7 @@ const UserDeleteDialog: React.FC<Props> = ({ rows: [user], onClose }) => {
 
   return (
     <Dialog open onClose={onClose} onClick={stopPropagation} classes={{ paper: classes.dialog }}>
-      <DialogTitle>
+      <DialogTitle data-testid={generateTestId('delete', 'title')}>
```
jruddell, line 67: maybe add `user` in the test-id?

jruddell (changes requested):
Few things to tweak. Check the titles of your test suites. I see a lot of `describe('tenant ...` inside user specs.

cypress.example.json
```diff
@@ -36,6 +40,10 @@
     "scaleMasterNode.spec.ts",
     "tenantCreation.spec.ts",
     "editTenant.spec.ts",
-    "deleteTenant.spec.ts"
-  ]
+    "deleteTenant.spec.ts",
```
jruddell, line 43: this is an example file, you probably shouldn't be making changes here

src/app/plugins/account/components/userManagement/users/EditUserPage.js
```diff
@@ -73,7 +73,7 @@ const TogglableTextField = ({
           disabled
         />
       )}
-      <SimpleLink className={classes.togglableFieldBtn} onClick={toggleField}>
+      <SimpleLink className={classes.togglableFieldBtn} onClick={toggleField} label={label}>
```
jruddell, line 76: why do we need to add a prop to SimpleLink here? can't you just use the children?

hgupta2363: children is same for all fields like usename,password so if we use only children to generate testid then multiple fields will have same testid

jruddell: children is the text content for SimpleLink. what makes label unique? SimpleLink is used in many places in the application, are we going to add a label to each time we render it?

hgupta2363: label has more specific info about textfield children value is "change" which is same for each textfield and I don't think we need to pass label prop in every SimpleLink render because here only multiple textfield is sharing single simpleLink render

cypress/integration/users/createUser.spec.ts.ts
```diff
@@ -0,0 +1,41 @@
...
+    cy.getByTestId('tenants-and-roles')
+      .find('p')
+      .should('contain', 'Tenants and Roles')
+    cy.getByTestId('check-box-cluster-selection')
+      .last()
+      .click()
+    cy.getByTestId('arrow-next').click()
+  })
+  it('should be able to verify user creation completed succesfully', () => {
+    cy.wait(4000)
```
kingshuk-nandy, line 36: Same as above. Try avoid using hardcoded wait

PR #1233, opened 2021-09-17
added some missing default values

jruddell:


knnguy (approved):

xantorres:
Fixed some problems regarding the syncing between the Wizard and the yaml editor

PR #1246, opened 2021-09-24
UX-1166,UX-1176,UX-1179,UX-1183/tests for rbac section

Prathameshs:
Scripts for RBAC section
rbac_suite_report_screenshot


cypress/integration/rbac/add-cluster-role.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    const clusterSelector: SelectorDetails = {
+      id: 'clusterId',
+      dataTestId: 'clusterid-value',
+      itemToBeSelected: config.singleMasterClusterData.clusterName,
+    }
+    cy.selectFromDropdown(clusterSelector)
+
+    cy.intercept('**/apis').as('getApis')
+    cy.intercept('**/v1').as('getV1')
```
jruddell, line 48: what api is this capturing? this is excessively vague... if we have more than one api with /v1 what happens? Can we be a bit more explicit with these intercepts? instead of `**/v1` or `**/apis` can it be a more explicit path that denotes what item were working with?

kingshuk-nandy: Updated the Code. Have given more explicit path for interception

PR #1247, opened 2021-09-27
Integrated Cypress Teamcity Reporter

kingshuk-nandy:


package.json
```diff
@@ -19,6 +19,7 @@
     "concurrently": "^4.1.1",
     "cors": "^2.8.4",
     "cypress-mochawesome-reporter": "^2.1.1",
+    "cypress-teamcity-reporter": "^3.0.0",
```
jruddell, line 22: I have a few concerns with this. 
1. This package looks like its more for mocha than cypress, what kind of subdependencies is it bringing in? like is it bringing in mocha?
2. Why is this in the dependencies section? this is specific to testing and as such should be a dev dependency. 
3. What is the purpose of bringing in this package? is it to report live to a github PR the status of the tests?
4. I don't see you specifying anywhere how we configure / use this reporter. I.e - on the NPM docs they have this example (`cypress run --reporter cypress-teamcity-reporter`) where are we doing this?

kingshuk-nandy: Query 1:
cypress-teamcity-reporter is a Teamcity reporter which makes it possible to display test results in real-time, makes test information available on the Tests tab of the Build Results page. It is based on the package @cypress/mocha-teamcity-reporter and was improved with new informations and support to the newer versions of mocha. Refer to the links, https://www.npmjs.com/package/cypress-teamcity-reporter , https://docs.cypress.io/guides/tooling/reporters#Custom-reporter
Query 2:
Not sure. If we put this in dev dependency, Will the dependencies get installed during the TeamCity run? If yes then we will update it in the dev dependencies

jruddell (approved):

PR #1249, opened 2021-09-28
UX-1249/Add container runtime to cluster creation wizard

kevinylee1991:
Screen Shot 2021-09-27 at 6 11 00 PM
Screen Shot 2021-09-27 at 6 11 20 PM
Screen Shot 2021-09-27 at 6 11 27 PM
Screen Shot 2021-09-27 at 6 12 01 PM


src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/container-runtime-picklist.tsx
```diff
@@ -0,0 +1,76 @@
...
+      }
+      if (compareVersions(kubeRoleVersion, '1.21') < 0) {
+        onChange('docker')
+        return false
+      }
+      if (!value) {
+        onChange('containerd')
+      }
+      return true
+    }, [kubeRoleVersion, value])
```
jruddell, line 58: We shouldn't need to check value changes... Maybe disable the field if the version< 1.21 to avoid this

jruddell (approved):

PR #1251, opened 2021-09-28
UX-1250/pushed test cases for storage class creation and verify

rajshekhar23:
Storage class creation and verify.
Screenshot 2021-10-06 at 8 00 09 PM
Screenshot 2021-10-06 at 9 18 01 PM
Screenshot 2021-10-06 at 9 29 32 PM


cypress/config/storage_suite.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../cypress.json",
+  "testFiles": [
+    "dashboard/login.spec.ts",
+    "infrastructure/single-master-cluster-create.spec.ts",
+    "infrastructure/scale-worker.spec.ts",
```
jruddell, line 6: do you need to scale workers for a storage class test?

cypress/integration/storage/csi-drivers-list.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+  before(() => {
+    config = loadConfig()
+    cy.login()
+  })
+
+  it('should be able to navigate to csi drivers tab', () => {
+    cy.getByTestId('storage', 9000).click()
+    cy.getByTestId('csi-drivers-tab', 5000).click()
+
+    cy.intercept('**/csidrivers').as('csidrivers')
```
jruddell, line 20: lets name this better `.as('getCSIDrivers')`. I think the convention we are using here is to append the verb associated with the request... aka a GET request we append `get`.

cypress/integration/storage/csi-drivers-list.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    cy.wait('@csidrivers', { timeout: 60000 }).then((res) => {
+      assert.equal(res.response.statusCode, 200)
+    })
+    cy.getByTestId('clustername').should('contain.text', config.singleMasterClusterData.clusterName)
+  })
+
+  it('should be able to verify that CSI Drivers with required cluster name exist', () => {
+    cy.getByTestId('check-box-cluster-selection', 30000).should('be.visible')
+
+    cy.get('#mui-component-select-Cluster', { timeout: 8000 }).click()
```
jruddell, line 31: why is there casing differences here? the test id helper lowercases everything if im not mistaken

cypress/integration/storage/storage-list.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  before(() => {
+    config = loadConfig()
+    cy.login()
+  })
+
+  it('should be able to navigate to storage class', () => {
+    cy.getByTestId('storage', 9000).click()
+    cy.getByTestId('storage-classes-tab', 5000).click()
+
+    cy.intercept('**/storageclasses').as('storageclasses')
```
jruddell, line 20: follow the naming convention here as well please

cypress/config/storage_suite.json
rajshekhar23: No need. will remove it.

cypress/integration/storage/csi-drivers-list.spec.ts
rajshekhar23: Updated!

cypress/integration/storage/storage-list.spec.ts
rajshekhar23: Updated!

src/app/plugins/kubernetes/components/common/ClusterPicklist.js
```diff
@@ -112,7 +112,7 @@ ClusterPicklist.propTypes = {
 ClusterPicklist.defaultProps = {
   ...Picklist.defaultProps,
   name: 'clusterId',
-  label: 'Cluster',
+  label: 'cluster',
```
jruddell, line 115: lets not make this lowercase here, rather add a test id to the picklist. The helper method I wrote for generating test ids lowercases and adds a `-`.

`data-testid={generateTestId(rest.label, 'picklist-component')`

rajshekhar23: updated.!

cypress/integration/storage/csi-drivers-list.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+  before(() => {
+    config = loadConfig()
+    cy.login()
+  })
+
+  it('should be able to navigate to csi drivers tab', () => {
+    cy.getByTestId('storage', 9000).click()
+    cy.getByTestId('csi-drivers-tab', 5000).click()
+
+    cy.intercept('**/csidrivers').as('getCSIDrivers')
```
jruddell, line 20: we have a better way of intercepting requests now, please update this to follow the new pattern

PR #1258, opened 2021-10-01
Cypress multi reporter feature

Prathameshs:
Added cypress multi reporter feature.


package.json
```diff
@@ -20,6 +20,7 @@
     "cors": "^2.8.4",
     "cypress-mochawesome-reporter": "^2.1.1",
     "cypress-teamcity-reporter": "^3.0.0",
+    "cypress-multi-reporters": "^1.5.0",
```
jruddell, line 23: there shouldn't be any testing packages as a part of the dependencies.. this ends up getting pulled into our production bundle that we ship to customers.

kingshuk-nandy: Updated the Package.json

PR #1280, opened 2021-10-14
UX-1144, UX-1208/Added drift analysis list page, as well as analysis …

kevinylee1991:
…import and export functionality

Screen Shot 2021-10-13 at 4 41 12 PM
Screen Shot 2021-10-13 at 6 22 06 PM
Screen Shot 2021-10-13 at 6 22 23 PM
Screen Shot 2021-10-13 at 6 24 26 PM


src/app/plugins/kubernetes/components/rbac/profiles/drift/import-drift-analysis-page.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const [analysis, setAnalysis] = useState('')
+  const [clusterId, setClusterId] = useState('')
+  const [profileName, setProfileName] = useState('')
+  const classes = useStyles({})
+
+  const [clusters] = useDataLoader(clusterActions.list)
+  const clusterName = findClusterName(clusters, clusterId)
+
+  const onDrop = useCallback((acceptedFiles) => {
+    const file = acceptedFiles[0]
```
jruddell, line 101: should we validate a file type here? what is the format we should be accepting here?

PR #1283, opened 2021-10-15
UX-1262/Include container runtime in cluster edit & upgrade wizards

kevinylee1991:
Waiting on pushkar to respond about API about whether a cluster with containerd currently that is upgraded and sent containerd as runtime in the upgrade request body will cause any unnecessary processing on the backend.

Screen Shot 2021-10-14 at 6 11 38 PM
Screen Shot 2021-10-14 at 6 15 01 PM


src/app/plugins/kubernetes/components/infrastructure/clusters/UpgradeClusterPage.tsx
```diff
@@ -262,6 +265,24 @@ const UpgradeClusterForm = ({ cluster, onFormSubmit }) => {
                           )}
                         </FormFieldCard>
                       )}
+                      {!isWorking && cluster.containerRuntime !== 'containerd' && (
```
jruddell, line 268: probably best to include a compare versions check for the future case when the cluster version is >= 1.23. What do you think?

kevinylee1991: I think perhaps it could stay, but instead of saying will be deprecated, say is deprecated in the description? Since we can't control for sure whether or not they will choose to change it to containerd

PR #1288, opened 2021-10-19
First batch of initial loading changes and refactoring

xantorres:


jruddell:
looking a lot better already! 👍

PR #1316, opened 2021-10-28
UX-872/Only allow user to import clusters under the service tenant and show API error messages in the import forms

knnguy:
### Only users under the service tenant can click into the import forms
Screen Shot 2021-10-28 at 2 55 56 PM

### Also added a check where user can't proceed to the next step unless they have selected a cluster

Screen Shot 2021-10-29 at 3 01 00 PM

### If the user clicks on a region and the API request for that regions results in an error, the checkbox for that region will automatically uncheck. If we don't automatically uncheck it for the user, when they select another region and the UI attempts to discover clusters, that request will fail as well because it includes the first failed region in the request body

https://user-images.githubusercontent.com/23369276/139506226-06bb63c3-62f2-46e5-af53-64304b713a62.mov

Screen Shot 2021-10-29 at 3 23 15 PM


jruddell (approved):
🚢

jruddell (approved):

PR #1322, opened 2021-11-02
UX-1288/Fix issue where screen flickers between onboarding flow and dashboard page

knnguy:
### Issue:
When the user deletes all their clusters and all their users, clears cache, and then logs back in, the screen flickers from the onboarding flow to the dashboard. 

Also, I found out that when the user finishes the onboarding flow, the onboarding flag was being backfilled in a `useEffect` in `AuthenticatedContainer.js` instead of through a `handleFormCompletion` function in the onboarding wizard (onboarding-page.tsx). Works either way, but I fixed it so that it's going through the correct function

### Fix:
We have to wait for the region features, the global feature flags, clusters, and users to finish loading before we decide to show the onboarding flow or not. At the very least, we need to wait for the region features to finish loading. If the onboarding env conditions are not met, then no need to wait for other data to load

**Scenario: User has 0 clusters, 0 users, cache is cleared, and user attempts to log back in**
https://user-images.githubusercontent.com/23369276/139948849-fa3599df-b3c9-4a74-ac3d-1ad0897b99fb.mov


jruddell (approved):
Sync with Xan on this, he's been reworking this component so you're going to need to work on updating this in the refactor

PR #1328, opened 2021-11-09
UX-1305/Track each tab on workload & storage detail pages

kevinylee1991:


src/app/core/utils/routes.ts
```diff
@@ -597,15 +597,35 @@ export const routes = {
       },
       name: 'Pods:Add',
     }),
-    details: Route.register<{ clusterId: string; id: string; tab: IPodDetailsPageTabs }>({
-      url: `${pluginRoutePrefix}/workloads/clusters/:clusterId/pods/:id#:tab`,
+    overview: Route.register({
+      url: `${pluginRoutePrefix}/workloads/clusters/:clusterId/pods/:id#${IPodDetailsPageTabs.Overview}`,
```
jruddell, line 601: What is the reason for changing this from a dynamic tab?

kevinylee1991: This is so that the analytics.page function will track each individual tab as a different page (the analytics.page function uses the route.name property)

PR #1329, opened 2021-11-09
UX-1189, UX-1214/tests for App section

Prathameshs:
deploy-infra-repository
)


Prathameshs:
edit_deployed_app

cypress/integration/apps/edit-deployed-app.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+     cy.intercept('**/releases').as('getReleases')
+     cy.intercept('**/releases/*').as('getRelease')
+     cy.intercept('**/clusters').as('getClusters')
+     cy.intercept('**/pods').as('getPods')
+     cy.intercept('**/namespaces').as('getNamespaces')
+   })
+
+   it('should be able to navigate deployed apps', () => {
+
+     const apiCalls = ['getCharts','1', 'getReleases']
```
jruddell, line 28: what is the purpose of the `1` here?

Prathameshs: We have dividing the API calls in 2 sections, main section and sub-section. so we are using the 1 to divide the API calls between 2 sections.

jruddell: I am not a fan of this pattern. Why do you need to divide them? and how does a `1` accomplish this divide? The reason why I dont like it is it requires some form of implicit knowledge of how this should work verses some explicit action that is very clear what it is doing.

Prathameshs: Resolved.

jruddell (approved):

PR #1330, opened 2021-11-09
UX-1349/Disable editing published profiles, add note for editing prof…

kevinylee1991:
…ile resources, disable deploy profile button for ineligible clusters

Screen Shot 2021-11-09 at 11 52 22 AM
Screen Shot 2021-11-09 at 11 54 02 AM
Screen Shot 2021-11-09 at 11 54 53 AM


src/app/plugins/kubernetes/components/rbac/profiles/deploy/deploy-rbac-profile-page.tsx
```diff
@@ -159,11 +161,23 @@ const renderCustomNext = (handleNext, wizardContext, activeStep) => (
...
 )
 
+const validateCluster = ({ clusters }) => {
+  if (!clusters) {
+    return false
+  }
+  if (!clusters.length) {
+    return false
+  }
+  const cluster = clusters[0]
```
jruddell, line 171: This seems error prone... What is the reason to only look at the first cluster?

kevinylee1991: Only one cluster can be selected, but the picklist/radio output is always in an array. So I just use clusters (plural) to make it more clear that the picklist is returning an array

jruddell (approved):

PR #1341, opened 2021-11-12
UX-1322/Fix issue where setting default cluster/namespace in one brow…

kevinylee1991:
…ser does not affect the other browser

Problem was that if the default cluster/namespace was set on a particular browser in the redux store, it would not try to update ever unless editing it in that same browser.


jruddell (changes requested):
I dont think we want to go this route. maybe we should look at invalidating the preferences on logout / login

PR #1345, opened 2021-11-16
UX-1307/Added tracking events to RBAC profile pages & wizard

kevinylee1991:


src/app/plugins/kubernetes/components/infrastructure/clusters/tracking.ts
```diff
@@ -318,3 +318,68 @@ export const bareOSClusterTracking = {
     }),
   }),
 }
+
+export const rbacProfileCreationTracking = {
```
jruddell, line 322: can we follow this pattern and create / expose these trackers from a `tracking` file inside the `RBAC` folder?

jruddell (approved):

src/app/plugins/kubernetes/components/infrastructure/clusters/tracking.ts
kevinylee1991: oh oops I thought that was the base tracking.js file for some reason, yup lemme move that

PR #1358, opened 2021-11-22
UX-1316/Add customer tier label for all tiers in user menu

knnguy:
Before, only enterprise users had the ENTERPRISE label in the user menu. Now we add it for all tiers

Screen Shot 2021-11-22 at 1 49 04 PM

Screen Shot 2021-11-22 at 1 48 28 PM


src/app/core/components/UserMenu.tsx
```diff
@@ -97,6 +104,16 @@ class MenuListItem extends React.PureComponent<any> {
...
 }
 
+const getCustomerTierClassName = (customerTier) =>
+  objSwitchCase(
+    {
+      [CustomerTiers.Enterprise]: 'enterpriseLabel',
+      [CustomerTiers.Freedom]: 'freedomLabel',
+      [CustomerTiers.Growth]: 'growthLabel',
+    },
+    'enterpriseLabel',
```
kevinylee1991, line 114: may be better to have freedomLabel as the default value

kevinylee1991 (approved):
lgtm

PR #1377, opened 2021-12-07
UX-1359/Added ideas card to help page

kevinylee1991:
Before (spacing):
Screen Shot 2021-12-06 at 4 27 13 PM
After (spacing):
Screen Shot 2021-12-06 at 4 29 09 PM


chrisj-platform9:
@kevinylee1991 we should move the Support Ticket card to the top row

PR #1389, opened 2021-12-13
UX-1159/Add DNS Autoscaler addon to Edit Cluster page

knnguy:
For newer versioned clusters (1.21 and up), the dns autoscaler addon option is added to the coreDNS addon fields card in the edit cluster page

Screen Shot 2021-12-13 at 11 59 40 AM
Screen Shot 2021-12-13 at 11 59 30 AM


src/app/plugins/kubernetes/components/infrastructure/clusters/helpers.ts
```diff
@@ -102,6 +102,7 @@ export const getAddonVersions = async (clusterId) => {
       casaws: '1.14.7',
       casazure: '1.13.8',
       dashboard: '2.0.1',
+      dnsautoscaler: '1.8.1',
```
jruddell, line 105: Can we remove these hardcoded versions now that there is a way to load the versions for an addon?

knnguy: We are getting the cluster addon versions from an api in the `loadAddonVersions` function.

I just checked and realized that this function, `getAddonVersions` isn't even used anywhere. I have removed it from the code

jruddell (approved):

PR #1408, opened 2021-12-21
Create list-container component as a replacement for current CRUD helpers

xantorres:


src/app/plugins/kubernetes/components/storage/storage-classes-picklist.tsx
```diff
@@ -32,10 +35,16 @@ const StorageClassesPicklist: React.ComponentType<Props> = forwardRef<HTMLElemen
       showNone = true,
       ...rest
     } = props
-
-    const [storageClasses, storageClassesLoading] = useDataLoader(storageClassActions.list, {
-      clusterId,
-    }) as any
+    const params = useMemo(
```
jruddell, line 38: do we still need to `useMemo` on params? I thought we addressed this internally with memoization

src/app/core/containers/list-container.tsx
```diff
@@ -0,0 +1,285 @@
...
+      {!DeleteDialog && handleRemove && showingConfirmDialog && (
+        <ConfirmationDialog
+          open={showingConfirmDialog}
+          text={deleteConfirmText}
+          onCancel={toggleConfirmDialog}
+          onConfirm={handleDeleteConfirm}
+        />
+      )}
+      {addEnabled && (
+        <PageContainerHeader>
```
jruddell, line 209: We should either update the PageContainer to integrate with the portals or just remove PageContainer completely. If we integrate then we should be really consistent with using this component to update header elements in the page

src/app/plugins/kubernetes/components/storage/storage-classes/list-page.tsx
```diff
@@ -77,12 +42,35 @@ export const options = {
...
-  cacheKey: ActionDataKeys.StorageClasses,
-  actions: storageClassActions,
   name: 'StorageClasses',
   title: 'Storage Classes',
-  ListPage,
 }
 
-const components = createCRUDComponents(options)
-export default components.ListPage
+const ListPage = () => {
```
jruddell, line 49: 😍

jruddell (approved):

src/app/plugins/kubernetes/components/storage/storage-classes-picklist.tsx
xantorres: no, we don't need that anymore

PR #1412, opened 2022-01-05
UX-1449/Tabs Component

jruddell:
image

TODO: update all references to use the new tabs component


src/app/core/components/Navbar.js
```diff
@@ -28,6 +28,10 @@ import { memoize } from 'utils/misc'
 import generateTestId from 'utils/test-helpers'
 import SimpleLink from './SimpleLink'
 
+/**
+ * @deprecated Use core/elements/sidebar instead
+ */
+
```
xantorres, line 34: can we remove this Navbar if is not being used anymore?

src/app/core/elements/tabs/tab.tsx
```diff
@@ -0,0 +1,22 @@
...
+  useEffect(() => {
+    const { addTab, value, label } = props
+    addTab({ value, label })
+  }, [])
+
+  const { activeTab, value, children } = props
+  if (value !== activeTab) {
+    return null
+  }
+  return <>{children}</>
```
xantorres, line 19: You can just `return children` here

xantorres (approved):
Looks good to me

src/app/core/components/Navbar.js
jruddell: Yes, theres a number of deprecated files that have been marked like this. Will definitely clean them all up

PR #1433, opened 2022-01-11
UX-1417/Created list page for StatefulSets

knnguy:
Screen Shot 2022-01-11 at 10 26 49 AM

Also fixed a bug with the `useListAction` function that was used in the storage class list page. I noticed that in the storage class list page, when you set cluster to all, the data is not refetched


src/app/plugins/kubernetes/components/stateful-sets/actions.ts
```diff
@@ -0,0 +1,31 @@
...
+import { someAsync } from 'utils/async'
+import { flatten, pluck } from 'ramda'
+
+const { qbert } = ApiClient.getInstance()
+
+const statefulSetActions = new ActionsSet<'StatefulSets'>({
+  uniqueIdentifier: 'metadata.uid',
+  indexBy: 'clusterId',
+  entityName: 'StatefulSets',
+  cacheKey: 'StatefulSets',
```
jruddell, line 16: shouldn't this be a DataKey?

src/app/plugins/kubernetes/components/stateful-sets/actions.ts
```diff
@@ -0,0 +1,31 @@
...
+})
+
+export default statefulSetActions
+
+export const listStatefulSets = statefulSetActions.add(
+  new ListAction<'StatefulSets', { clusterId: string }>(async (params) => {
+    Bugsnag.leaveBreadcrumb('Attempting to get StatefulSets', params)
+    console.log('clusterId', params)
```
jruddell, line 24: console.log

jruddell (approved):

PR #1435, opened 2022-01-11
UX-1415/Updated sidenav for user settings and my profile tab to match…

kevinylee1991:
… new designs
Screen Shot 2022-01-11 at 1 10 21 PM
Screen Shot 2022-01-11 at 1 10 32 PM


chrisj-platform9:
@jruddell @kevinylee1991 When a user is on enterprise we should hide the Growth and Enterprise part of the card. We also need to hide the entire piece when the Tier is OEM. 
just thought of this now, apologies. 

For Free: Show Growth and Enterprise
For Growth Show Enterprise Only
For Enterprise: Show only the top portion
For OEM: Don't show the plan option at all


Thinking ahead, I wonder if we can make the are that shows plan details show content dynamically. And I wonder if there is a way to allow OEMs to add their own content/links. (just thinking)

jruddell:
I would push back on dynamically rendered content until we have a proper storage + delivery method defined and implemented.

kevinylee1991:
Screen Shot 2022-01-11 at 3 56 07 PM

src/app/plugins/kubernetes/components/common/ClusterPicklist.js
```diff
@@ -0,0 +1,132 @@
+import React, { useMemo, useEffect, forwardRef } from 'react'
```
jruddell, line 1: is there a reason why were recreating this component?

src/app/core/components/back-to-dashboard.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
```
jruddell, line 1: it probably makes more sense to update simple link to support an icon (i think it already does) and put the styling for icon with text there as an icon in a link is going to be quite a common use case.

src/app/core/components/back-to-dashboard.tsx
```diff
@@ -0,0 +1,48 @@
...
+import SimpleLink from 'core/components/SimpleLink'
+import FontAwesomeIcon from 'core/components/FontAwesomeIcon'
```
jruddell, line 3: instead of a component that is specifically about redirecting to the dashboard page, maybe it should be a simple wrapper to simple link that just styles it gray / passive?

jruddell (approved):

PR #1442, opened 2022-01-12
UX-1417/Created overview, pods, events, logs, and yaml pages for StatefulSets

knnguy:
- Created all the details pages for StatefulSets

- Also attempted to create the `useUpdateAction` hook. Not fully complete yet. Missing the `onComplete` function handler. Need to ask Xan some questions. But as for now, updating works

### Overview Page

Screen Shot 2022-01-14 at 10 33 25 AM

### Pods List Page

Screen Shot 2022-01-14 at 10 33 52 AM

### Events List Page

Screen Shot 2022-01-12 at 4 22 59 PM

### Logs Page
Screen Shot 2022-01-14 at 10 23 13 AM

### YAML Page
Screen Shot 2022-01-14 at 10 23 20 AM


knnguy:
@chrisj-platform9 @jruddell 

What is this System Prometheus field? I can't find it in the pod templates.

Screen Shot 2022-01-12 at 11 25 27 AM

src/app/core/themes/serenity/serenity-ultra-dark.ts
```diff
@@ -117,6 +117,7 @@ const components = {
     primary: colors.blue[500],
     success: colors.green[500],
     warning: colors.yellow[500],
+    pending: colors.grey[300],
```
jruddell, line 120: should this be `unknown`? a pending name could mean yellow just as well as gray. but mapping unknown to pending more applicable

src/app/plugins/kubernetes/components/common/entity/cards-header.tsx
```diff
@@ -0,0 +1,24 @@
...
+  cards: {
+    display: 'grid',
+    gridAutoFlow: 'column',
+    gridAutoColumns: 'max-content',
+    gridGap: theme.spacing(1),
+    padding: '0 0 66px',
+  },
+}))
+
+const CardsHeader = ({ children }) => {
```
jruddell, line 15: we have a card header component already, can we reuse that component instead of this one

src/app/plugins/kubernetes/components/common/entity/pods-status-card.tsx
```diff
@@ -0,0 +1,74 @@
...
+  },
+  mainDiv: {
+    display: 'grid',
+    gridAutoFlow: 'column',
+  },
+}))
+
+const quantityFn = (pods, statusCounts) => ({
+  quantity: pods?.length,
+  pieData: [
```
jruddell, line 20: shouldn't we be using the current definition for the pods graph? if the colors are wrong we can update the mapping there.

src/app/plugins/kubernetes/components/pods/metrics/actions.ts
```diff
@@ -8,9 +8,35 @@ import namespaceActions from 'k8s/components/namespaces/actions'
 import createContextLoader from 'core/helpers/createContextLoader'
 import { INamespaceSelector } from 'k8s/components/namespaces/model'
 import { podMetricsByClusterIdAndNamespaceSelector } from './selectors'
+import Bugsnag from '@bugsnag/js'
+import ActionsSet from 'core/actions/ActionsSet'
+import ListAction from 'core/actions/ListAction'
 
 const { qbert } = ApiClient.getInstance()
 
+const podMetricsActionSet = new ActionsSet<DataKeys.PodMetrics>({
```
jruddell, line 17: are we using new-actions or actions? this is in two places

knnguy: It's not in two places. This is the list action for pod metrics. The one in pods/new-actitons.ts is the list action for pods

src/app/plugins/kubernetes/components/common/entity/pods-status-card.tsx
knnguy: Yep. Didn't know pods had one

knnguy: Fixed it. Modified it a little. Added 'Succeeded' as a status and made it blue. Also, I think "Pending" makes more sense than "Waiting". That's the actual pod status

Screen Shot 2022-01-12 at 4 17 50 PM

src/app/plugins/kubernetes/components/pods/metrics/actions.ts
jruddell: Sorry my comment wasn't clear, I mean we should put new list actions in the new actions file, keep them separate so dropping one is easier

src/app/plugins/kubernetes/components/common/entity/pods-status-card.tsx
jruddell: I want to avoid adding excess color mappings, `default` is blue already so just map succeeded to the default color variable. I'm ok adding unknown as thats generic and would constitute a new color variable mapping

knnguy: Yep, the only new color variable I added was `unknown`.

src/app/plugins/kubernetes/components/common/entity/cards-header.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const classes = useStyles()
+  return (
+    <header>
+      <article className={classes.cards}>{children}</article>
```
jruddell, line 19: a card should be an article, this is the container around cards so a section or a div is more appropriate. Each card that is rendered as a child inside here should be an article

jruddell (approved):

PR #1444, opened 2022-01-14
UX-1415/Preserve selected user theme in store, and interact properly with global theme settings

kevinylee1991:
Added a little text about the global theme setting in the theme selection card area, any ideas on how we want to display that to the user?

Also added a disabled cursor over the cards when a global theme is set.

Removed an e.stopPropagation after encapsulating a function in a useCallback, it didn't seem to be doing anything.

Screen Shot 2022-01-13 at 5 47 15 PM
Screen Shot 2022-01-13 at 5 47 32 PM
Screen Shot 2022-01-13 at 5 48 09 PM
Screen Shot 2022-01-13 at 5 48 40 PM


jruddell:
Lets not add another card for the custom theme choice, I'd only like to push / present the ones we designed

kevinylee1991:
Removed Custom as a choice in the cards (will show Default as checked) but kept it as-is in the implementation.

jruddell (approved):

PR #1448, opened 2022-01-19
UX-1415/Update logo upload feature

kevinylee1991:
One question regarding the login screen, see last screenshot.

Do we want to have an uploaded logo / logo URL influence what is shown there? Since we're looking to provide them the option to change the logo specifically for the header, perhaps the code that changes the logo on the login page needs to be removed?

Also I forgot that I could look at all values on the validator until just now, so i'm going to update the image drop component & validator tomorrow morning, right now it's kind of awkward how the values are being set.

Screen Shot 2022-01-18 at 7 53 22 PM
Screen Shot 2022-01-18 at 7 59 54 PM
Screen Shot 2022-01-18 at 8 00 13 PM
Screen Shot 2022-01-18 at 8 04 10 PM
Screen Shot 2022-01-18 at 8 05 09 PM


kevinylee1991:
Screen Shot 2022-01-19 at 12 17 37 PM

src/app/core/elements/tools/image-drop.tsx
```diff
@@ -0,0 +1,138 @@
+import React, { forwardRef, useCallback } from 'react'
```
jruddell, line 1: not sure if a tools folder makes sense... lets just have the ImageDrop file in elements

jruddell (approved):

PR #1450, opened 2022-01-19
UX-1533/Remove openstack and metalstack plugins

kevinylee1991:
Some other things we could do could be to remove deprecated components as well, such as the Toolbar if we don't want to keep them around. There are also some old references to things like intercom that are no longer used and I think some unused helper functions.


jruddell:
Lets also remove support for linking back to clarity, the link generation... etc. We don't have a use case to support switching between apps anymore! 🎉  🎊

jruddell (approved):

PR #1455, opened 2022-01-20
UX-1524/ fix eslint issues

patil-pratik-87:
- Updated linting plugins and configs in .eslintrc.
- Updated specific rules  in .eslintrc.
- Updated the formatting with prettier (eslint and standalone)
- Fixed the pre-commit hook to format and lint successfully before committing.

Note -->
- Still, we have a lot of warnings in linting. Some of them might be legitimate errors. Feel free to change the rules to throw errors and the issues if you feel so. 
- I suggest you turn on the Eslint extension of the vscode and add 
` "editor.codeActionsOnSave": {
    "source.organizeImports": false,
    "source.fixAll.eslint": true
  },
  "editor.formatOnSave": true,` in your settings.json in vscode. 
If this does not work Enable the "Format On Save' checkbox in Prefereces>Settings
Screenshot 2022-01-20 at 6 12 04 PM

- I surely might have made some errors while fixing the issues. Let me know what I broke :) and I will fix it.
- Let me know if you find something weird in the application :)

- I see the compilation, lining and formatting green at this stage.


.eslintrc
```diff
@@ -12,17 +12,21 @@
...
-    "standard-with-typescript",
-    "standard-react",
-    "prettier",
-    "prettier/react",
-    "prettier/@typescript-eslint"
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:react/recommended",
+    "prettier"
+    // "standard",
```
xantorres, line 21: We can remove these lines, no need to keep outdated stuff

src/app/bootstrap.ts
```diff
@@ -1,41 +1,40 @@
...
 interface IWindow extends Window {
   customLogger: (msg: string) => void
   process: any
 }
-declare var window: IWindow
+declare let window: IWindow
 
 const myCustomLogger = {
-  debug: function() {
+  debug: function () {
```
xantorres, line 22: Maybe is just me, I don't like these spaces before parenthesis. Can we disable `space-before-function-paren` ?

.eslintrc
patil-pratik-87: sure.
I kept is for the explanation of why the configs were removed. Will remove them

src/app/bootstrap.ts
patil-pratik-87: It is off in eslint, It may be coming for prettier defaults. I am okay anyways as long as we are consistent.

patil-pratik-87:
> What about the `.prettierrc.yaml` file we have? Is it taken into consideration by the eslint plugin? Would it be possible to just get rid of that file and make prettier use the eslint config instead ? If not, let's make sure eslint is accounting for that as well.
> 
> By the way, I suggest using a line width of 80 instead of 100, as they are used in most projects. It just makes it easier to split the code comfortably in 3 panels when working on large screens, and 2 on smaller screens.

yeah `.prettierrc.yaml` is picked up. 
 I thought of maybe converting it to JSON for consistency. In eslint, we do not run the --fix command in pre-commit as I though sometimes we might not agree with the fix.
I think the prettier --write picks the config from the JSON, hence keeping it.

Although, I agree maintaining the rules at two places is not a good idea. Let us talk about it.

src/app/bootstrap.ts
jruddell: Yes I am also of the same mind here. The rational for this is to more easily find a function definition by there being a space between the parenthesis in the definition vs no space for when its called. But I don't like this pattern so good with formatting it without the space

jruddell:
> yeah `.prettierrc.yaml` is picked up. I thought of maybe converting it to JSON for consistency. In eslint, we do not run the --fix command in pre-commit as I thought, sometimes we might not agree with the lint fix. We are doing prettier --write pre-commit and it picks the config from the JSON, hence keeping it.
> 
> Although, I agree maintaining the rules at two places is not a good idea. Let us talk about it.

I would prefer we stick with one config as much as possible. If prettier can work off of eslint and the settings we have in the .prettierrc.yaml file can be propagated to eslint then i'm all for it

src/app/core/components/AutocompleteBase.js
```diff
@@ -87,16 +87,8 @@ class AutocompleteBase extends PureComponent {
...
-      className,
-      classes,
-      suggestions,
-      onChange,
-      initialValue,
-      label,
-      id,
-      ...other
-    } = this.props
+    const { className, classes, suggestions, onChange, initialValue, label, id, ...other } =
```
jruddell, line 90: Should we enforce something where more than x number of parameters will automatically render items on new lines? I personally dont like the reformat here where this.props is on a new line

src/app/core/elements/menu/helpers.ts
```diff
@@ -1,13 +1,15 @@
 import { AlignVertical, AlignHorizontal } from './model'
 
-export const getMenuTransform = (scale) => ({ vertAlign, horizAlign }) => {
-  const tLeft = horizAlign === 'middle' ? '50%' : '0%'
-  const tTop = vertAlign === 'middle' ? '-50%' : '0%'
-  if (vertAlign === 'middle' || horizAlign === 'middle') {
-    return `scale(${scale}) translate(${tLeft}, ${tTop})`
+export const getMenuTransform =
+  (scale) =>
```
jruddell, line 4: this looks a bit strange... shouldn't the format put it like

```
export const getMenuTransform = (scale) =>
  ({ vertAlign, horizAlign }) => {
```

src/app/plugins/account/components/userManagement/tenants/UserRolesTableField.js
```diff
@@ -1,13 +1,13 @@
-import withFormContext from 'core/components/validatedForm/withFormContext'
-import React, { useState, useCallback, useMemo } from 'react'
-import { noop, emptyArr } from 'utils/fp'
-import { pluck, pickAll, prop, assoc, partition } from 'ramda'
-import RolesPicklist from 'account/components/userManagement/common/RolesPicklist'
 import { FormControl, FormHelperText } from '@material-ui/core'
-import ListTable from 'core/components/listTable/ListTable'
 import { makeStyles } from '@material-ui/styles'
+import RolesPicklist from 'account/components/userManagement/common/RolesPicklist'
```
jruddell, line 3: Are we reordering imports? I'm not a fan of that.

xantorres: Yeah, also this could potentially break things specially in places like `bootstrap.ts` where the order of things actually matters and changing one import could blow up the app.

patil-pratik-87:
Removed prettier and resolved the comments here --> https://github.com/platform9/pf9-ui-plugin/pull/1455

PR #1456, opened 2022-01-21
Added new Grid component

xantorres:
To-do list
- [x] Implement  base Grid components, interfaces and hooks
- [x] Create Grid stories in the storybook
- [x] Implement filtering
- [x] Implement sorting
- [x] Implement pagination
- [x] Make all the Grid parts optionally controllable (eg async filters, etc)
- [x] Rows selection 
- [x] Custom actions (eg: delete, refresh, etc)
- [ ] Column hiding/ordering - Logic (hook) implemented, view part on hold until dropdown component is ready
- [ ] Dynamic width columns -> Todo in a separate PR
- [ ] Styling fixes & theming -> Todo in a separate PR after design is finished

image


src/stories/data/movies.js
```diff
@@ -0,0 +1,1919 @@
+// Author https://github.com/yegor-sytnyk/movies-list
```
jruddell, line 1: lets name this something more related to it being a list of items... or that it is for the grid. `movie-list` or `grid-list` reads better imo

src/stories/data/movies.js
```diff
@@ -0,0 +1,1919 @@
...
+
+export default [
```
jruddell, line 3: this should be a Typescript file

src/stories/elements/grid.stories.tsx
```diff
@@ -0,0 +1,186 @@
...
+    margin: theme.spacing(0, 1),
+    textAlign: 'center',
+    border: '1px solid #E6E6EA',
+    color: '#868696',
+    borderRadius: 3,
+    padding: theme.spacing(0.5, 1),
+  },
+}))
+
+interface Movie {
```
jruddell, line 28: Hmm, maybe we should put the interface with the seed data? `import data, { Movie } from '../data/movie-list`

src/stories/data/movies.js
xantorres: ok

src/app/core/components/grid/GridDefaultActionButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+    color: '#3D3D57',
+  },
+  btn: {
+    '& > .button-text': {
+      fontWeight: 'normal',
+      color: '#3D3D57',
+    },
+    '&:hover > .button-text': {
+      fontWeight: 'normal',
+      color: '#3D3D57',
```
jruddell, line 18: we have color variables. This is grey 700. or better components.typography.default But you should be using a secondary button here instead and not worry about colors

src/app/core/components/grid/GridDefaultDeleteButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+export const GridDefaultDeleteButton = ({
+  disabled,
+  children = 'Delete',
+  runAction,
+}: ActionComponentProps) => {
+  const classes = useStyles()
+
+  return (
+    <Button className={classes.btn} variant="table" disabled={disabled} onClick={runAction}>
+      <FontAwesomeIcon solid aria-hidden="true" className={classes.icon}>
```
jruddell, line 35: just pass an icon to the button instead `<Button icon="trash-alt"`

src/app/core/components/grid/GridToolbar.tsx
```diff
@@ -0,0 +1,92 @@
...
+        ))}
+        <div className={classes.verticalLine} aria-hidden="true" />
+        {filters.map(({ FilterComponent, ...rest }) => (
+          <FilterComponent {...rest} />
+        ))}
+        {/* <Button>Manage Columns</Button> */}
+        {onRefresh && (
+          <>
+            <div className={classes.verticalLine} aria-hidden="true" />
+            <GridDefaultActionButton runAction={onRefresh}>
```
jruddell, line 81: can pass the icon as a prop here as well

src/app/core/components/progress/Progress.js
```diff
@@ -204,7 +204,7 @@ Progress.defaultProps = {
   renderContentOnMount: true,
   renderLoadingImage: true,
   message: 'Loading',
-  minHeight: 400,
+  minHeight: 500,
```
jruddell, line 207: I really dislike our progress with a minHeight. can we just remove this?

jruddell (approved):

src/app/core/components/progress/Progress.js
xantorres: It's been there for ages, if we remove this now many things will look bad because the animation logo will be positioned at the top most

src/app/core/components/grid/GridDefaultActionButton.tsx
xantorres: I was waiting for the definitive designs and and theming to put these correctly, didn't really pay much attention. In any way the secondary button didn't look at all like the current designs so I did not use that

src/app/core/components/progress/Progress.js
xantorres: We can address this and other stuff related to Progress in another PR

PR #1458, opened 2022-01-23
UX-1447/converted all core elements to use our new naming convention

jruddell:
I also Removed MUI from our text component and built it from scratch


jruddell:
The naming change is an issue when it is just a capitalization change, need to move things around twice to do this correctly

PR #1460, opened 2022-01-24
UX-1524: Moved to older prettier(1.19.1) and re-formatted

patil-pratik-87:
Comments here -> https://github.com/platform9/pf9-ui-plugin/pull/1455 are addressed in this PR.

Changes:-
Updated linting plugins and configs in .eslintrc.
Updated specific rules in .eslintrc.
Moved to older prettier(1.19.1)
Fixed the pre-commit hook to format and lint successfully before committing.

Note -->
- Still, we have a lot of warnings in linting. Some of them might be legitimate errors. Feel free to change the rules to throw errors and fix the issues if you feel so.

- I suggest you turn on the Eslint extension of the vscode and add

**"editor.codeActionsOnSave": { 
"source.organizeImports": false, 
"source.fixAll.eslint": true }, 
"editor.formatOnSave": true,**              in your settings.json in vscode to quick fix the formatting issues in the editor.

- Enable  **Prettier** extension from vscode.

- Enable **FormatOnSave** check from Preferences>Settings

Let me know if you find something weird in the application :)

I see the compilation and lining green at this stage.


xantorres (approved):

jruddell (changes requested):
Unfortunately were not going to be able to remove prettier. For now we should just stick on an older version that honors our code style and isn't as opinionated. the point here is devs shouldn’t care about code style. It should just work. And if we dont have something that will consistently format everything we will expose a number of issues that will reduce dev productivity

patil-pratik-87:
> Unfortunately were not going to be able to remove prettier. For now we should just stick on an older version that honors our code style and isn't as opinionated. the point here is devs shouldn’t care about code style. It should just work. And if we dont have something that will consistently format everything we will expose a number of issues that will reduce dev productivity

Yeah, after our discussion, moved to the older version

jruddell (approved):

jruddell:
Looks great! thanks for taking the initiative here @patil-pratik-87

PR #1461, opened 2022-01-24
UX-1538/Added ArgoCD and Metal³ plugins

jruddell:
image (48)

image (47)

image (49)
)


knnguy (approved):

knnguy:
Are we not doing kebab case for file names anymore?

jruddell:
> Are we not doing kebab case for file names anymore?

We are... which files are you referring to? The only one I can see as a deviation is the `logo.tsx` file name, however, I consider this an edge case where we don't follow the rules. Each plugin should have a consistent form, and a plugin can have the concept of a logo. I want the plugin structure to follow this format so we can improve things with dynamic loading and not need to map variable names into an import statement (as an example).

knnguy:
> > Are we not doing kebab case for file names anymore? Noticed some files with kebab case and the new files with pascal case
> 
> We are... which files are you referring to? The only one I can see as a deviation is the `logo.tsx` file name, however, I consider this an edge case where we don't follow the rules. Each plugin should have a consistent form, and a plugin can have the concept of a logo. I want the plugin structure to follow this format so we can improve things with dynamic loading and not need to map variable names into an import statement (as an example).

I was referring to `InformationalLandingPage.tsx` and `current-plan-card.tsx`

jruddell:
> > > Are we not doing kebab case for file names anymore? Noticed some files with kebab case and the new files with pascal case
> > 
> > 
> > We are... which files are you referring to? The only one I can see as a deviation is the `logo.tsx` file name, however, I consider this an edge case where we don't follow the rules. Each plugin should have a consistent form, and a plugin can have the concept of a logo. I want the plugin structure to follow this format so we can improve things with dynamic loading and not need to map variable names into an import statement (as an example).
> 
> I was referring to `InformationalLandingPage.tsx` and `current-plan-card.tsx`

Thats just a file we haven't refactored, I just did an update inside it. `InformationalLandingPage` is a new file and hence the new naming convention. renaming files should be in its own PR

PR #1465, opened 2022-01-25
UX-1394/Create list and detail pages for ReplicaSets

knnguy:
Changes:
- Added list and detail pages for replica sets
- Modified events table columns
- Also modified all current workload columns so that it is consistent with the pods one
- Modified workload events page so that it only shows events for the specified resource. We shouldn't be pulling in events from all other related resources. For example, for services, only show events for the service, not all the related pods, pvcs, ingresses, etc. 

Screen Shot 2022-01-24 at 11 13 55 AM

Screen Shot 2022-01-24 at 11 02 52 AM

Screen Shot 2022-01-24 at 11 05 36 AM

Screen Shot 2022-01-24 at 11 05 53 AM

Screen Shot 2022-01-24 at 11 06 00 AM

Screen Shot 2022-01-25 at 10 57 04 AM


knnguy:
@jruddell @chrisj-platform9 

Is there a reason why the replicas card for stateful sets shows desired/total and for replica sets, it's current/desired? Shouldn't desired always be the denominator for consistency? Or does the order actually matters?

Stateful Sets
Screen Shot 2022-01-25 at 11 00 19 AM

Replica Sets
Screen Shot 2022-01-25 at 11 00 33 AM

src/app/plugins/kubernetes/components/replica-sets/list-page.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+const ListPage = ({ ListContainer }) => {
+  return () => {
+    const { params, getParamsUpdater } = usePrefParams(defaultParams)
+    const { message, loading, reload } = useListAction(listReplicaSets, {
+      params,
+      requiredParams,
+    })
+
+    const replicaSetsSelector = makeReplicaSetsSelector(params)
```
jruddell, line 35: Isn't this re-creating a selector each time you render? the makers are usually constants defined outside of components or memoized

xantorres: Yes, and you don't need to provide any param, that's just if you wanted to pass some default parameters (usually not needed)

src/app/plugins/kubernetes/components/replica-sets/actions.ts
```diff
@@ -0,0 +1,52 @@
...
+export const updateReplicaSet = replicaSetActions.add(
+  new UpdateAction<
+    DataKeys.ReplicaSets,
+    {
+      clusterId: string
+      namespace: string
+      name: string
+      body: any
+    },
+    any
```
xantorres, line 43: I think this should this be of type `IGenericResource<IReplicaSetSelector>`, or so

src/app/plugins/kubernetes/components/replica-sets/list-page.tsx
jruddell: yep basically just move this line outside of the component. params still go through to the selector instance

jruddell (approved):

PR #1466, opened 2022-01-25
UX-1395/Added cronjobs list page

kevinylee1991:
Rest of screens & fleshing out of list page still in progress

Screen Shot 2022-01-25 at 12 11 35 PM


src/app/core/utils/routes.ts
```diff
@@ -962,5 +963,32 @@ export const routes = {
       tab: 'support',
     },
   }),
+  cronjobs: {
+    list: Route.register({
+      url: `${pluginRoutePrefix}/workloads/cronjobs`,
```
jruddell, line 968: `/cron-jobs/`

src/app/plugins/kubernetes/components/cronjobs/job-type-picklist.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { forwardRef } from 'react'
```
jruddell, line 1: These file names should reflect the default export

jruddell (approved):

PR #1469, opened 2022-01-25
Fix tsc script in pre-commit hook and add tsc script to check TS errors

xantorres:
Now ts errors will show up when trying to commit invalid code:

image

You can also test ts errors by running `yarn check-ts`


package.json
```diff
@@ -184,7 +183,8 @@
     "test:ci:windows:record": "cross-env NODE_ENV=test bin-up cypress run --record --group",
     "test:unit": "cross-env NODE_ENV=test jest --coverage src/app src/server",
     "test:unit:watch": "cross-env NODE_ENV=test jest --watch --coverage src/app src/server",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "tsc": "./node_modules/.bin/tsc --noEmit -p './tsconfig.json'"
```
jruddell, line 187: can we call this `check-ts` so its more declarative as to what its doing?

jruddell (approved):

PR #1475, opened 2022-01-27
UX-1395/Cronjob & job details pages

kevinylee1991:
There are a few small portions I left out that were in the wireframes where I need to get an example cronjob/job that has those values populated. Also the list tables were not in the wireframes so need to make sure I get the right columns in the tables in another commit.

Screen Shot 2022-01-26 at 4 24 11 PM
Screen Shot 2022-01-26 at 4 24 43 PM
Screen Shot 2022-01-26 at 4 24 54 PM
Screen Shot 2022-01-26 at 4 25 03 PM
Screen Shot 2022-01-26 at 4 25 59 PM
Screen Shot 2022-01-26 at 4 26 08 PM
Screen Shot 2022-01-26 at 4 26 15 PM
Screen Shot 2022-01-26 at 4 26 30 PM


src/app/plugins/kubernetes/components/cronjobs/CronjobsListPage.tsx
```diff
@@ -35,6 +34,8 @@ const ListPage = ({ ListContainer }) => {
       requiredParams,
     })
 
+    // Having this line here is the only way to pass in these params, useSelectorWithParams is not sending them through
```
kevinylee1991, line 37: I moved this back for a bit because the filter no longer worked (the params were not being passed onto the selector). Going to try to figure out what's going on to fix it & move the selector maker out of the component.

jruddell (approved):

PR #1483, opened 2022-01-29
UX-1395/Added cronjob & job creation page

kevinylee1991:
Have some questions about some things this PR touches on:

1. The CreateAction sends a dispatch to add the newly created item to the cache, but since this is a non-converted resource it doesn't mesh well with the other resources and causes a duplicate entry. Is this dispatch required?
2. When creating a new resource, the list is refetched from the backend anyways. What piece of code is triggering this refetch?
3. I made a useCreateAction which ended up being a complete clone of the useUpdateAction except I changed the param typing, but wouldn't useUpdateAction be usable for any CRUD operation?

Screen Shot 2022-01-28 at 5 04 45 PM


src/app/core/actions/CreateAction.ts
```diff
@@ -19,15 +19,17 @@ class CreateAction<
...
-        item: result,
-      }),
-    )
+    // dispatch(
+    //   cacheActions.addItem({
+    //     cacheKey,
+    //     params: indexedParams,
+    //     item: result,
+    //   }),
+    // )
```
xantorres, line 32: If you remove this, the new created item won't be added to the cache and it will become unsynced with the server values. You probably haven't noticed it since we are usually calling "onRefresh" after a table update so the values are fetched from the server again, but it doesn't feel right, and in my opinion we shouldn't probably be doing that in the first place.
To fix this, we should probably "convert" the resource before adding it to the cache.

jruddell:
I think having a single updater makes a lot of sense, lets repurpose useUpdateAction here

src/app/core/hooks/useCreateAction.ts
```diff
@@ -0,0 +1,69 @@
...
+  )
+
+  useEffect(() => {
+    return () => {
+      // Set the unmounted ref to true to prevent further state updates
+      unmounted.current = true
+    }
+  }, [])
+
+  return { update, updating }
```
xantorres, line 66: Can we change these variable names to `create` and `creating` to differentiate them from the `useUpdateAction` hook?

xantorres:
> I think having a single updater makes a lot of sense, lets repurpose useUpdateAction here

After looking at it, yeah, I think just using a union type in the current `useUpdateAction` hook would be enough
image

src/app/plugins/kubernetes/components/cronjobs/actions.ts
```diff
@@ -49,6 +52,23 @@ export const updateJob = jobActions.add(
...
+      trackEvent('Create Job', {
+        id: pathStr('metadata.uid', created),
+        name: pathStr('metadata.name', created),
+      })
+      return created
+    },
+  ),
+)
+
+console.log(jobActions, 'job actions')
```
jruddell, line 70: console.log

jruddell (approved):
remove the console.log first and were good to merrge this!

PR #1487, opened 2022-02-01
UX-1402/Default to the newest kubernetes version when creating a cluster

Nalinmujoo:
The field is pre-filled with the newest Kubernetes version.
Screenshot 2022-02-01 at 11 31 45 AM

**Will not be sorting in descending order.**
~~The options are sorted in descending order by version (newest first).~~
Screenshot 2022-02-01 at 11 32 01 AM


patil-pratik-87 (approved):

src/app/plugins/kubernetes/components/common/kubernetes-version-picklist.tsx
```diff
@@ -1,16 +1,43 @@
...
+  onChange: (key: string) => void
+  value: string
+  selectFirst?: boolean
+}
+
+const KubernetesVersionPicklist = forwardRef(
+  ({ onChange, value, selectFirst, ...rest }: Props, ref) => {
+    const [roles, loading] = useDataLoader(loadSupportedRoleVersions)
+    const options = useMemo(() => {
+      roles.sort((a, b) => compareVersions(b.roleVersion, a.roleVersion)) //sort in descending order
```
jruddell, line 19: I dont think we should be sorting the array here, this will mutate the array as well. Instead lets keep it in asc order, but just pick the latest / highest value as the default selection.

src/app/plugins/kubernetes/components/infrastructure/clusters/form-components/kubernetes-version.tsx
```diff
@@ -17,6 +17,7 @@ export default function KubernetesVersion({
       onChange={(value) => setWizardContext({ kubeRoleVersion: value })}
       value={wizardContext.kubeRoleVersion}
       required={required}
+      selectFirst={true}
```
xantorres, line 20: you don't need to explicitly set `true`, if you set `selectFirst` implicitly it will be true

src/app/plugins/kubernetes/components/common/kubernetes-version-picklist.tsx
```diff
@@ -1,16 +1,43 @@
...
+      return projectAs({ value: 'roleVersion', label: 'roleVersion' }, roles)
+    }, [roles])
+
+    useEffect(() => {
+      if (selectFirst) {
+        onChange(options[0]?.value)
+        selectFirst = false
```
jruddell, line 26: This is assigning a value to a prop which is an anti-pattern in React. What is the reason for doing this? help explain the goal and I can help rejigger things for you

jruddell (changes requested):
Please address the review comments before moving forward

jruddell (approved):

PR #1488, opened 2022-02-02
UX-1521/Fixed storage class wizard mask inputs and yaml editor

kevinylee1991:


src/app/core/elements/text.tsx
```diff
@@ -48,23 +48,28 @@ const useStyles = makeStyles<Theme, { variant: TextVariant }>((theme) => ({
...
-export default function Text({
-  className,
-  variant = 'body1',
-  component = variantMap[variant] || 'span',
-  noWrap = false,
-  children,
-  ...props
-}: PropsWithChildren<TextProps>) {
+// forwardRef required for mask from Input component to work
+const Text = forwardRef<HTMLElement, TextProps>((props, ref) => {
```
jruddell, line 52: why do we need to forward ref here?

kevinylee1991: the input mask wasn't working anymore -- refs don't work with functional components, which the input mask relies on. When the Text component got converted to a functional component, the mask stopped working

jruddell (approved):

PR #1489, opened 2022-02-02
UX-1418/Daemon Set Overview Page

knnguy:
- Added overview page for daemon sets
- Fixed the volume types in the volumes table for replica sets and stateful sets. Realized that a workload resource could have multiple volumes of different types

Screen Shot 2022-02-01 at 3 56 57 PM
Screen Shot 2022-02-01 at 3 57 03 PM


src/app/plugins/kubernetes/components/storage/helpers.ts
```diff
@@ -0,0 +1,56 @@
+import { isNilOrEmpty } from 'utils/fp'
+
+export const getVolumeType = (volume) => {
+  if (isNilOrEmpty(volume)) return null
+  if (volume.hasOwnProperty('awsElasticBlockStore')) {
```
jruddell, line 5: can we not do a dynamic mapping here? im not a fan of a lot of if else's like this

jruddell: Something more like 

```
const volumeFormattedNames = {
  awsElasticBlockStore: 'AwsElasticBlockStore',
  cephfs: 'Cephfs',
  cinder: 'Cinder',
  configMap: 'ConfigMap',
  downwardAPI: 'DownwardAPI',
  emptyDir: 'EmptyDir',
  fc: 'FC',
  gcePersistentDisk: 'GcePersistentDisk',
  glusterfs: 'Glusterfs',
  hostPath: 'HostPath',
  iscsi: 'Iscsi',
  local: 'Local',
  nfs: 'Nfs',
  persistentVolumeClaim: 'PersistentVolumeClaim',
  portworxVolume: 'PortworxVolume',
  projected: 'Projected',
  secret: 'Secret',
  vsphereVolume: 'VsphereVolume',
}

const lookupKey = Object.keys(volumeFormattedNames).find(key => volume.hasOwnProperty(key))
return volumeFormattedNames[lookupKey]  || null
```

knnguy: Yep, I agree. This is much easier to read and maintain. I'll change it

jruddell (approved):

PR #1492, opened 2022-02-03
UX-1528/Added delete operation to CSI drivers

kevinylee1991:


src/app/plugins/kubernetes/components/storage/csi-drivers/actions.ts
```diff
@@ -23,6 +24,13 @@ const csiDriverActions = createCRUDActions(DataKeys.CSIDrivers, {
     },
     [persistentVolumeActions.list, persistentVolumeClaimActions.list],
   ),
+  deleteFn: async ({ id }, currentItems) => {
+    const { clusterId, name } = await currentItems.find((x) => x.id === id)
```
jruddell, line 28: do you need to await here?

jruddell (approved):

PR #1496, opened 2022-02-04
Integrate ListContainer with new Grid and multiple fixes and optimizations

xantorres:


src/app/core/components/grid/GridSearchFilter.tsx
```diff
@@ -0,0 +1,122 @@
+import React, { useCallback, useMemo, useState, useEffect } from 'react'
+import { debounce } from 'utils/async'
+import { makeStyles } from '@material-ui/styles'
+import Theme from 'core/themes/model'
+import { TextField, InputAdornment } from '@material-ui/core'
```
jruddell, line 5: why are we using material ui core elements?

jruddell (approved):

src/app/core/components/grid/GridSearchFilter.tsx
xantorres: For now, I just copied and pasted the styles of the old search input, until we get the definitive design

jruddell: ok, we do have a design in place but I don't mind working on the styling some

PR #1497, opened 2022-02-04
UX-1454/Breadcrumb component styling and interactions

jruddell:
still need to add / work on

- [x] hover states 
- [x] dynamic collapsing

image

opinionated collapsing 
image

dark
image

ultra dark
image


src/app/core/elements/breadcrumbs/helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+})
+
+const getCrumbName = (pathPart, params) => {
+  if (pathPart.includes(':')) {
+    const key = pathPart.replace(':', '')
+    return params[key]
+  } else {
+    return pathPart
+      .replace(/-/g, ' ')
+      .split(' ')
```
xantorres, line 16: Why not just split by '-' so you can skip the "replace" step?

src/app/core/plugins/route.ts
```diff
@@ -57,22 +67,27 @@ export class Route<T extends OptionalGenericKVP = null> {
     return route
   }
 
-  static getRoutes(): Route[] {
+  static getRoutes<T extends OptionalGenericKVP = null>(): Route<T>[] {
```
xantorres, line 70: This method shouldn't be using a generic, since it could be given any arbitrary key value pair interface which could have nothing to do with the routes object declared on the constructor.
I think the appropiate way of dealing with this could be having a static function behaving as a factory that would return a singleton typed routes object that you would then use everywhere in the app.

xantorres: In my oppinion the approach to follow should be something similar to what we do with the reducers and the store, which are strongly typed.

jruddell: Yea I was playing with the typings because i was getting a Route<null> but ended up not needing the change at all so I'll go ahead and remove it

src/app/core/plugins/route.ts
```diff
@@ -61,18 +71,23 @@ export class Route<T extends OptionalGenericKVP = null> {
...
-    return Route.getRoutes().find((r) => {
-      return !!r.pattern.match(pathname)
-    })
+    return Route.find(pathname)
   }
+
+  static find = memoize(<T extends OptionalGenericKVP = null>(pathname: string): Route<
+    T
+  > | null => {
+    return Route.getRoutes().find((r) => !!r.pattern.match(pathname))
```
xantorres, line 84: Why not just use `Route.routes` and avoid the overhead of calling a function?

jruddell: We can do that if you want, it doesn't seem like really much overhead since its essentially just a `get property()`

xantorres (approved):

PR #1499, opened 2022-02-04
UX-1564/Fixed issue where sidenav & header display can be incorrect

kevinylee1991:
Needed to blacklist the frame or else the persist keeps the incorrect frame properties & the bootstrapping that resets the frame properties gets overwritten.


jruddell (approved):

src/app/store.ts
```diff
@@ -1,14 +1,15 @@
...
+import { clientStoreKey } from 'core/client/clientReducers'
 import { TypedUseSelectorHook, useDispatch, useSelector } from 'react-redux'
 import { persistReducer, persistStore } from 'redux-persist'
 import storage from 'redux-persist/lib/storage' // defaults to localStorage for web
 
 const persistConfig = {
   key: 'root',
   storage,
-  blacklist: [cacheStoreKey],
+  blacklist: [cacheStoreKey, clientStoreKey],
```
xantorres, line 12: Isn't this blacklisting the whole client store state?

jruddell: The client store has its own persist config that specifies only `frame` to be omitted

kevinylee1991: their blacklist is kinda confusing, since the reducers are chained (the reducer in store.ts and the rootReducer), the blacklist should only apply to the nested properties.

https://github.com/rt2zz/redux-persist#nested-persists

PR #1511, opened 2022-02-09
Prepared new actions for account pages before ListContainer refactor

kevinylee1991:


jruddell (commented):
Lets try and fix the todos before merging

src/app/plugins/account/components/userManagement/users/new-actions.ts
```diff
@@ -0,0 +1,382 @@
...
+          console.warn((err || {}).message)
+          return emptyArr
+        },
+      )(null)
+
+      // Todo: do a listTenants.call()
+      listTenants.call({})
+      // Refresh the user/roles cache
+      // Todo: do a userRoleAssignmentsList.call()
+      await listUserRoleAssignments.call({ userId })
```
xantorres, line 331: maybe we can remove this `await` here, so the load of the tenants won't block the main thread and the user experience will be better.

PR #1524, opened 2022-02-11
UX-1599/Fix storage provider list page bug

knnguy:
**Issue:**

Storage providers list page was showing that the PF9 repository was not connected even though it was. Reason was that the code was looking for the url `https://platform9.github.io/helm-charts` without the slash at the end but the api was giving us `https://platform9.github.io/helm-charts/`  which has the slash at the end

**Fix:**

Accept the url with or without the slash


src/app/plugins/kubernetes/components/storage/storage-providers/list-page.tsx
```diff
@@ -57,7 +56,7 @@ const StorageProvidersListPage = () => {
   }, [clusters])
 
   const hasPf9Repository = useMemo(() => {
-    return !!repositories.find((r) => r.url === pf9DefaultRepository.url)
+    return !!repositories.find((r) => r.url.match(`^${pf9DefaultRepository.url}(/?)$`))
```
xantorres, line 59: wouldn't a simpler `r.url.includes(pf9DefaultRepository.url)` have worked too?

xantorres (approved):

PR #1526, opened 2022-02-11
Prepped new action files for infrastructure resources

kevinylee1991:


src/app/plugins/kubernetes/components/infrastructure/cloudProviders/new-actions.ts
```diff
@@ -0,0 +1,112 @@
...
+export const listCloudProviders = cloudProviderActions.add(
+  new ListAction<DataKeys.CloudProviders>(async () => {
+    Bugsnag.leaveBreadcrumb('Attempting to get cloud providers')
+    return qbert.getCloudProviders()
+  }),
+)
+
+export const createCloudProvider = cloudProviderActions.add(
+  new CreateAction<DataKeys.CloudProviders, any, any>(async (params) => {
+    Bugsnag.leaveBreadcrumb('Attempting to create cloud provider', params)
```
xantorres, line 31: No `any`'s please

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/new-actions.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export const listCloudProviders = cloudProviderActions.add(
+  new ListAction<DataKeys.CloudProviders>(async () => {
+    Bugsnag.leaveBreadcrumb('Attempting to get cloud providers')
+    return qbert.getCloudProviders()
+  }),
+)
+
+export const createCloudProvider = cloudProviderActions.add(
+  new CreateAction<DataKeys.CloudProviders, any, any>(async (params) => {
```
xantorres, line 30: No `any`'s please, you can remove the third generic and let the typescript engine infer the return value by returning a explicitly defined value in `qbert.createCloudProvider(params)`

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/new-actions.ts
```diff
@@ -0,0 +1,112 @@
...
+    Bugsnag.leaveBreadcrumb('Attempting to create cloud provider', params)
+    const result = await qbert.createCloudProvider(params)
+    trackEvent('Create Cloud Provider', {
+      cloud_provider_name: params.name,
+      cloud_provider_type: params.type,
+    })
+    return result
+    // const { uuid } = result || {}
+
+    // Below todo was already there. This below part seems weird, can we remove it?
```
jruddell, line 40: I think there is some missing data in the returned result.. specifically it looks like nodepooluuid. And this is needed for mapping clusters to cloud providers it looks like. So we probably still need this? can you confirm the result from the create call is missing this field?
image

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/new-actions.ts
```diff
@@ -0,0 +1,109 @@
...
+})
+
+export const listAddonVersions = addonVersionActions.add(
+  new ListAction<DataKeys.AddonVersions, { clusterId: string }, any>(async ({ clusterId }) => {
+    Bugsnag.leaveBreadcrumb('Attempting to get addon versions')
+    return qbert.getAddonVersions(clusterId)
+  }),
+)
+
+export const loadAddonVersions = createContextLoader(
```
jruddell, line 99: do we still need this? isn't the class above the same thing?

src/app/plugins/kubernetes/components/infrastructure/clusters/newActions.ts
```diff
@@ -187,3 +215,104 @@ export const upgradeClusterNodes = clusterActions.add(
...
+      numNodes: (nodes || []).length,
+    })
+    await qbert.detachNodes(cluster.uuid, nodes)
+    trackEvent('Cluster Detach Nodes', {
+      numNodes: (nodes || []).length,
+      clusterUuid: cluster.uuid,
+    })
+
+    // Old one had return prevItems, adding equivalent here
+    const clusters = clustersSelector(store.getState(), {})
```
jruddell, line 249: Isn't this the view model getting written back to the store?

src/app/plugins/kubernetes/components/infrastructure/clusters/newActions.ts
```diff
@@ -187,3 +215,104 @@ export const upgradeClusterNodes = clusterActions.add(
...
+      numNodes: (nodes || []).length,
+    })
+    await qbert.attachNodes(cluster.uuid, nodes)
+    trackEvent('Cluster Attach Nodes', {
+      numNodes: (nodes || []).length,
+      clusterUuid: cluster.uuid,
+    })
+    // Old one had return prevItems, adding equivalent here
+    const clusters = clustersSelector(store.getState(), {})
+    return clusters
```
xantorres, line 226: Unless you update the clusters in some way, there's no need to return anything here, if nothing is returned then the store won't be updated.

src/app/plugins/kubernetes/components/infrastructure/clusters/newActions.ts
xantorres: > Isn't this the view model getting written back to the store?

yeah, this is not needed

src/app/plugins/kubernetes/components/infrastructure/nodes/new-actions.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export const authNode = nodeActions.add(
+  new CustomAction<DataKeys.Nodes, any, any>('authNode', async (node) => {
+    Bugsnag.leaveBreadcrumb('Attempting to authorize node', { node })
+    await resMgr.addRole(node.uuid, 'pf9-kube', {})
+    trackEvent('Authorize Node', {
+      node_name: node.name,
+    })
+    // Todo: Once integrated, check if below is necessary or if the table already does it
+    listNodes.call({})
```
xantorres, line 74: This is only necessary if something has changed in the nodes, but in this case I would probably lean towards updating the node in the cache rather than doing this.

src/api-client/Qbert.ts
```diff
@@ -186,7 +192,7 @@ class Qbert extends ApiService {
   })
   getCloudProviderRegionDetails = async (cpId, regionId) => {
     const url = `/cloudProviders/${cpId}/region/${regionId}`
-    return this.client.basicGet({
+    return this.client.basicGet<any>({
```
kevinylee1991, line 195: this response can vary greatly between cloud provider types

xantorres: Don't they have anything in common? This looks quite unreliable, we need to have consistent data in the cache store, so they should (at least) have a common primary identifier (id), so you can define the generic as `{ id: string }`

kevinylee1991: I changed it to use a CustomAction instead

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/new-actions.ts
```diff
@@ -0,0 +1,105 @@
...
+  new CustomAction<
+    DataKeys.CloudProviderRegionDetails,
+    { cloudProviderId: string; cloudProviderRegionId: string }
+  >('loadRegionDetails', async ({ cloudProviderId, cloudProviderRegionId }) => {
+    Bugsnag.leaveBreadcrumb('Attempting to load cloud provide region details', {
+      cloudProviderId,
+      cloudProviderRegionId,
+    })
+    // This API returns an object, not an array
+    return qbert.getCloudProviderRegionDetails(cloudProviderId, cloudProviderRegionId)
```
xantorres, line 103: This is not correct, you will need to provide a custom `postProcessor` as a second parameter of the `CustomAction`, otherwise this return here will replace the whole cache array.

xantorres: You can probably use a `UpdateAction` here as well

jruddell (approved):

PR #1527, opened 2022-02-12
Task/table design

jruddell:
With selection options

image
image

Without selection options
image


xantorres (approved):
Looks good so far, but the horizontal scroll bar should be on top of the pagination, not after it.

xantorres:
> > Looks good so far, but the horizontal scroll bar should be on top of the pagination, not after it.
> 
> Actually the design included it this way when the actions bar was at the top so I just did that for now. There's a few issues that need design first before we do much more for styling

If you do that  the pagination controls will move when you scroll, which is a weird behavior and probably not a good user experience.

jruddell:
No

> > Looks good so far, but the horizontal scroll bar should be on top of the pagination, not after it.
> 
> Actually the design included it this way when the actions bar was at the top so I just did that for now. There's a few issues that need design first before we do much more for styling

Nope, I made the pagination controls sticky so they don't move. I can add a video to help display that interaction

PR #1553, opened 2022-02-22
Attempt running multiple cypress containers, one per suite (#1430)

ParagDoke:
Allow capturing several env vars in cypress json format


vijay-nandagavali (approved):
Approved

PR #1554, opened 2022-02-22
Converted tenants pages to use new actions

kevinylee1991:
Had to make some changes within some of the core components/functions, please check those out to make sure they make sense thanks!

Screen Shot 2022-02-22 at 1 07 23 PM


src/app/plugins/account/components/userManagement/UserManagementIndexPage.js
```diff
@@ -2,14 +2,15 @@ import React from 'react'
 import { routes } from 'core/utils/routes'
 import Tabs from 'core/elements/tabs'
 import Tab from 'core/elements/tabs/Tab'
-import TenantsListPage from 'account/components/userManagement/tenants/TenantsListPage'
+// import TenantsListPage from 'account/components/userManagement/tenants/TenantsListPage'
```
jruddell, line 5: Id rather not make a new file, just update this existing file to use the new list system. The goal here is to remove the excess old code

src/app/core/actions/ListAction.ts
```diff
@@ -31,10 +31,10 @@ class ListAction<
   }
 
   public addDependency = <D extends keyof IDataKeys>(listAction: IListAction<D, P>) => {
-    if (this.deps.has(listAction.name)) {
-      throw new Error(`Action "${listAction.name}" already exists in ActionsSet`)
+    if (this.deps.has(listAction.config.cacheKey)) {
```
kevinylee1991, line 34: Xan/John could you confirm if this change is valid?

src/app/core/components/progress/Progress.tsx
```diff
@@ -66,6 +66,9 @@ export default function Progress({ children, ...props }: PropsWithChildren<Progr
   })
 
   useEffect(() => {
+    if (props.loading) {
```
kevinylee1991, line 69: This one as well, it didn't make sense to me to set loadedOnce to true in the case where props.loading is true.

src/app/core/hooks/useListAction.ts
```diff
@@ -37,7 +37,9 @@ const useIsDataLoadingSelector = (...cacheKeys): string[] => {
     () => cacheKeys.map((cacheKey) => [cacheStoreKey, loadingStoreKey, cacheKey]),
     cacheKeys,
   )
-  const keysLoadingStatus = useSelector<RootState, boolean[]>(paths(keyPaths))
+  const keysLoadingStatus = useSelector<RootState, boolean[]>(paths(keyPaths)).map((status) =>
```
kevinylee1991, line 40: If loading status is undefined, it is the equivalent of loading = false currently, but if the loading status is undefined it makes more sense for it to be true, since that resource has not loaded at all yet.

src/app/core/hooks/useListAction.ts
```diff
@@ -118,10 +126,12 @@ const useListAction = <
   const loadData = useCallback(
     (refetch = false) => {
       if (refetch || isNil(cachedParams)) {
-        action.call(memoizedIndexedParams, loadingFeedback)
+        // Adding refetch: true here is required because without adding
```
kevinylee1991, line 129: can you confirm this part makes sense?

src/app/core/actions/ListAction.ts
xantorres: Looks good to me

src/app/core/hooks/useListAction.ts
```diff
@@ -118,10 +126,12 @@ const useListAction = <
...
+        // refetch: true here, the action will only upsert & not replace
+        action.call({ ...memoizedIndexedParams, refetch: true }, loadingFeedback)
```
xantorres, line 131: If you do this you will break the memoization since it will be using a new object every time

xantorres: Maybe we could add another param to the `Action`, to be able to pass the `refetch` variable

src/app/core/components/progress/Progress.tsx
jruddell: yes this is correct, we can also cleanup the state some as `loading` isn't used in state.

src/app/core/hooks/useListAction.ts
kevinylee1991: Maybe if we used memoizedObj again? Or just remove this refetch block entirely? If it's the function returned as reload, it seems strange to have this refetch param defaulted as false

src/app/plugins/account/components/userManagement/groups/EditUserGroupPage.tsx
```diff
@@ -56,15 +59,18 @@ const EditUserGroupPage = () => {
   const { match, history } = useReactRouter()
   const classes = useStyles({})
   const groupId = match.params.id
-  const [tenants, loadingTenants] = useDataLoader(mngmTenantActions.list)
+
+  const { loading: loadingTenants } = useListAction(listTenants)
+  const tenants = useSelectorWithParams(tenantsSelector, {})
```
xantorres, line 64: is this necessary? Could we simply use `useSelector` instead?

src/app/plugins/account/components/userManagement/tenants/AddTenantPage.js
```diff
@@ -23,26 +24,24 @@ const initialContext = {
...
+
+  const { update, updating } = useUpdateAction(createTenant)
+
+  const submitForm = async (params) => {
+    await update(params)
+    history.push(listUrl)
+  }
+
+  const { loading, reload } = useListAction(listUsers)
+  const users = useSelectorWithParams(usersSelector)
```
xantorres, line 40: `useSelector` if we are not using any params

src/app/plugins/account/components/userManagement/tenants/TenantsListPage.tsx
```diff
@@ -0,0 +1,78 @@
...
+  // if it shows perma-loading. Currently is bc the tenants are loaded
+  // but the namespaces are not, but the namespaces dependency is not
+  // being fetched
+
+  const data = useSelectorWithParams(tenantsSelector, params)
+
+  return (
+    <>
+      <DocumentMeta title="Replica Sets" />
+      <ListContainer<ModelDataKey>
```
xantorres, line 57: Please check out some other ListPages since I updated the models and we should use the selector model rather than the api model

src/app/plugins/account/components/userManagement/tenants/selectors.ts
```diff
@@ -50,3 +50,20 @@ export const makeFilteredTenantsSelector = (
     )(tenants)
   })
 }
+
+export const userTenantsSelector = createSharedSelector(
```
xantorres, line 54: `export const userTenantsSelector = getDataSelector<DataKeys.UserTenants>(DataKeys.UserTenants)`

src/app/plugins/account/components/userManagement/tenants/selectors.ts
```diff
@@ -50,3 +50,20 @@ export const makeFilteredTenantsSelector = (
...
+  getDataSelector<DataKeys.UserTenants>(DataKeys.UserTenants),
+  (userTenants) => {
+    return userTenants
+  },
+)
+
+export const tenantRolesSelector = createSharedSelector(
```
xantorres, line 61: Same here

src/app/plugins/kubernetes/components/infrastructure/clusters/import/ImportAKSRequirements.tsx
```diff
@@ -53,7 +54,7 @@ const ImportAKSRequirements = ({ onComplete, platform }) => {
   const [showDialog, setShowDialog] = useState(false)
   const [errorMessage, setErrorMessage] = useState(null)
   const [cloudProviders] = useDataLoader(cloudProviderActions.list)
-  const [userTenants] = useDataLoader(loadUserTenants)
+  const userTenants = useSelectorWithParams(userTenantsSelector, {})
```
xantorres, line 57: `useSelector`

xantorres (approved):

PR #1555, opened 2022-02-22
Refactor Deployments and Services list pages to use new ListContainer

knnguy:
**Changes:**
- Deployment and Services list pages are now using the new actions and new `ListContainer` component
- Modified the `DeleteDeploymentDialog` to use the new delete action
- Fixed a bug where the delete action wasn't deleting the correct item in the cache

Screen Shot 2022-02-22 at 2 12 23 PM
Screen Shot 2022-02-22 at 2 12 38 PM


src/app/core/actions/DeleteAction.ts
```diff
@@ -17,7 +17,14 @@ class DeleteAction<
   protected postProcess = (result, params: P) => {
     const { indexBy = emptyArr, cacheKey, uniqueIdentifier } = this.config
     const allIndexKeys = indexBy ? ensureArray(indexBy) : emptyArr
-    const indexedParams = pipe(pickAll(allIndexKeys), reject(either(isNil, equals(allKey))))(params)
+    const indexedParams = pipe(
```
knnguy, line 20: The delete action wasn't deleting the right action in the cache. Looks like it was because the `uniqueIdentifier` wasn't included in the `indexedParams`

jruddell (approved):

src/app/core/actions/DeleteAction.ts
xantorres: I think we don't even need indexedParams here to begin with, we could just pass the raw params

PR #1557, opened 2022-02-23
Refactor DaemonSet and Namespace list pages to use new actions and new ListContainer

knnguy:
**Changes:**
- `DaemonSet` and `Namespace` list pages are now using the new actions and new `ListContainer` component
- The `DeleteDialog`  and `SetAsDefaultDialog` components for namespaces are now using the new update actions
- Fixed a bug where the icon wasn't showing for the grid batch actions
- Added a delete action for `ReplicaSets` and `StatefulSets`

Screen Shot 2022-02-23 at 1 43 49 PM

Screen Shot 2022-02-23 at 1 42 57 PM


src/app/plugins/kubernetes/components/namespaces/selectors.ts
```diff
@@ -3,8 +3,11 @@ import DataKeys from 'k8s/DataKeys'
 import getDataSelector from 'core/utils/getDataSelector'
 import { importedClustersSelector } from '../infrastructure/importedClusters/selectors'
 import { createSharedSelector } from 'core/utils/selectorHelpers'
+import { RootState } from 'app/store'
+import { Selector } from '@reduxjs/toolkit'
+import { INamespaceSelector } from './model'
 
-export const namespacesSelector = createSharedSelector(
+export const namespacesSelector: Selector<RootState, any[]> = createSharedSelector(
```
knnguy, line 10: I kept getting a Typescript error when trying to specify the type as `Selector<RootState, INamespaceSelector[]>` and I just can't figure out why. It keeps complaining that the status type doesn't match but we already specified that the status type has been changed to a string.

In namespaces/model.ts:
```
type INamespaceApi = IGenericResource<GetClusterNamespacesItem>
export interface INamespaceSelector extends Omit<INamespaceApi, 'status'> {
  clusterName: string
  status: string
}
```

This is the error I got in `NamespacesListPage.tsx`
Screen Shot 2022-02-22 at 9 23 01 PM

jruddell: Lets do a call and discuss this issue in more detail, I think we are using the system incorrectly if were getting this error

jruddell (approved):

PR #1561, opened 2022-02-25
Task/labels and annotations

jruddell:
TODO:

- [x] consolidate the annotations logic with labels
- [x] figure out if we need all of the different components / ways to render these

new rendering style for labels and annotations (max 15ch, max 3 in a table rest collapse into a count of overflow). 

image
image

New badge component for rendering status's, labels/annotations and badges (like early access).
image
image

text ellipsis at a dynamic number of lines
image


kevinylee1991 (approved):
lgtm

PR #1563, opened 2022-02-25
Remove indexed params from UpdateAction and DeleteAction

knnguy:
Remove indexed params from UpdateAction and DeleteAction. Instead, just pass in the raw params


src/app/core/actions/DeleteAction.ts
```diff
@@ -15,23 +15,14 @@ class DeleteAction<
...
-    )(params)
+    const { cacheKey, uniqueIdentifier } = this.config
     const { dispatch } = store
 
     dispatch(
       cacheActions.removeItem({
         uniqueIdentifier,
         cacheKey,
-        params: indexedParams,
+        params: params,
```
xantorres, line 25: leave it as just `params,`

xantorres (approved):

jruddell (approved):

PR #1564, opened 2022-02-26
Added Dropdown component and some refactoring

xantorres:
Added 2 separate components: `Dropdown` and `MultiDropdown` and refactored `ClusterPicklist` and `NamespacePicklist` to use the new Dropdown. Additional Picklists will be done in a separate PR.

image

Take a look at the dropdown stories in Storybook and play around with the props for an example of use.


src/app/plugins/account/components/user-settings/theme-choice-card.tsx
```diff
@@ -1,21 +1,22 @@
...
 
 import Theme from 'core/themes/model'
-import { withCustomTheme } from 'core/themes/ThemeManager'
-import defaultTheme from 'core/themes/serenity/serenity-default'
-import lightTheme from 'core/themes/serenity/serenity-light'
-import darkTheme from 'core/themes/serenity/serenity-dark'
-import ultraDarkTheme from 'core/themes/serenity/serenity-ultra-dark'
-import customTheme from 'core/themes/serenity/serenity-custom'
+import { withCustomTheme, useCustomTheme } from 'core/themes/ThemeManager'
+import defaultTheme from 'core/themes/serenity/serenityDefault'
```
jruddell, line 9: Do you need to import all of the themes here? wouldn't it be better to use the index.ts file that exports the themes?

xantorres: You are right, didn't do much in this file tbh, will fix that in my next PR

PR #1565, opened 2022-02-26
Task/workloads lists

jruddell:
Started some cleanup on our models, started some generic cell components.

We need to fix the table typings to honor the selector model vs the API model

### Pods
image

### Deployments
image

### Services
image


src/api-client/qbert.model.ts
```diff
@@ -72,17 +75,12 @@ export interface PurpleMetadata {
   uid: string
   resourceVersion: string
   creationTimestamp: string
-  annotations?: PurpleAnnotations
+  annotations?: { [key: string]: string }
```
xantorres, line 78: I suggest using `Record<string, string>` for consistency sake

src/app/core/components/grid/cell-components/GridStatusCell.tsx
```diff
@@ -0,0 +1,33 @@
...
+function GridStatusCell<T>({ value, dataFn }: GridStatusCellProps<T>) {
+  const { variant, label = value } = useMemo(() => dataFn(value), [value])
+  return <Badge variant={variant} text={label} />
+}
+
+export default typeUnsafeMemoizeShallow(GridStatusCell, {
+  maxSize: 1000,
+}) as FC<GridStatusCellProps<any>>
+
+export function createGridStatusCell<T>({ dataFn }: BaseGridStatusCellProps) {
```
xantorres, line 24: I'm not a fan of putting business specific GridCells in the `grid` folder, I think the Grid itself should be agnostic about it and only provide the bare minimum default elements like a basic cell etc.

src/app/core/components/grid/cell-components/GridStatusMessageCell.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <Text variant="body2" lineClamp={2}>
+        <b>{reason}</b>
+        {separator}
+        {message}
+      </Text>
+    </Tooltip>
+  )
+}
+
+export default typeUnsafeMemoizeShallow(GridStatusMessageCell, {
```
xantorres, line 30: The `useGridRows` hook already memoizes cells by default, so we basically don't need to do any of this memoization in the cells themselves.
Maybe that's the reason you were getting a Typescript error to begin with, because of the double memoization

src/app/plugins/kubernetes/components/alarms/model.ts
```diff
@@ -39,15 +34,15 @@ export interface IAlertOverTime extends IGetPrometheusAlertsOverTime {
...
+  annotations: { [key: string]: string }
   clusterId: string
   clusterName: string
   description: string
   duration: number
   grafanaLink: string
   health: string
   id: string
-  labels: Labels
+  labels: { [key: string]: string }
```
xantorres, line 45: Again, I suggest using `Record`'s everywhere for consistency sake

src/app/core/components/grid/cell-components/GridStatusCell.tsx
jruddell: I dont see these as business specific. They take a generic common model and render this kind of item. This is going to be reused in other applications in the future (showing a status on an internal app for instance). So I see this as a good use case for our core package.

xantorres (approved):

src/app/core/elements/tooltip/Tooltip.tsx
```diff
@@ -45,27 +45,28 @@ export default function Tooltip({
...
-            {hasCustomBody ? customBody : message}
-          </Text>
-        ))}
+      {hasContent && (
+        <Text
+          variant="body2"
+          component="div"
+          className={clsx('tooltip', classes.tooltip, { [classes.tooltipBody]: !hasCustomBody })}
+        >
+          {hasCustomBody ? customBody : message}
```
xantorres, line 67: you could just do `customBody ?? message`

xantorres (approved):

PR #1570, opened 2022-03-02
Added Row and Column components for vertical and horizontal alignment

patil-pratik-87:
Screenshot 2022-03-02 at 2 08 57 PM
Screenshot 2022-03-02 at 2 09 32 PM

* Added Row and Column components for vertical and horizontal alignment instead of changing the `FormFieldSection` itself. The `Row` wraps the content automatically when needed. This seems more verbose and doesn't break the previous implementation.


xantorres:
Good job! I wouldn't define it as "more verbose" but rather as more readable and self-explanatory, which is great

xantorres:
I think these files should go into a "containers" folder as I see them as more utilitarian and general purpose and not necessarily form related.

kevinylee1991 (approved):
Looks cool

jruddell (approved):

patil-pratik-87:
> I think these components should go into a "containers" folder as I see them more as a utilitarian and general purpose elements and not necessarily form related.

okay.will move it "containers" folder

PR #1572, opened 2022-03-03
Added the data test id on the infrastructure tool bar icons

kingshuk-nandy:


xantorres:
We will no longer be using the ListTable component and its subcomponents

jruddell:
this isn't needed as this component is going away in 5.5

PR #1575, opened 2022-03-04
Updated the single step wizards in Workloads to ModalForm + Fixed too…

patil-pratik-87:
* Updated the single-step wizards in the Workloads section to ModalForm except for the Namespaces. It already had some issues. Will fix them in another PR.
* Fixed some tooltip issues
Screenshot 2022-03-04 at 5 55 42 PM
Screenshot 2022-03-04 at 5 56 04 PM
Screenshot 2022-03-04 at 5 56 17 PM


src/app/plugins/kubernetes/components/daemon-sets/DaemonSetsListPage.tsx
```diff
@@ -103,6 +105,13 @@ const columns: GridViewColumn<SelectorModel>[] = [
 ]
 
 export default function DaemonSetsListPage() {
+  const { match } = useReactRouter()
+  const isCreating =
```
jruddell, line 109: you shouldn't need this and this is also error prone. The existence of the create param should be sufficient enough no?

src/app/plugins/metal3/components/overview/InstallMetal3Form.tsx
```diff
@@ -6,7 +6,9 @@ import TextField from 'core/components/validatedForm/TextField'
 import { MetalLbCidrField } from 'k8s/components/infrastructure/clusters/cluster-addons/metal-lb'
 
 export default function InstallMetal3Form({ open, onClose }) {
-  const handleSubmit = (values, test) => {}
+  const handleSubmit = (values) => {
+    console.log(values)
```
jruddell, line 10: console.log

src/app/plugins/kubernetes/components/daemon-sets/DaemonSetsListPage.tsx
patil-pratik-87: After our discussion, I have modified the ModalForm to accept a route that operates the open state.
We do not need this logic now in the parent components now, hence removed it

src/app/core/elements/modal/Modal.tsx
```diff
@@ -6,9 +6,10 @@ import Theme from 'core/themes/model'
 import Text from 'core/elements/Text'
 import { hexToRgbaCss } from 'core/utils/colorHelpers'
 import DocumentMeta from 'core/components/DocumentMeta'
+import { Route } from 'core/plugins/route'
 
 export interface ModalProps {
-  open: boolean
+  open: Route | boolean
```
jruddell, line 12: My thought process here was to model this after how I did the tooltip props. Either an `open` prop or a `route` prop, require one or the other but not both. Let the modal if it gets a route track its isOpen flag via the route (`route.pattern.match(pathname)`).

src/app/core/elements/modal/ModalForm.tsx
```diff
@@ -16,10 +19,17 @@ export default function ModalForm({
...
   submitting = false,
+  open,
   ...props
 }: PropsWithChildren<Props>) {
   const submitFuncRef = useRef(null)
 
+  const { match } = useReactRouter()
+
+  if (open instanceof Route) {
+    open = open?.path() === match?.url
```
jruddell, line 30: Remember `open` is a prop in react and as such should be immutable. Think of it as if it is `protected`. Track the open value via a flag instead of mutating. `const isOpen = !!route ? route.pattern.match(location.pathname) : open`

patil-pratik-87: That is a stupid mistake, apologies will fix it

src/app/core/elements/modal/Modal.tsx
patil-pratik-87: done

src/app/core/elements/modal/ModalForm.tsx
```diff
@@ -4,32 +4,39 @@ import Modal, { ModalProps } from './Modal'
...
+interface BaseModalFormProps extends Omit<ModalProps, 'footer' | 'open'> {
   submitTitle?: string
   onSubmit?: any
   fieldSetter?: any
   submitting?: boolean
 }
 
+interface PropsWithOpenRoute extends BaseModalFormProps {
+  open?: undefined
+  openRoute: Route
```
jruddell, line 17: I'd rather we just call this `route`

src/app/core/elements/modal/ModalForm.tsx
```diff
@@ -38,9 +45,11 @@ export default function ModalForm({
     submitFuncRef.current.handleSubmit()
   }
 
+  const toOpen = openRoute ? useMemo(() => !!openRoute?.pattern.match(match.url), match.url) : open
```
jruddell, line 48: while this probably wont throw an error, its still not a good pattern to conditionally call a hook. React complains about a different number of hooks being rendered. Also the useMemo dependencies should be in an array. 

```
const isOpen = useMemo(() => !!route ? route.pattern.match(pathname) : open,  [route, open])
```

jruddell: Also `isOpen` is a better description than `toOpen`.

patil-pratik-87: I thought since this is an instruction that is being passed to the Modal, it's not already open, hence the `to` instead of 
'is'. Anywez just not a big thing, can change it

jruddell (approved):

PR #1576, opened 2022-03-05
Updated persistent volume claims to use ListContainer

kevinylee1991:
Also made a few changes on things unrelated to PVCs

Screen Shot 2022-03-04 at 5 54 54 PM


src/app/plugins/kubernetes/components/common/NamespacePicklist.tsx
```diff
@@ -32,8 +32,9 @@ export default function NamespacePicklist({
   value = allKey,
   ...rest
 }: NamespacePicklistProps) {
-  const { loading: namespacesLoading } = useListAction(listNamespaces)
+  const { loading: namespacesLoading } = useListAction(listNamespaces, { params: { clusterId } })
```
kevinylee1991, line 35: can you make sure this makes sense? Sometimes the namespaces would not load in until I added this

src/app/plugins/kubernetes/components/deployments/overview.tsx
```diff
@@ -76,7 +81,14 @@ const DeploymentOverview = ({ deployment, loading }) => {
     clusterId: deployment?.clusterId,
     namespace: deployment?.namespace,
   })
-  const [allPvcs, loadingPvcs] = useDataLoader(persistentVolumeClaimActions.list, {
+
+  const { loading: loadingPvcs } = useListAction(listPersistentVolumeClaims, {
+    params: {
+      clusterId: deployment?.clusterId, // namespace not required here right? is not an action param
```
kevinylee1991, line 87: namespace doesn't need to be passed to the action right? just to the selector?

xantorres: It depends on the specific api call, but generally yes. Another thing to keep in mind is, you can either use a selector maker to filter the data by namespace, or if you are using the Grid you don't need to do that. 
In any case, if you just filter by namespace and you dont need to do anything else (ie sorting, etc), probably using a selectorMaker is a bit overkill and you can filter the data locally with a simple `useMemo`, but that's up to you.

src/app/plugins/kubernetes/components/common/NamespacePicklist.tsx
xantorres: Yes, it makes sense. I must've forgotten that when refactoring to use the new Dropdown

src/app/plugins/kubernetes/components/storage/persistent-volume-claims/PersistentVolumeClaimsListPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+        columnKey: 'clusterId',
+        FilterComponent: ClusterPicklist,
+        onChange: getParamsUpdater('clusterId'),
+        controlled: true,
+      } as GridFilterSpec<SelectorModel, Params, 'clusterId'>,
+      {
+        columnKey: 'namespace',
+        FilterComponent: ({ filterValues, ...filterProps }) => (
+          <NamespacePicklist
+            // Why do we set allKey = "" here? allKey no longer works if we do this
```
xantorres, line 90: Not sure, is something legacy thing, if you have tested it and we don't need to keep it please feel free to remove it.

src/app/plugins/kubernetes/components/storage/persistent-volume-claims/PersistentVolumeClaimsListPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+export const pvcDetailOptions = {
+  title: 'Persistent Volume Claims',
+  columns: [
+    ...columns.filter((column) => !toFilterColumns[column.key]),
+    {
+      key: 'created',
+      label: 'Age',
+      render: (value) => <DateCell value={value} formatOpts={formatOpts} showToolTip difference />,
+    },
+  ].sort((a, b) => order.indexOf(a.key) - order.indexOf(b.key)),
```
xantorres, line 146: Please use the `columnsOrder` Grid prop for this

src/app/plugins/kubernetes/components/storage/persistent-volume-claims/PersistentVolumeClaimsListPage.tsx
kevinylee1991: Looks like this is affecting https://github.com/platform9/pf9-ui-plugin/commit/fd2892df9118a18a611dfb7fdf9971827c64da82#diff-9affbcb0c9b9c8924ca6e75dd8fea1a86e2c3f55ca2bb2db91f158f64caae942R32

Do we know why we added value = allKey here?

Also the namespacePicklist is being affected by the bug in useListAction file on the useIsDataLoadingSelector function. Where loading always starts as undefined, which is determined as boolean false. Which tricks namespacePicklist into thinking the namespaces are loaded when it has not loaded yet.

kevinylee1991: have a different PR trying to address this

jruddell (approved):

PR #1579, opened 2022-03-08
Converted user pages and user actions to newer architecture

kevinylee1991:
Screen Shot 2022-03-07 at 6 23 21 PM


src/app/core/actions/UpdateAction.ts
```diff
@@ -27,6 +27,7 @@ class UpdateAction<
         item: result,
       }),
     )
+    return result
```
kevinylee1991, line 30: Cool to add this here? Had to add it for receiving the response in the component

src/app/core/components/grid/Grid.tsx
```diff
@@ -149,15 +152,22 @@ export default function Grid<
     () => ({
       triggerRefresh: onRefresh,
       selectedItems: rowActionsProps.selectedItems,
+      clearSelectedRows: rowActionsProps.clearSelectedRows,
```
kevinylee1991, line 155: The table wasn't clearing after doing the custom batch actions, so I added this

src/app/core/components/grid/GridToolbar.tsx
```diff
@@ -61,6 +62,7 @@ export default function GridToolbar<
       </div>
       <div className={classes.gridRow}>
         {/* <div className={classes.verticalLine} aria-hidden="true" /> */}
+        {extraToolbarContent}
```
kevinylee1991, line 65: extra toolbar content for the system users toggler

src/app/core/components/grid/helpers/getGridDialogButton.tsx
```diff
@@ -8,12 +8,15 @@ export default function getGridDialogButton<T>(DialogComponent) {
     children,
     ...buttonProps
   }) => {
-    const { triggerRefresh, selectedItems } = useGridContext<T>()
+    const { triggerRefresh, selectedItems, clearSelectedRows } = useGridContext<T>()
+
     const [dialogOpened, toggleDialogOpened, setDialogOpened] = useToggler()
     const handleDialogClose = useCallback((success) => {
       if (success === true && triggerRefresh) {
+        clearSelectedRows()
```
kevinylee1991, line 16: clearSelectedRows used here

src/app/plugins/account/components/userManagement/users/new-actions.ts
```diff
@@ -339,6 +343,80 @@ export const deleteUser = userActions.add(
...
+  blob: string
+  type: string
+  user_id: string
+}
+
+interface IUserOptions {
+  multi_factor_auth_rules: string[][]
+}
+
+export const enableMfa = userActions.add(
```
kevinylee1991, line 356: I added the MFA actions here. At first I had these as standalone actions, but I wasn't able to get standalone custom actions to do anything in the component. Worked once I added it to an ActionsSet

xantorres: Did you set the proper configuration settings when creating the standalone custom actions? eg uniqueIdentifier, indexBy, cacheKey, etc

src/app/core/actions/UpdateAction.ts
xantorres: No, please change Action to return the result directly. We should not modify the result in any way before returning it to the view, since the only use of `postProcess` is to dispatch the relevant actions to update the cache. 
image

src/app/core/hooks/useUpdateAction.ts
```diff
@@ -87,6 +87,7 @@ const useUpdateAction = <
   // of the sequencing of multiple concurrent calls
   const update = useCallback(
     async (params = emptyObj) => {
+      let response = null
```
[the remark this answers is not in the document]
xantorres: This is incorrect, the result is already being assigned to the hook local state by using `localDispatch`, and that is being then sent to the view.

src/app/plugins/account/components/user-settings/MFASettingsPage.tsx
```diff
@@ -165,11 +178,19 @@ const MFASettingsPage = () => {
...
+  const { loading: loadingCredentials } = useListAction(listCredentials)
+  const credentials = useSelector(credentialsSelector)
+
+  const { loading: loadingUsers } = useListAction(listUsers)
+  const users = useSelector(usersSelector)
+
   const { params, updateParams, getParamsUpdater } = useParams<Params>(defaultParams)
   const { mfaSecret } = params
 
+  const { update: enable } = useUpdateAction(enableMfa)
```
xantorres, line 191: wouldn't be better to just have one action that receives a boolean param?

src/app/plugins/account/components/userManagement/users/EditUserPage.js
```diff
@@ -86,14 +87,25 @@ const EditUserPage = () => {
...
   const tenants = useSelector(tenantsSelector)
-  const [update, updating] = useDataUpdater(mngmUserActions.update)
-  const [roleAssignments, loadingRoleAssignments] = useDataLoader(mngmUserRoleAssignmentsLoader, {
-    userId,
-  })
+
+  // Question about the new actions here:
+  // Had to add this bc it was never getting called otherwise, why is that?
+  // Is it because user roles are not part of a ListContainer? Maybe
+  // ListContainer is handling the .call?
```
xantorres, line 102: `ListContainer` does not call any list action. If `useListAction` could be for multiple factors, including `userId` being null, improperly defined config, or some undiscovered/unknown bug, but the only way to find out is to debug the code.
So if you are unable to find the reason please add a @todo or @fixme statement here so we can investigate it later on.

src/app/plugins/account/components/userManagement/users/model.ts
```diff
@@ -1,8 +1,8 @@
 // TODO: fix these typings
```
xantorres, line 1: Can we remove this comment now?

xantorres (changes requested):

src/app/plugins/account/components/user-settings/MFASettingsPage.tsx
kevinylee1991: the two functions have different credential interface typings so I feel it may be better this way

xantorres (approved):

PR #1583, opened 2022-03-09
Create a modal to view configs for cluster addons

knnguy:
This change is for the addon cards in cluster edit.

Screen Shot 2022-03-09 at 2 25 04 PM

Screen Shot 2022-03-09 at 4 23 55 PM


src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/edit-cluster-page/ClusterAddonsPage.tsx
```diff
@@ -226,6 +233,9 @@ export default function ClusterAddonsPage({ clusterId, cluster, loading }) {
       {showDisableDialog && (
         <DisableAddonDialog addon={selectedAddon} open={open} onClose={toggleDisableDialog} />
       )}
+      {showConfigModal && (
```
knnguy, line 236: Is correct to have the parent component, `ClusterAddonsPage`, control the rendering of the modals? As opposed to having `ClusterAddonCard` do it?

jruddell: I think its better to do it this way so that way incase there is more than one card with a modal we aren't rendering extra modals when they're not needed to all be rendered

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/ConfigModal.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import CodeMirror from 'core/components/validatedForm/CodeMirror'
+import ValidatedForm from 'core/components/validatedForm/ValidatedForm'
+import Modal from 'core/elements/modal/Modal'
+import { decodeStr } from 'utils/misc'
+import Button from 'core/elements/button'
+
+export default function ConfigModal({ title, config, onClose }) {
```
jruddell, line 8: Why do you need a config modal? can't you use the ModalForm here?

jruddell (approved):

PR #1585, opened 2022-03-09
Add Kubevirt to addons page in cluster edit and add unknown as an addon health status

knnguy:
### Added Kubevirt to addons page in cluster edit
Screen Shot 2022-03-09 at 3 58 08 PM

### Added unknown as a health status
This is when an addon's `phase === undefined ` and` healthy === undefined`. Usually occurs right before the addon installation begins

Screen Shot 2022-03-09 at 4 02 40 PM

Screen Shot 2022-03-09 at 4 03 52 PM


jruddell:
lets not use the cluster status span anymore for a status. We can show the new badge status instead

jruddell (approved):

PR #1586, opened 2022-03-10
Fixed issue with selecting namespace as all in picklist, and where de…

kevinylee1991:
…fault namespace from user preference was being ignored


src/app/core/components/grid/useGridFiltering.ts
```diff
@@ -237,7 +238,7 @@ export default function useGridFiltering<
           valuesByKey[columnKey] !== undefined &&
           (allowEmpty || !isNilOrEmpty(valuesByKey[columnKey]))
         ) {
-          if (controlled) {
+          if (controlled || valuesByKey[columnKey] === allKey) {
```
kevinylee1991, line 241: without this, selecting All would make it filter into an empty list

src/app/plugins/kubernetes/components/common/NamespacePicklist.tsx
```diff
@@ -64,19 +67,12 @@ export default function NamespacePicklist({
   useEffect(() => {
     if (!clusterId) return
     // Reset the namespace when the clusterId changes
-    onChange(allKey)
+    onChange('')
```
kevinylee1991, line 70: I think we do not want to default to showing all namespaces right?

src/app/plugins/kubernetes/components/common/NamespacePicklist.tsx
```diff
@@ -64,19 +67,12 @@ export default function NamespacePicklist({
...
   }, [clusterId])
 
   // Select the first item as soon as data is loaded
   useEffect(() => {
     if (!clusterId || value) return
-
```
kevinylee1991, line 73: This part seems to be unnecessary with the new Dropdown component

src/app/plugins/kubernetes/components/rbac/profiles/create/roles-form.tsx
```diff
@@ -133,6 +133,8 @@ const RolesForm = ({ wizardContext, setWizardContext }) => {
       >
         <RolesListTable
           filters={
+            // fixme: This namespace picklist does not work with clusterId as allkey due to
+            // validateParams method in ListAction. Until fixed this will only have all option
             <NamespacePicklist
```
kevinylee1991, line 138: Low priority fix i think, can do it later

jruddell (approved):

PR #1589, opened 2022-03-10
Updated RBAC pages to use new actions

kevinylee1991:


src/app/core/hooks/useUpdateAction.ts
```diff
@@ -109,6 +109,7 @@ const useUpdateAction = <
         }
       } catch (error) {
         localDispatch({ type: 'endUpdate', payload: { success: false, error } })
+        throw error
```
jruddell, line 112: I dont think we should be throwing the error inside the hook. what is the point of this?

kevinylee1991: I put this here at the time because when I had a create role attempt error out, the form still redirected, which doesn't seem like good behavior

jruddell: This is requiring every place we use this hook to wrap it in a try catch so we handle an error.. or else you have a thrown uncaught exception that can really break the application. The proper way to handle this is to return a success out of the update hook that you can use to determine if you should redirect or not.

kevinylee1991: Ooh okay that makes sense, that matches what we did before

jruddell (approved):

PR #1594, opened 2022-03-15
Grid style fixes

xantorres:
Updated the Grid styles to match the new design guidelines.
Also added some minor fixes in other components such as dropdowns and inputs and added generic css classes (non jss) to some elements to be able to fine-tune their styling from parent containers.

image
image


src/app/core/elements/dropdown/Dropdown.tsx
```diff
@@ -192,13 +133,13 @@ export default function Dropdown<V, T extends DropdownItemSpec<V> = DropdownItem
               }}
               component="label"
               variant="inputLabel"
-              className={classes.label}
+              className={clsx(classes.label, 'label')}
```
jruddell, line 136: I'd prefer we include the element type in the class name so we dont have to worry too much about conflicts `dropdown-label`

jruddell (approved):

PR #1595, opened 2022-03-15
Updated user roles & group pages to use new actions

kevinylee1991:


src/app/plugins/account/components/userManagement/roles/UserRolesListPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+  { id: 'displayName', label: 'Role' },
+  { id: 'description', label: 'Description' },
+]
+
+export default function UserRolesListPage() {
+  const { params, getParamsUpdater } = usePrefParams({})
+  const { message, loading, reload } = useListAction(listMngmRoles)
+  const data = useSelector(managementRolesSelector)
+
+  // Using ListTable here still bc this table does not look good with Grid
```
jruddell, line 26: what do you mean by it doesn't look good?

xantorres: I think he means that it doesn't look as optimized since this table was used as a field instead of a table and it had some specifics like more compact appearance, less controls etc.
I'd say most of these things can be done using the new Grid (e.g. using the `compact` prop), but it's definitely something that still needs some tweaking and we'll need to do some extra work to get the Grid up to behave the way we want on this specific page.
Something I might look into, after working with the clusters list page.

kevinylee1991: The biggest thing is the limitation of the grid cell size -- the description is really large and takes up most of the table

jruddell: @kevinylee1991 lets look at it on a call then, I have a generic cell component that should handle this already

kevinylee1991: Screen Shot 2022-03-15 at 2 20 57 PM

src/app/plugins/account/components/ssoManagement/groups/new-actions.ts
```diff
@@ -111,14 +111,14 @@ export const createMngmGroup = mngmGroupActions.add(
 export const updateMngmGroup = mngmGroupActions.add(
   new UpdateAction<DataKeys.ManagementGroups, any, any>(
     async ({
-      groupId,
+      id,
```
xantorres, line 114: you could have used an alias here so you wouldn't need to alias the next `groupIds` to `id`

src/app/plugins/account/components/userManagement/groups/UserGroupsListPage.tsx
```diff
@@ -1,25 +1,57 @@
...
-const { ListPage: UserGroupsListPage } = createCRUDComponents(options)
+const columns: GridViewColumn<SelectorModel>[] = [
+  { key: 'id', label: 'OpenStack ID', display: false },
+  { key: 'name', label: 'Group Name' },
+  { key: 'description', label: 'Description' },
+]
 
-export default UserGroupsListPage
+export default function UserGroupsListPage() {
+  const { params, getParamsUpdater } = usePrefParams({})
```
xantorres, line 32: No need to pass an empty object here

kevinylee1991: Removing {} from usePrefParams causes an error -- maybe if we changed the usePrefParams function to initialize the defaultParms to emptyObj? But then that caused a ts error since emptyObj was not assignable to type P

xantorres: I would suggest initializing a `defaultParams` object outside of the render function (even if it is an empty object) as a generic way of doing this across all our views.

jruddell (approved):

PR #1598, opened 2022-03-15
Updated Add/Edit SAML groups to use ModalForms

patil-pratik-87:


patil-pratik-87:
Screenshot 2022-03-14 at 4 32 17 PM
Screenshot 2022-03-14 at 4 38 27 PM
Screenshot 2022-03-14 at 4 38 30 PM

jruddell (changes requested):
Dont use `FormFieldCard` for the different sections when inside a modal form. Use the `FormFieldSection` element instead

patil-pratik-87:
> Dont use `FormFieldCard` for the different sections when inside a modal form. Use the `FormFieldSection` element instead

done

jruddell (approved):

jruddell (changes requested):
Oh can we update the table here to use the new Grid as well?

patil-pratik-87:
> Oh can we update the table here to use the new Grid as well?

I was just targeting the Single Step Wizards but sure, why not?

patil-pratik-87:
> Oh can we update the table here to use the new Grid as well?

@jruddell while migrating to the new grid, found that this was a special case where batch actions are not needed but if not provided the grid doesn't show any selection like checkboxes or radios. Confirmed this with @xantorres and the solution will need a special GridField implementation which might take some time. Should we stick to the ListTable here for now?

jruddell:
> > Oh can we update the table here to use the new Grid as well?
> 
> @jruddell while migrating to the new grid, found that this was a special case where batch actions are not needed but if not provided the grid doesn't show any selection like checkboxes or radios. Confirmed this with @xantorres and the solution will need a special GridField implementation which might take some time. Should we stick to the ListTable here for now?

I understand the issue, and since we haven't gotten to the form component design/architecture, I don't want us to spend extra time trying to make the grid fit inside the existing form system. Can we instead just honor the colors from our theme system in the old ListTable till we get the forms architecture nailed down?

patil-pratik-87:
> > > Oh can we update the table here to use the new Grid as well?
> > 
> > 
> > @jruddell while migrating to the new grid, found that this was a special case where batch actions are not needed but if not provided the grid doesn't show any selection like checkboxes or radios. Confirmed this with @xantorres and the solution will need a special GridField implementation which might take some time. Should we stick to the ListTable here for now?
> 
> I understand the issue, and since we haven't gotten to the form component design/architecture, I don't want us to spend extra time trying to make the grid fit inside the existing form system. Can we instead just honor the colors from our theme system in the old ListTable till we get the forms architecture nailed down?

Yup...I have updated the ListTable with the theme colors and also upgraded the dropdown and input it uses

patil-pratik-87:
Screenshots for colours in different themes -->

Screenshot 2022-03-22 at 10 49 12 AM
Screenshot 2022-03-22 at 10 49 54 AM
Screenshot 2022-03-22 at 10 50 53 AM
Screenshot 2022-03-22 at 10 51 08 AM
Screenshot 2022-03-22 at 10 51 40 AM
Screenshot 2022-03-22 at 10 51 53 AM

xantorres:
Maybe remove the labels in each of the drop downs, it looks a bit too bloated

xantorres (approved):

jruddell (approved):

PR #1604, opened 2022-03-16
UX-1679/Fixed edit Helm deployed app bug

knnguy:
**Issue:**
User had 2 deployed apps with the same name, `"cert-manager-pf9"` under the same cluster and the same namespace. However, they were under different repositories and charts. One was `"cert-manager-pf9"` from  `"jetstack/cert-manager"` and one was `"cert-manager-pf9"` from `"bitanmi/cert-manager"`. 

When the user went to edit `"cert-manager-pf9"` from  `"jetstack/cert-manager"`, the PUT request was made to `"bitanmi/cert-manager"`.  This happened because in the edit deployed apps screen, I was only finding the deployed app using the clusterId, namespace, and name

**Fix:**
Find the deployed app using the clusterId, namespace, repository, chart and name


knnguy:
NOTE: Could not test this at the moment. The releases API is down. Got a 500 error

jruddell (approved):

PR #1605, opened 2022-03-16
UX-1602/Added Metal3 option for existing clusters in Metal3 plugin page. Adde…

knnguy:
…d Metal3 to addons page in cluster edit.

### Added functionality to the 'Select an Existing Cluster' button
Screen Shot 2022-03-16 at 3 41 21 PM

Screen Shot 2022-03-16 at 2 55 15 PM

### Metal3 add-on card in the add-ons page in cluster edit
Screen Shot 2022-03-16 at 11 28 31 AM


src/app/plugins/kubernetes/components/infrastructure/nodes/NodeInterfacesDropdown.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const { loading: importedClustersLoading } = useListAction(listImportedClusters)
+  const clusters = useSelectorWithParams(clustersSelector, defaultParams)
+  const cluster = useMemo(() => clusters.find((cluster) => cluster.uuid === clusterId), [
+    clusterId,
+    clusters,
+  ])
+  const options = useMemo(() => {
+    const nodes = cluster?.nodes
+    if (isNilOrEmpty(nodes)) return []
+    const nodeInterfaces = flatten(nodes.map((node) => Object.keys(node.networkInterfaces || {})))
```
jruddell, line 28: if you're using ramda methods to flatten, you should be able to use uniq here as well, would be more optimal than doing it in a separate step

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/param-fields/Metal3ParamFields.tsx
```diff
@@ -0,0 +1,69 @@
...
+        DropdownComponent={NodeInterfacesDropdown}
+        id={Metal3Params.Metal3DhcpInterface}
+        name={Metal3Params.Metal3DhcpInterface}
+        label={paramDisplayNames[Metal3Params.Metal3DhcpInterface]}
+        clusterId={values?.clusterId}
+        onChange={setFieldValue(Metal3Params.Metal3DhcpInterface)}
+        value={values?.Metal3DhcpInterface}
+        required
+      />
+      <TextField
```
jruddell, line 21: the range and gateway in the design are side by side, we can use the `Row` component to accomplish that

image

jruddell (approved):

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/helpers.ts
```diff
@@ -219,6 +245,7 @@ export const getClusterAddonBody = ({
     name: key,
     value: value,
   }))
+  console.log(type, overrideParamsList)
```
xantorres, line 248: console.log here

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/edit-cluster-page/AddonConfigurationsModal.tsx
```diff
@@ -22,7 +29,7 @@ const getTitle = (addonType, action) => {
 const stripUnitFromValue = (value) => {
   const hasNumber = /\d/.test(value)
   if (!hasNumber) return value
-  return parseInt(value, 10)
+  return parseFloat(value)
 }
 
 const getOverrideParams = (addonType, fieldValues, cluster) => {
```
xantorres, line 35: this const below is a pointer to fieldValues so you are mutating the original variable which is probably not what you want.
To make this a pure function you could do something like 
`const overrideParams = {...fieldValues}`

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/edit-cluster-page/AddonConfigurationsModal.tsx
```diff
@@ -22,7 +29,7 @@ const getTitle = (addonType, action) => {
 const stripUnitFromValue = (value) => {
```
xantorres, line 29: We could put this into the `misc.ts` utilities

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/edit-cluster-page/AddonConfigurationsModal.tsx
```diff
@@ -41,6 +48,8 @@ interface Props {
   action: 'Enable' | 'Edit'
 }
 
+const paramsWithUnitsAttached = new Set(Object.keys(paramsWithUnits))
+
 export default function AddonConfigurationsModal({ addon, open, onClose, action }: Props) {
   const formFieldSetter = useRef(null)
```
xantorres, line 54: This formFieldSetter stuff is really overengineered, perhaps we could just create a subcomponent that gets `setFieldValue` and just use that instead of doing all of this.

src/app/plugins/kubernetes/components/storage/storage-classes-picklist.tsx
```diff
@@ -18,52 +19,40 @@ interface Props {
...
-      value,
-      clusterId,
-      name = 'storageClass',
-      label = 'Storage Class',
-      selectFirst = false,
-      formField = true,
-      showNone = true,
-      ...rest
-    } = props
+export default function StorageClassesPicklist(props: Props) {
```
xantorres, line 22: It would be nice to rename this file in PascalCase to match the exported function.

src/app/plugins/metal3/components/overview/InstallMetal3Form.tsx
```diff
@@ -1,60 +1,90 @@
...
+const { qbert } = ApiClient.getInstance()
+
+const bareOsClustersOnly = (clusters) =>
+  clusters.filter((cluster) => cluster.cloudProviderType === CloudProviders.BareOS)
 
 export default function InstallMetal3Form({ open, onClose }) {
-  const handleSubmit = (values) => {}
+  const { update: enableAddon, updating: enablingAddon } = useUpdateAction(createClusterAddon)
+
+  const handleSubmit = useCallback(async (values) => {
```
xantorres, line 28: I think all this logic should be exported to an actions file, even if you are not using the Action class. The view should not handle all this logic.

src/app/core/elements/modal/ModalForm.tsx
```diff
@@ -10,6 +10,7 @@ interface BaseModalFormProps extends Omit<ModalProps, 'open'> {
   onSubmit?: any
   fieldSetter?: any
   submitting?: boolean
+  initialValues?: any
```
xantorres, line 13: could you use `Record<string, unknown>` here? We should stop using `any`

PR #1610, opened 2022-03-21
Added flow for create namespace and added data test ids or common methods.

Komaljadhav12:
create-namespace


src/app/core/components/grid/GridRow.tsx
```diff
@@ -4,6 +4,8 @@ import { SelectableParsedGridRow } from './useGridBatchActions'
 import Radio from 'core/elements/input/Radio'
 import { noop } from 'utils/fp'
 import { memoizeShallow } from 'utils/misc'
+import generateTestId from 'utils/test-helpers'
+import { Title } from '@material-ui/icons'
```
jruddell, line 8: why are you importing a material icon here?

Komaljadhav12: Removed unused import from the file.

jruddell (approved):

PR #1616, opened 2022-03-23
Update cloud provider screens to use updated actions

kevinylee1991:


src/app/core/hooks/useUpdateAction.ts
```diff
@@ -109,10 +110,12 @@ const useUpdateAction = <
           localDispatch({ type: 'endUpdate', payload: { result, success: true, error: null } })
         }
       } catch (error) {
+        updaterPromisesBuffer.current.shift() // Delete the oldest promise in the sequence (FIFO)
```
kevinylee1991, line 113: Can you confirm this makes sense? I needed to add this so that when a create/update request failed, then I would be able to resend it from the same component.

src/app/core/hooks/useUpdateAction.ts
```diff
@@ -109,10 +110,12 @@ const useUpdateAction = <
...
         localDispatch({ type: 'endUpdate', payload: { success: false, error } })
+        success = false
       }
       dispatch(cacheActions.setUpdating({ cacheKey, updating: false }))
-      return response
+      return { success, response }
```
kevinylee1991, line 118: Can confirm if this return type is ok?

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/AwsCloudProviderVerification.tsx
```diff
@@ -121,10 +125,18 @@ const AwsCloudProviderVerification = ({
...
+    },
+    requiredParams: ['cloudProviderId'],
+  })
+  const regions = useSelectorWithParams(cloudProviderDetailsSelector, {
     cloudProviderId: wizardContext.cloudProviderId,
   })
 
+  // Todo: Swap cloudProviderRegionDetails to new actions
+  // Do not know how to use this custom action like a list action
   const [details, loading] = useDataLoader(loadCloudProviderRegionDetails, {
```
kevinylee1991, line 140: I didn't update this one because I couldn't figure out how to treat a custom action like a list action

kevinylee1991: or i had some difficulty figuring out how to use it in the component -- since we don't have a UseCustomAction hook

xantorres: You have to use the `useUpdateAction` hook

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/CloudProvidersListPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+}
+
+const usePrefParams = createUsePrefParamsHook<TablePrefsParams>('Cloud Providers', listTablePrefs)
+
+const searchTargets = ['name', 'descriptiveType']
+
+export const columns: GridViewColumn<SelectorModel>[] = [
+  { key: 'name', label: 'Name' },
+  { key: 'descriptiveType', label: 'Type' },
+  { key: 'deployedCapacity', label: 'Deployed Capacity', render: DeployedCapacityCell },
```
jruddell, line 131: shouldn't this be `CellComponent` instead of `render`?

src/app/plugins/kubernetes/components/infrastructure/cloudProviders/UpdateCloudProviderPage.tsx
```diff
@@ -67,133 +70,142 @@ const formCpBody = (data) => {
...
+      cloudProviderId: cloudProvider.uuid,
     }
-  }, [initialValues.uuid])
+  }, [cloudProvider])
 
-  const updateCloudProvider = (data) => {
+  const submitForm = async (data) => {
     const body = formCpBody(data)
-    return onComplete(body)
+    await update({
```
jruddell, line 109: Now that you have success, you should probably only change routes here if it was successful

jruddell (approved):

PR #1617, opened 2022-03-24
UX-1602/Added tooltip descriptions to Metal3 params, changed the labels, and …

knnguy:
…changed the `DHCP Range` input textfield to a KeyValues one

Screen Shot 2022-03-24 at 10 45 27 AM


src/app/core/components/KeyValues.js
```diff
@@ -10,14 +10,16 @@ import Text from 'core/elements/Text'
 const useKeyValueStyles = makeStyles((theme) => ({
   root: {
     display: 'grid',
-    gridTemplateColumns: '200px 6px 200px 32px',
+    // gridTemplateColumns: '200px 6px 200px 32px',
+    gridTemplateColumns: ({ showDeleteButton }) =>
+      showDeleteButton ? '172px 6px 172px 32px' : '185px 6px 185px',
```
knnguy, line 15: I changed the grid columns here because our form input fields are  set to 400px but the KeyValues component goes over 400px so the input fields weren't aligned

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/param-fields/Metal3ParamFields.tsx
```diff
@@ -73,10 +86,7 @@ export default function Metal3ParamFields({ setFieldValue, values, showStorageCl
 }
 
 const useStyles = makeStyles<Theme>((theme) => ({
-  rowLayout: {
-    display: 'grid',
-    maxWidth: '400px', // 400 is the max with of the input component
-    gridAutoFlow: 'column',
-    gridGap: theme.spacing(1),
+  picklistField: {
```
knnguy, line 89: Had to add this here because the tooltip component wraps around the Picklist component and has a width of max-content. I had to set the width to 400px to match the textfield inputs

jruddell: Could it be a maxWidth instead of width? this way it can shrink if the modal does

src/app/core/components/KeyValues.js
```diff
@@ -129,15 +134,18 @@ const KeyValues = ({
...
   const classes = useStyles()
-  const [entries, setEntries] = useState([newEntry()])
+  const [entries, setEntries] = useState(allowMultipleValues ? [newEntry()] : [])
   const [loadedInitialEntries, setLoadedInitialEntries] = useState(false)
 
   const addBlankEntry = () => setEntries([...entries, newEntry()])
   const deleteEntry = (id) => () => setEntries(entries.filter((x) => x.id !== id))
   const handleChange = (entry) => setEntries(entries.map((x) => (x.id === entry.id ? entry : x)))
 
+  console.log(keyLabel, entries)
```
jruddell, line 147: console.log

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/param-fields/Metal3ParamFields.tsx
knnguy: This is what happens when I don't set a fixed with on the PXE Interface Name picklist. 

Screen Shot 2022-03-25 at 10 08 21 AM

src/app/core/components/KeyValues.js
```diff
@@ -129,22 +133,26 @@ const KeyValues = ({
   addLabel,
   keyLabel,
   valueLabel,
+  allowMultipleValues = true,
 }) => {
   const classes = useStyles()
   const [entries, setEntries] = useState([newEntry()])
```
xantorres, line 139: Why are we creating a new `entry` object every time? Can't we just reuse the same object?

src/app/core/components/KeyValues.js
```diff
@@ -129,22 +133,26 @@ const KeyValues = ({
...
+  console.log(keyLabel, entries)
+
   useEffect(() => {
     // Load initial entries
     if (loadedInitialEntries || _entries === undefined) return
     const initialEntries = getInitialEntries(entries, _entries)
     setLoadedInitialEntries(true)
     if (initialEntries.length === 0) return
-    setEntries([..._entries, ...entries])
+    allowMultipleValues ? setEntries([..._entries, ...entries]) : setEntries([..._entries])
```
xantorres, line 155: The `_entries` variable name is not very intuitive, could we rename it to something like `initialEntries`?

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/edit-cluster-page/AddonConfigurationsModal.tsx
```diff
@@ -59,6 +61,10 @@ export default function AddonConfigurationsModal({ addon, open, onClose, action
         return formFieldSetter?.current?.setField(key)(parseMetalLbCidr(value))
       }
 
+      if (key === Metal3Params.Metal3DhcpRange) {
+        return formFieldSetter?.current?.setField(key)(parseMetal3IpAddressRange(value))
```
xantorres, line 65: You could have assigned `formFieldSetter?.current?.setField(key)` to a variable (eg `fieldSetter`) and reuse it everywhere instead of recalling this every time.

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/param-fields/Metal3ParamFields.tsx
jruddell: `width: 100%; max-width: 400px`?

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/edit-cluster-page/AddonConfigurationsModal.tsx
knnguy: Yep, I changed this in my most recent PR

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/param-fields/Metal3ParamFields.tsx
knnguy: Yes, have tried that already.

jruddell (approved):

PR #1628, opened 2022-03-29
UX-1193/Add add-ons page for imported clusters

knnguy:
### Add-ons page is now added to import clusters edit page
Screen Shot 2022-03-29 at 2 01 00 PM

### Add-ons page is also now added to import cluster details page
Screen Shot 2022-03-29 at 2 00 42 PM


jruddell:
why is the text color not matching the theme in the addons table?

jruddell (approved):

PR #1631, opened 2022-03-30
UX-1531/ Updated the header and details card designs for clusters and nodes. R…

knnguy:
…emoved add-on details in the cluster details card

Changes Made: 

- Removed ETCD backup info card and the load balance details from the cluster details page. Add-on info should only be shown in the add-ons page.
- Updated the design of the node and cluster header cards as well as the cluster and node detail cards to match the current design

### Cluster Details
Screen Shot 2022-03-30 at 4 25 10 PM

### Imported Cluster Details
Screen Shot 2022-03-31 at 2 31 13 PM

### Node Details
Screen Shot 2022-03-30 at 5 41 10 PM


knnguy:
@jruddell Not sure what the design for the cluster and node details should be, but since I was removing some stuff from the detail cards, I thought I would just fix up the header and detail cards so that it matches the current design

knnguy:
In the design, the links in the header cards are not blue, but all our current links are so I'm not sure if I should leave as it or change it to the new design

jruddell (approved):

PR #1637, opened 2022-04-01
UX-1710/Change add storage class page to a modal instead of a 2 step wizard

knnguy:
Screen Shot 2022-04-01 at 3 21 25 PM


src/app/plugins/kubernetes/components/storage/storage-classes/add-page/AddStorageClassModal.tsx
```diff
@@ -0,0 +1,257 @@
...
+  const selectedCluster = getSelectedCluster(clusters, values.clusterId)
+
+  useEffect(() => {
+    if (selectedCluster) {
+      setFieldValue('cloudProviderType')(selectedCluster.cloudProviderType)
+    }
+  }, [selectedCluster])
+
+  useEffect(() => {
+    const newStorageClassYaml = getInitialStorageClassYaml(contextWithoutYaml)
```
jruddell, line 106: I know this is a direct translation over so we can leave this here for now... but the yaml editor shouldn't be as controlled in this way. It creates some weird behavior when deleting items in the yaml file.

jruddell (approved):

PR #1638, opened 2022-04-01
Fixed the way active region is loaded

kevinylee1991:


src/api-client/helpers.ts
```diff
@@ -37,8 +38,15 @@ export const makeScopedClient = async (tenantName = defaultTestTenant) => {
...
-  store.dispatch(sessionActions.updateSession({ activeRegion }))
+export const setActiveRegion = async (activeRegion) => {
+  const { keystone } = ApiClient.getInstance()
+  store.dispatch(sessionActions.updateSession({ activeRegion, loadingRegion: true }))
+  await keystone.resetCookie()
+  await ApiClient.refreshApiEndpoints()
+  await store.dispatch(cacheActions.clearCache())
+  const features = await keystone.getFeatures()
+  // Store entirety of features json in context for global usage
+  store.dispatch(sessionActions.updateSession({ features, loadingRegion: false }))
```
jruddell, line 49: can any of these actions throw an exception that needs to be caught here (so we can avoid an infinite loading state)?

jruddell (approved):

PR #1639, opened 2022-04-01
Update storage class overview page to new design

knnguy:
Screen Shot 2022-04-01 at 4 38 05 PM


knnguy:
I was trying to fix a blank screen bug for storage class overview but then realized that the page was using the old design so I went ahead and updated it. Let me know if there's an actual design for it and if I need to update what we have now

jruddell (approved):

PR #1646, opened 2022-04-06
UX-1602/Create Metal3 Dashboard page

knnguy:
Screen Shot 2022-04-06 at 2 39 35 PM


src/app/core/containers/ListContainer.tsx
```diff
@@ -118,6 +119,7 @@ export default function ListContainer<D extends keyof IDataKeys, T = ArrayElemen
     showBreadcrumbs = true,
     extraHeaderContent,
     loading,
+    label,
```
jruddell, line 122: probably best to just default the label here. `label = entityNamesByKey[dataKey]` and let a direct label prop override this.

src/app/core/containers/ListContainer.tsx
```diff
@@ -258,7 +260,7 @@ export default function ListContainer<D extends keyof IDataKeys, T = ArrayElemen
       )}
       <Grid
         {...props}
-        label={dataKey ? entityNamesByKey[dataKey] : null}
+        label={label ? label : dataKey ? entityNamesByKey[dataKey] : null}
```
jruddell, line 263: with the above change this just turns into `label={label}`

src/app/core/hooks/useListAction.ts
```diff
@@ -52,9 +52,14 @@ const useIsDataLoadingSelector = (...cacheKeys): string[] => {
   }, [cacheKeys, keysLoadingStatus])
 }
 
+// const getIndexedParams = memoize(
```
jruddell, line 55: Is this supposed to be a workaround to the allkey issue? I think @kevinylee1991 is looking into this

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/cluster-details-page/ClusterAddonsPage.tsx
```diff
@@ -43,7 +43,7 @@ const usePrefParams = createUsePrefParamsHook<ActionParams & TablePrefsParams>(
   listTablePrefs,
 )
 
-const getAddonHealthStatus = (addon) => {
+export const getAddonHealthStatus = (addon) => {
```
jruddell, line 46: should this be inside a helper file? I dont want us importing methods between different view files as this can end up with circular imports

src/app/plugins/metal3/model.ts
```diff
@@ -0,0 +1,333 @@
...
+  resourceVersion: string
+  selfLink: string
+  uid: string
+}
+
+export interface Annotations {
+  'kubectl.kubernetes.io/last-applied-configuration': string
+}
+
+export interface ManagedField {
```
jruddell, line 49: please remove the managed field typings. They are pointless and add a lot of extra indexing in the typescript system

src/app/core/hooks/useListAction.ts
knnguy: That was a comment that I forgot to delete

However, about the `allKey` issue, I was able to get it to work by not filtering out the `allKey` in `getIndexedParams`

```
const getIndexedParams = memoize(
  <T>(indexBy: string | string[], params: T): T =>
    pipe(pickAll(ensureArray(indexBy)), reject(either(isNil, equals(allKey))))(params),
)
```

So instead of 
`pipe(pickAll(ensureArray(indexBy)), reject(either(isNil, equals(allKey))))(params),`

I changed it to
` pipe(pickAll(ensureArray(indexBy)), reject(isNil))(params),`

@xantorres I did what you suggested as a workaround for the `clusterId = allKey` issue by setting `requiredParams = ['clusterId']` but it still didn't work. Only this did.

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/cluster-details-page/ClusterAddonsPage.tsx
knnguy: Yep. This is actually not being used in any other files right now so I just removed the export

src/app/plugins/metal3/actions.ts
```diff
@@ -41,3 +48,21 @@ export const enableMetal3 = async (clusterId, values, enableAddonFn, checkForMet
...
+  indexBy: 'clusterId',
+  entityName: entityNamesByKey.BareMetalHosts,
+  cacheKey: DataKeys.BareMetalHosts,
+})
+
+export const listBareMetalHosts = bareMetalHostActions.add(
+  new ListAction<DataKeys.BareMetalHosts, { clusterId: string }>(async (params) => {
+    Bugsnag.leaveBreadcrumb('Attempting to get Bare Metal Hosts')
+    await listClusters.call({})
+    const [clusterId, clusters] = await parseClusterParams(params)
```
knnguy, line 63: I just noticed a bug where sometimes after a refresh, when `parseClusterParams` gets called, clusters haven't been loaded into the cache yet so then `clusters = []` which then this function returns nothing. As a result, in the metal3 table, the columns that come from the bareMetalHosts are empty. This only happens occasionally after a refresh

Screen Shot 2022-04-07 at 12 49 23 PM

To ensure that clusters are always loaded first, I added the line `await listClusters.call({})`. Not sure if this is the correct approach, but it works.

knnguy: I have move the fix for this issue to `parseClusterParams` instead. If the clusters have not been loaded into cache yet, then we refetch the clusters to ensure that the `parseClusterParams` will always return the list of clusters available

src/app/core/hooks/useListAction.ts
jruddell: go ahead and remove this since @kevinylee1991 is also doing the same work

jruddell (approved):

PR #1647, opened 2022-04-07
Cluster Details Nodes Tab

jruddell:
image


jruddell:
I didn't update the rendering for each column yet as @xantorres is already doing this same work for the NodesListPage.

jruddell:
This was added in Xan's PR so we can close it out

PR #1661, opened 2022-04-11
No compact dropdowns in create operations in workloads screens

patil-pratik-87:
Screenshot 2022-04-11 at 12 14 35 PM


src/app/core/components/InfoTooltip.js
```diff
@@ -1,38 +1,8 @@
-import { withStyles } from '@material-ui/styles'
 import Tooltip from 'core/elements/tooltip'
 import PropTypes from 'prop-types'
 import React, { PureComponent } from 'react'
 import { memoize } from 'utils/misc'
 
-const styles = (theme) => ({
```
patil-pratik-87, line 7: These styles are not needed and were interfering with the theme colours. Now it picks up the tooltip styling
Probably the `InfoTooltip` is not needed at all.

src/app/plugins/kubernetes/components/common/entity/AddResourcePage.tsx
```diff
@@ -35,6 +34,8 @@ import { createDeploymentLink, createPodLink, createServiceLink } from 'k8s/link
 import SimpleLink from 'core/components/SimpleLink'
 import ModalForm from 'core/elements/modal/ModalForm'
 import { Route } from 'core/plugins/route'
+import Column from 'core/containers/Column'
```
jruddell, line 37: is this used?

src/app/core/components/InfoTooltip.js
jruddell: this is used in many places, did you check a few of the create flows? we should verify it isn't conflicting with other things

jruddell:
Does this cover all of the forms presented inside this modal? what you've changed currently is just for workloads, but the ModalForm extends to other views if I'm not mistaken

src/app/core/components/InfoTooltip.js
patil-pratik-87: Should not conflict with other places, still lemme check

src/app/plugins/kubernetes/components/common/entity/AddResourcePage.tsx
patil-pratik-87: nope, removed

jruddell (approved):

PR #1675, opened 2022-04-13
UX-1610/Fixed a bug when there is no default namespace already chosen

jruddell:


src/app/plugins/kubernetes/components/namespaces/AddNamespacePage.js
```diff
@@ -21,15 +21,15 @@ export default function AddNamespaceForm() {
   const { history } = useReactRouter()
   const { params, getParamsUpdater, setParams } = useParams(defaultParams)
 
-  const { update, updating, error, clearAll } = useUpdateAction(createNamespace)
+  const { update, updating, error, reset } = useUpdateAction(createNamespace)
```
patil-pratik-87, line 25: oops, did i forget about this? thanks for fixing it :)

patil-pratik-87 (approved):

src/app/plugins/kubernetes/components/namespaces/AddNamespacePage.js
jruddell: Yep, I noticed it while fixing the other thing. Went ahead and searched around to verify it was the only place and it was... so no further action is needed 👍

PR #1694, opened 2022-04-15
UX-1769/Allow viewing of published RBAC profiles, and RBAC profile bu…

kevinylee1991:
…gfixes

To view details I added a link for the name column:
Screen Shot 2022-04-14 at 7 19 27 PM

The read-only version doesn't have CRUD features, but reuses the same view as the edit screen
Screen Shot 2022-04-14 at 7 19 54 PM

Roles:
Screen Shot 2022-04-14 at 7 20 17 PM
Screen Shot 2022-04-14 at 7 20 26 PM

Cluster Roles:

I borrowed the key/value display from alarms. This doesn't look too good, open to any tweaks but I think we should have a proper design for this. I also updated the buttons to all say Close instead of Cancel (the role bindings screenshots are outdated)

Screen Shot 2022-04-14 at 7 20 36 PM
Screen Shot 2022-04-14 at 7 20 42 PM

Role Bindings:
Screen Shot 2022-04-14 at 7 21 27 PM
Screen Shot 2022-04-14 at 7 21 34 PM

Cluster Role Bindings:
Screen Shot 2022-04-14 at 7 21 58 PM
Screen Shot 2022-04-14 at 7 22 04 PM


src/app/plugins/kubernetes/components/common/NamespacePicklist.tsx
```diff
@@ -67,9 +70,16 @@ export default function NamespacePicklist({
 
   useEffect(() => {
     if (!clusterId) return
+    // If initialValue was being set, it would just be overridden here
```
kevinylee1991, line 73: I added something so that if initialValue is set, the namespace won't get reset. Added this bc otherwise this would get set to '' every time & the initial value wouldn't stick or the value would be '' even though the picklist looked like it had a value.

jruddell (approved):

PR #1701, opened 2022-04-15
UX-1574/ trimmed trailing whitespaces in Input

patil-pratik-87:
https://user-images.githubusercontent.com/63217589/163549485-9f88488c-6d68-4e7d-8aca-3cabe2298528.mov


patil-pratik-87:
Thanks @jruddell  for the tip 👍🏽

src/app/core/elements/input/Input.tsx
```diff
@@ -65,6 +65,15 @@ const Input = ({
     [onChange],
   )
 
+  const handleFocusOut = useCallback(
+    (event) => {
+      event.target.value = event?.target?.value?.trim()
```
jruddell, line 70: I'm really not a fan of this. We should think about changing this model to surface out the value only and or the event as a second param

src/app/core/elements/input/Input.tsx
```diff
@@ -65,6 +65,15 @@ const Input = ({
...
+      onChange(event)
+      toggleFocused()
+    },
+    [toggleFocused],
+  )
+
   const handleFocusToggle = useCallback(() => {
```
jruddell, line 77: is this fn still used?

jruddell (approved):

src/app/core/elements/input/Input.tsx
patil-pratik-87: Yeah, its a `onFocus` handler

src/app/core/elements/input/Input.tsx
patil-pratik-87: yeah, did true, did not wanna make a lot of changes at this point to pass in just the value for the the fear of regression, hence this...

PR #1708, opened 2022-04-15
UX-1804/Update deployed apps page in cluster details

knnguy:
Screen Shot 2022-04-15 at 4 31 22 PM

Screen Shot 2022-04-15 at 4 32 54 PM


knnguy:
When the user clicks on the 'Connect Repository' button, a list of available repositories (repositories that haven't been connected the to the cluster yet) will be displayed. When the user clicks on the repository, the edit option will be available and if they click on the 'Edit' button, they will be taken to the edit form for that repository. From there, they can choose to attach any cluster they want.

Is this behavior ok for now? Old behavior was that the user gets a table with available repositories and they are able to select the repositories from the table. Then they can connect the repositories from the dialog.

I am currently not sure if our new list tables allows us to do this so it'll take me a bit to figure out. Since this is due by EOD, is it ok to have this new behavior instead?

### Current behavior
https://user-images.githubusercontent.com/23369276/163652632-bea818fd-5496-4765-9117-239c3554a95b.mov

### Old behavior

Screen Shot 2022-04-15 at 4 40 39 PM

jruddell:
> When the user clicks on the 'Connect Repository' button, a list of available repositories (repositories that haven't been connected the to the cluster yet) will be displayed. When the user clicks on the repository, the edit option will be available and if they click on the 'Edit' button, they will be taken to the edit form for that repository. From there, they can choose to attach any cluster they want.
> 
> Is this behavior ok for now? Old behavior was that the user gets a table with available repositories and they are able to select the repositories from the table. Then they can connect the repositories from the dialog.
> 
> I am currently not sure if our new list tables allows us to do this so it'll take me a bit to figure out. Since this is due by EOD, is it ok to have this new behavior instead?
> 
> ### Current behavior
>  Screen.Recording.2022-04-15.at.4.35.34.PM.mov 
> ### Old behavior
> Screen Shot 2022-04-15 at 4 40 39 PM

Yes this fine

src/app/plugins/kubernetes/components/infrastructure/clusters/deployed-apps/ConnectRepositoriesCard.tsx
```diff
@@ -0,0 +1,73 @@
...
+import generateTestId from 'utils/test-helpers'
+import Button from 'core/elements/button'
+
+export default function ConnectRepositoriesCard({ handleConnectRepositoryClick }) {
+  const classes = useStyles()
+  return (
+    <Card
+      title="Repositories"
+      footer={
+        <footer className={classes.connectRepositoriesFooter}>
```
jruddell, line 17: Don't render a footer element here as the card also renders a footer... Might be better to just render the card footer directly here

jruddell (approved):

PR #1718, opened 2022-04-19
cleaned up a majority of types that were exported but not used

jruddell:
### Before
image
image
image

### After
image
image
image


patil-pratik-87:
wow 🤩 , that's crazy. Thanks for cleaning it up dude :)

src/app/plugins/metal3/model.ts
```diff
@@ -165,12 +160,12 @@ export interface Provisioning {
   state: string
 }
 
-export interface RAID {
-  hardwareRAIDVolumes: null
-  softwareRAIDVolumes: any[]
+interface RAID {
```
patil-pratik-87, line 163: should this be `unknown` rather?

src/app/plugins/kubernetes/components/storage/persistent-volume/model.ts
```diff
@@ -49,14 +44,14 @@ export interface Spec {
   persistentVolumeReclaimPolicy: string
   storageClassName: string
   volumeMode: string
-  nodeAffinity: NodeAffinity
+  nodeAffinity: any
```
patil-pratik-87, line 47: should this be `unknown` rather?

patil-pratik-87 (approved):

src/app/plugins/metal3/model.ts
jruddell: I'd rather strongly type it correctly by looking at the k8s models. unknown can cause ts errors

patil-pratik-87: Yeah, it will cause errors until the consumer does a type check or a type assertion while using the values. I thought it will enforce folks to either update the model or do a type assertion while using the values.

PR #1721, opened 2022-04-19
UX-1816/updated the azs in req payload to an array in one-click aws

patil-pratik-87:


src/app/plugins/kubernetes/components/infrastructure/clusters/aws/aws-availability-zone.tsx
```diff
@@ -13,7 +13,7 @@ const AwsAvailabilityZoneField = ({
     info="Select from the Availability Zones for the specified region"
     cloudProviderId={values.cloudProviderId}
     cloudProviderRegionId={values.region}
-    onChange={(value) => setWizardContext({ azs: Array.isArray(value) ? value : [value] })}
```
patil-pratik-87, line 16: The onChange was triggered twice here and the second time with the already set [],  hence the check failed

PR #1732, opened 2022-04-20
Made some small fixes to RBAC and RBAC profiles pages

kevinylee1991:
Moved summary card info into wizard summary

Screen Shot 2022-04-21 at 1 30 20 PM


src/app/plugins/kubernetes/components/rbac/profiles/drift/DriftAnalysisPage.tsx
```diff
@@ -65,9 +65,9 @@ const DriftAnalysisPage = () => {
               stepId="step1"
               label="Select a Cluster"
               onNext={() => trackEvent('Drift Analysis - Finished Drift Analysis')}
-              summaryTitle="New Drift Analysis"
-              summaryKeyOverrides={wizardMetaFormattedNames}
-              summaryCalloutFields={wizardMetaCalloutFields}
+              // summaryTitle="New Drift Analysis"
+              // summaryKeyOverrides={wizardMetaFormattedNames}
```
jruddell, line 69: Why should we remove the summary overview?

kevinylee1991: I think so, the info isn't useful and the summary display doesn't work well when there is also the DriftSummary component in the DriftAnalysisResult component

jruddell: The drift summary should be in the wizard summary. That's the place for this content

src/app/plugins/kubernetes/components/rbac/profiles/drift/ClusterRoleBindingAnalysisDialog.tsx
```diff
@@ -85,9 +90,11 @@ const ClusterRoleBindingAnalysisDialog = ({ clusterRoleBinding, onClose }: Props
           </FontAwesomeIcon>
         </div>
         <div className={classes.details}>
-          <div>
-            <span className={classes.label}>Cluster Name:</span>
-            <span>{clusterName}</span>
+          <div className={classes.gridColumn}>
```
xantorres, line 93: Do we really need double wrapping things? The less div nesting the better, in my opinion.

jruddell (approved):

PR #1733, opened 2022-04-20
Update design of add and edit repository forms

knnguy:
### Add Repository Form
Screen Shot 2022-04-21 at 4 53 56 PM

### Add Repository Form for Free Tier Users
Screen Shot 2022-04-21 at 4 51 36 PM

### Edit Repository Form
Screen Shot 2022-04-21 at 4 55 31 PM


src/app/plugins/kubernetes/components/common/MultiClusterSelectionDropdown.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const filteredClusters = useMemo(() => (filterFn ? filterFn(allClusters) : allClusters), [
+    allClusters,
+  ])
+
+  const options = useMemo(() => {
+    return [...projectAs({ label: 'name', value: 'uuid' }, filteredClusters)]
+  }, [filteredClusters])
+
+  return (
+    <MultiDropdown
```
jruddell, line 73: is the only real difference here us rendering a different dropdown element? I feel like this could be consolidated with the current cluster dropdown, but rather deterministically figure out which dropdown to render `const DropdownComponent = isMultiSelect ? MultiDropdown : Dropdown` or something like that

src/app/core/elements/dropdown/MultiDownshift.tsx
```diff
@@ -61,27 +63,43 @@ function dropdownReducer<T>(selectedItems: T[], { type, payload }: DropdownReduc
...
 
   const toggleItem = useCallback((item) => {
     dispatch({ type: 'toggleItem', payload: item })
   }, [])
-  const selectItem = useCallback((item) => {
-    dispatch({ type: 'select', payload: item })
-  }, [])
+  const selectItem = useCallback(
+    (item) => {
+      const isAlreadySelected = selectedItems.find((selectedItem) => equals(selectedItem, item))
```
xantorres, line 82: This shouldn't be necessary, still if you want to do it you should do it in the `dropdownReducer`, not here. And just do `selectedItems.includes(payload)` instead of all this

src/app/core/elements/dropdown/MultiDropdown.tsx
```diff
@@ -51,11 +56,31 @@ export default function MultiDropdown<V, T extends DropdownItemSpec<V> = Dropdow
     [onChange],
   )
 
+  const setSelectFuncRef = (select) => {
+    selectFuncRef.current = { select }
```
jruddell, line 60: I'm not a fan of this pattern, and its something we will be getting rid of when the form system is cleaned up. Is the goal here to expose the select function? could we give the dropdown a selection cb to call if it exists?

src/app/core/elements/dropdown/MultiDropdown.tsx
```diff
@@ -51,11 +56,31 @@ export default function MultiDropdown<V, T extends DropdownItemSpec<V> = Dropdow
...
+  }
+
+  useEffect(() => {
+    if (!initialValues || !items) return
+    initialValues.forEach((v) => {
```
xantorres, line 65: We shouldn't be setting initialValues with a `useEffect`, please just initialize the reduce in MultiDownshift with the provided values instead of all of this

jruddell (approved):

PR #1737, opened 2022-04-21
Minor Fixes on Bareos Create Cluster

patil-pratik-87:
Before :- 
Screenshot 2022-04-20 at 10 41 17 PM

After :-
 Screenshot 2022-04-21 at 10 39 45 AM


jruddell:
worker nodes are not required on a single master cluster, so we shouldn't be requiring validation on this

jruddell:
Please remove the validation changes you did. We have a validation system that checks for things. This here isn't using that. The form validation should be fine for now

src/app/core/elements/tooltip/Tooltip.tsx
```diff
@@ -183,6 +183,10 @@ const Tooltip = styled(
       opacity: 1,
       transform: getMenuTransform(1)({ vertAlign, horizAlign }),
     },
+    '& *': {
```
jruddell, line 186: Generally styling this way is not efficient. Can we update the children to be styled correctly instead?

jruddell (changes requested):
Let's clean this up some, please remove the validation stuff you did. I'll show you how this stuff works on a call later. 

Styling with a * selector should be avoided on highly used components because of the excess css styling in each instance

patil-pratik-87:
> Please remove the validation changes you did. We have a validation system that checks for things. This here isn't using that. The form validation should be fine for now

done

patil-pratik-87:
> Please remove the validation changes you did. We have a validation system that checks for things. This here isn't using that. The form validation should be fine for now

Yeah, we ve fixed the form validation and used that now :)

src/app/core/elements/tooltip/Tooltip.tsx
patil-pratik-87: true, used more specific classes

jruddell (approved):

PR #1738, opened 2022-04-21
Task/more dialogs

jruddell:
image
image
image
image
image
image


patil-pratik-87 (approved):

src/app/plugins/account/components/userManagement/users/enable-disable-user-dialog.tsx
```diff
@@ -41,20 +34,20 @@ const EnableDisableUserDialog = ({ rows: [user], onClose }) => {
...
-    <Dialog open onClose={onClose}>
-      <Progress loading={loading || updating}>
-        <DialogTitle>{action} User</DialogTitle>
-        <DialogContent>
-          <DialogContentText>
-            {action} user {user.username}?
-          </DialogContentText>
-        </DialogContent>
-        <DialogActions>
+    <Modal
```
xantorres, line 37: Shouldn't we prevent all user interactions in the modal when "loading"? What happens if the user clicks cancel when loading/updating? That's why we had a `Progress` component overlaying all the modal before.

jruddell: Progress in the component is a pretty bad piece of UX. I'll disable the close button when loading

xantorres: And what will happen when we have more interactive elements in the Modal? Eg forms, inputs, etc

jruddell: the form has been already submitted, and the buttons are disabled with a loading / working affordance. I don't think we should do much more in the old form system currently. When we redesign the forms having a submitting state where they are all disabled / locked would be a good user experience and we can do that there. But I dont see the immediate need to change this currently in the old system, more or less not a good use of our time.

PR #1741, opened 2022-04-21
Fixed infinite loop for SSO login and removed blur for group custom m…

kevinylee1991:
…apping inputs

Screen Shot 2022-04-22 at 12 06 57 PM


src/app/core/elements/input/Input.tsx
```diff
@@ -127,7 +130,7 @@ const Input = ({
         <input
           {...rest}
           onFocus={handleFocusToggle}
-          onBlur={handleFocusOut}
+          onBlur={disableBlur ? noop : handleFocusOut}
```
xantorres, line 133: Why not just check for `handleFocusOut` existence instead of adding another additional prop?

xantorres, line 133: This feels like a quick workaround and not actually addressing the underlying issue.

kevinylee1991: John is looking into this onBlur issue

jruddell (approved):

PR #1742, opened 2022-04-22
Update the links in deployed apps page in cluster details

knnguy:
Screen Shot 2022-04-21 at 5 16 52 PM


jruddell (approved):
We have a links file for each plugin, it would be best if we can define these in there and import them here. Helps us when we go to update links

PR #1748, opened 2022-04-22
Fix issue with initial wizard context missing data due to loading fla…

kevinylee1991:
…g and rendering interaction

I added this change to fix problems with the Wizard component where the initial context would be loaded into the Wizard too fast due to data loading flags initializing as false before any data came in.

Most noticeable areas are the tenant & user edit wizards. Go to edit a user or tenant for the first time, and the roles table will not be loaded in. Revisit it and it will be populated. With this change, they will be loaded in the first time around.


src/app/core/components/wizard/Wizard.js
```diff
@@ -43,6 +43,17 @@ class Wizard extends PureComponent {
     dispatch(clientActions.setSidebarPane('default'))
   }
 
+  componentDidUpdate(prevProps, prevState) {
```
jruddell, line 46: We shouldn't be doing this. Why not prevent the wizard from rendering with initial values until its loaded? or even set its context once the data loads in separately?

src/app/core/actions/Action.ts
```diff
@@ -118,10 +118,10 @@ abstract class Action<
           }
           const result = await this.callback(params)
 
+          this.postProcess(result, params, options)
```
kevinylee1991, line 121: This change is so that the data is dispatched to the cache before the loading flag is switched to false

src/app/plugins/account/components/userManagement/tenants/EditTenantPage.js
```diff
@@ -108,44 +113,46 @@ const EditTenantPage = () => {
...
-              </ValidatedForm>
-            </WizardStep>
-            <WizardStep stepId="users" label="Users and Roles">
-              <ValidatedForm
-                title="Users and Roles"
-                fullWidth
-                initialValues={wizardContext}
-                onSubmit={setWizardContext}
-                triggerSubmit={onNext}
+      {!loadingSomething && (
```
kevinylee1991, line 116: Need this so that the Wizard gets rerendered when loadingSomething flips from false -> true -> false

src/app/core/actions/Action.ts
xantorres: Why?

src/app/plugins/account/components/userManagement/tenants/EditTenantPage.js
xantorres: Maybe you could wrap the `Wizard` with a `withProgress` HoC and set `renderContentOnMount=false`, could you try that?

src/app/core/actions/Action.ts
kevinylee1991: When we have something like this

`const { loading: loadingUsers } = useListAction(listUsers)
  const users = useSelector(usersSelector)`
  
If loading is set to false before users data is populated, and we have the Wizard component which takes an initialContext that is gated by that loading flag, then we end up passing through an initialContext that doesn't have the required data. I think it makes sense that the data should be populated in the cache first before the loading flag is toggled to say that it has finished loading

src/app/plugins/account/components/userManagement/tenants/EditTenantPage.js
kevinylee1991: sure I can try

src/app/core/actions/Action.ts
xantorres: But loading will still be `undefined`, how does thks change solve that?

kevinylee1991: loading always returns a boolean now -- in useListAction:

`const loading = !!loadingKeys.length`

xantorres: so if it was returning `false` already, what does this change do?

kevinylee1991: The problem is that currently, loading is set to false when the data is not written to the store yet, so the selector returns empty array when supposedly it has been loaded in already. By setting loading to false after the data gets written to the store, then we will have the data present immediately when loading shifts from true -> false

src/app/core/hooks/useListAction.ts
```diff
@@ -48,7 +48,7 @@ const useIsDataLoadingSelector = (...cacheKeys): string[] => {
 
   return useMemo(() => {
     const loadingStatusByKey = zipObj(cacheKeys, keysLoadingStatus)
-    return Object.keys(filter(isTruthy, loadingStatusByKey))
+    return Object.keys(filter(either(isTruthy, isUndefined), loadingStatusByKey))
```
kevinylee1991, line 51: the undefined change

src/app/plugins/account/components/userManagement/tenants/TenantsListPage.tsx
```diff
@@ -38,7 +39,8 @@ const columns: GridViewColumn<SelectorModel>[] = [
 
 export default function TenantsListPage() {
   const { params, getParamsUpdater } = usePrefParams({})
-  const { message, loading, reload } = useListAction(listTenants)
+  // Need to pass clusterId in params for the namespaces dependency
+  const { message, loading, reload } = useListAction(listTenants, { params: { clusterId: allKey } })
```
kevinylee1991, line 43: Have to pass the required index params for the namespaces dependency, or else that dependency is never called

src/app/core/hooks/useListAction.ts
xantorres: This is wrong and unnecessary, since the result will be the same as not adding that at all.

src/app/plugins/kubernetes/components/infrastructure/nodes/download-ova-walkthrough.tsx
```diff
@@ -77,7 +78,7 @@ const DownloadOvaWalkthrough = (): JSX.Element => {
   const { getUserPrefs } = useScopedPreferences()
   const { currentTenant, currentRegion } = getUserPrefs(session.username)
 
-  const { loading: loadingTenants } = useListAction(listTenants)
+  const { loading: loadingTenants } = useListAction(listTenants, { params: { clusterId: allKey } })
```
xantorres, line 81: We are repeating ourselves too much here, why we don't just consider that not passing a `clusterId` is the same as passing an `allKey` so we avoid all this repeated code everywhere?

src/app/plugins/kubernetes/components/namespaces/new-actions.ts
```diff
@@ -24,6 +24,7 @@ export const namespaceActions = new ActionsSet<DataKeys.Namespaces>({
 export const listNamespaces = namespaceActions.add(
   new ListAction<DataKeys.Namespaces, { clusterId: string }>(async (params) => {
     Bugsnag.leaveBreadcrumb('Attempting to get namespaces', params)
+    console.log('-----------------trying to get namespaces------------------')
```
xantorres, line 27: `console.log` here

src/app/utils/fp.ts
```diff
@@ -39,6 +39,7 @@ export const emptyObj = {}
 export const pluck = (key) => (obj) => obj[key]
 export const isTruthy = (x) => !!x
 export const isFalse = (x) => !x
+export const isUndefined = (x) => x === undefined
```
xantorres, line 42: We could just use ramda [isNil](https://ramdajs.com/docs/#isNil)

src/app/core/hooks/useListAction.ts
xantorres: I was about to say we should also check that data is going to be fetched (by looking at cachedParams) but this is probably enough since theoretically every time this returns items that are `undefined` in the loading store it will implicitly means that data is going to be fetched since it has never been fetched before.

src/app/plugins/kubernetes/components/infrastructure/nodes/download-ova-walkthrough.tsx
kevinylee1991: sure i'll try that out

kevinylee1991: It seems that might not be a great idea because this means in the list pages, before the cluster picklist loads in, the list page will try to get its resources across all clusters. What about when we assign dependencies when creating a ListAction, we optionally assign defualt parameters for the dependency?

xantorres: So as discussed early in Slack, let's get rid of the `!clusterId` condition in the namespaces `ListAction`:
image

jruddell (approved):

PR #1750, opened 2022-04-25
Implement ControlledGrid in ClusterHostChooser and misc fixes

xantorres:
- Refactor `ClusterHostChoosers` to use `ControlledGrid` and fix some issues
- Add `withTooltip` HoC
- Fix multiple TS issues
- Refactor `withFormContext` and `ValidatedFormInput` to TS
- Refactor `UserMultiSelect` and `GroupMultiSelect` to TS
- Fix some `MultiSelect` issues
- Fix some styling issues
- Fix issue with tooltip not showing up on Modals

image


src/app/core/components/listTable/PerPageControl.js
```diff
@@ -19,7 +19,7 @@ const PerPageControl = ({ classes, value, onRowsPerPageChange, rowsPerPageOption
   const options = rowsPerPageOptions.map((x) => ({ label: x, value: x }))
   return (
     <FormControl className={classes.root}>
-      <Tooltip message={'Rows per page'}>
+      <Tooltip title={'Rows per page'}>
```
xantorres, line 22: This one is using the MUI Tooltip, so it needs to use `title` instead of `message`

src/app/core/components/validatedForm/withFormContext.tsx
```diff
@@ -42,7 +69,7 @@ const ValidatedFormInput = ({
     removeField,
     validateField,
     showErrorsOnBlur,
-  } = useContext(ValidatedFormContext)
+  } = useContext(ValidatedFormContext as any)
```
xantorres, line 72: This isn't ideal, but since we'll be implementing a new form component soon, I didn't want to dig much further down the rabbit hole.

src/app/core/components/MultiSelect.tsx
```diff
@@ -92,143 +93,140 @@ const useStyles = makeStyles((theme: Theme) =>
...
-      showSelectDeselectAllOption,
-    },
-    ref: React.Ref<HTMLDivElement>,
-  ) => {
-    const classes = useStyles()
+const getOptionsHeight = (maxOptions) => maxOptions * 28
 
-    const [term, setTerm] = useState('')
-    const fuse = useMemo(() => new Fuse(options, FUSE_OPTIONS), [options])
+// @deprecated Please use MultiDropdown instead
```
jruddell, line 121: we shouldn't be updating this component, why not just render a multi dropdown instead?

src/app/core/components/validatedForm/FormFieldSection.tsx
```diff
@@ -103,10 +114,10 @@ export const useStyles = makeStyles<Theme>((theme) => ({
   spacer: {
     marginTop: 2,
     marginBottom: 14,
-    marginLeft: 24,
+    // marginLeft: 24,
   },
   content: {
-    marginLeft: 24,
+    // marginLeft: 24,
```
jruddell, line 120: each form section indents itself, why remove this?

src/app/plugins/kubernetes/components/infrastructure/clusters/ScaleMastersPage.tsx
```diff
@@ -50,14 +52,19 @@ const useStyles = makeStyles((theme: Theme) => ({
 
 // Quorum Messages
 const zeroMasterMsg =
-  'You cannot remove master node from a single master cluster.  If you wish to delete the cluster, please choose the ‘delete’ operation on the cluster on the infrastructure page instead.'
+  'You cannot remove master node from a single master cluster.  ' +
+  'If you wish to delete the cluster, please choose the ‘delete’ operation ' +
```
jruddell, line 56: you can use a string template here and enter down with a new line to achieve the same. Is a bit cleaner

src/app/core/components/MultiSelect.tsx
xantorres: Well, this component has some minor design and behavioral differences over the MultidropDown so I wanted to wait a little before fully migrating everything to use that. 
In any case these fixes and changes to use Typescript will make it easier to replace it when we decide to.

src/app/core/components/validatedForm/FormFieldSection.tsx
xantorres: The problem was it was double indenting the form in a few places, and it was looking pretty bad, at least in my laptop screen. I can take a look though to see if I can fix the double wrapping with a FormSection instead of doing this.

jruddell: What? Where are we double wrapping?

jruddell: The solution here shouldn't be to remove a margin when the design calls for it. Is this because were not removing the `elevated` prop from `ValidatedForm`?

jruddell (approved):

PR #1751, opened 2022-04-25
Reverted sso login failure redirect link, and fix page crash on profi…

kevinylee1991:
…le page for SSO users

Might bring back the infinite login loop if SSO is not set up properly, but we need it bc the 302 redirect doesn't work


src/app/plugins/account/components/user-settings/my-profile-page.tsx
```diff
@@ -97,10 +97,13 @@ const MyProfilePage = () => {
   const loadingSomething = updatingUser || updatingPassword
 
   useEffect(() => {
+    if (isSsoToken) {
```
jruddell, line 100: you should also verify that the field setter exists `if (isSsoToken && userIdFormFieldSetter.current)`

PR #1753, opened 2022-04-26
UX-1663/Disable button to delete cloud provider when associated impor…

kevinylee1991:
…ted clusters are present, and fixed a strange display bug caused by a ref

Fixed:
Screen Shot 2022-04-25 at 5 39 03 PM

Before:
Screen Shot 2022-04-25 at 5 40 38 PM

Tooltip:
Screen Shot 2022-04-25 at 6 20 32 PM


src/app/core/elements/grid/hooks/useGridSelectableRows.ts
```diff
@@ -27,6 +27,7 @@ export interface BatchActionButtonProps<T> {
 
 export interface GridBatchActionSpec<T> {
   cond?: (selectedItems: T[]) => boolean
+  tooltipFn?: (selectedItems: T[]) => string
```
xantorres, line 30: This should be handled by the button itself, not by the Grid.

src/app/core/components/validatedForm/ListTableField.tsx
```diff
@@ -22,46 +22,48 @@ interface Props extends ValidatedFormProps {
...
-    checkboxCond,
-    loading,
-    uniqueIdentifier,
-    onReload,
-    searchTargets = ['name'],
-    orderBy,
-    dataTestId,
-  } = props
+export default withFormContext<any[], Props>(
+  React.forwardRef(function(props, ref) {
```
xantorres, line 26: Do we really need to keep doing this? With the new Tooltip this shouldn't be necessary

src/app/core/elements/grid/hooks/useGridSelectableRows.ts
jruddell: Agreed, we have examples of custom tooltips on batch actions already inside the clusters list page, would that suffice?

src/app/core/components/validatedForm/ListTableField.tsx
jruddell: The whole reason why we were wrapping everything with forward ref was for material ui tooltips but our new system just checks a client rect for positioning. Forwarding refs shouldn't be needed anymore

src/app/core/elements/grid/hooks/useGridSelectableRows.ts
kevinylee1991: Oh yeah, that example had what I was looking for

jruddell (approved):

PR #1764, opened 2022-04-29
Refactoring the useDataLoader to useListAction- Part1

patil-pratik-87:


src/app/core/elements/dropdown/useStyles.ts
```diff
@@ -13,7 +13,7 @@ export default makeStyles<Theme, DropdownStylesProps>((theme) => ({
     position: 'relative',
     margin: 0,
     maxWidth: ({ width = dropdownDefaultWidth }) => width,
-    width: ({ compact }) => (compact ? 'unset' : '100%'),
+    width: ({ compact, width = dropdownDefaultWidth }) => (compact ? 'unset' : width),
```
patil-pratik-87, line 16: https://user-images.githubusercontent.com/63217589/165889781-ef0d1308-d969-4ea1-8e1e-46fc1130dd3c.mov

 @xantorres @jruddell  made these changes to fix the above issue, as you can see the dropdown was resizing every time to the length of the content in the menu. Instead gave a standard size every time, see if its appropriate

jruddell: Theres a few ways this can be handled, probably what makes the most sense is some form of uniform sizing in these wizards. In the past i've found the largest element in the menu and updated that as the default width as well. But it is probably better if these are all uniform. What do you think @xantorres ?

src/api-client/Qbert.ts
```diff
@@ -662,14 +662,16 @@ class Qbert extends ApiService {
   })
   getAddonVersions = async (clusterId) => {
     const url = `/clusters/${clusterId}/addonversions`
-    return this.client.basicGet<GetAddonVersions>({
+    const addonVersions = await this.client.basicGet<GetAddonVersions>({
```
jruddell, line 665: Lets add a todo comment here clarifying what needs to change once a get one action is created

src/app/core/components/RegionMenu.tsx
```diff
@@ -44,7 +44,7 @@ const RegionMenu = (props) => {
     if (currentRegion && find(propEq('id', currentRegion), regions)) {
       return currentRegion
     }
-    return pipe(head, prop('id'))(regions)
+    // return pipe(head, prop('id'))(regions)
```
jruddell, line 47: is this supposed to be commented out?

src/app/core/elements/dropdown/useStyles.ts
jruddell: I think the only change I would really do here is have the validated form updated the min-width of our dropdown instead of putting it in the dropdown itself

src/app/core/utils/getDataSelector.ts
```diff
@@ -53,7 +53,7 @@ const getDataSelector = memoizeShallow(
         // Search either for the provided params or an empty object (it means we fetched for ALL items)
         if (
           isNilOrEmpty(indexBy) ||
-          find(either(equals(providedIndexedParams), equals(emptyObj)))
+          either(equals(providedIndexedParams), equals(emptyObj))(storeData)
```
jruddell, line 56: getDataSelector is very essential, and used in a lot of places, what is this change fixing?

src/app/plugins/account/components/userManagement/common/RolesPicklist.js
```diff
@@ -2,16 +2,19 @@ import React, { useMemo, useEffect, forwardRef } from 'react'
...
+import { listMngmRoles } from '../roles/new-actions'
+import { useSelector } from 'react-redux'
+import { managementRolesSelector } from '../roles/selectors'
 
 // We need to use `forwardRef` as a workaround of an issue with material-ui Tooltip https://github.com/gregnb/mui-datatables/issues/595
 const RolesPicklist = forwardRef(
   ({ allRoles, id, loading, onChange, value, selectFirst, disabled, ...rest }, ref) => {
-    const [roles, rolesLoading] = useDataLoader(mngmRoleActions.list, { allRoles })
+    const { loading: rolesLoading } = useListAction(listMngmRoles)
+    const roles = useSelector(managementRolesSelector)
```
jruddell, line 17: shouldn't this be a selector with params where you pass allRoles through?

src/app/plugins/kubernetes/components/common/CloudProviderRegionPicklist.js
```diff
@@ -1,18 +1,22 @@
...
+import { cloudProviderDetailsSelector } from '../infrastructure/cloudProviders/selectors'
 
 // We need to use `forwardRef` as a workaround of an issue with material-ui Tooltip https://github.com/gregnb/mui-datatables/issues/595
 // TODO use the CloudProviders enum to strongly type this type field once moved to TS
 const CloudProviderRegionPicklist = forwardRef(
   ({ cloudProviderId, type, onChange, ...rest }, ref) => {
-    const [details, loading] = useDataLoader(loadCloudProviderDetails, {
-      cloudProviderId: cloudProviderId,
+    const { loading } = useListAction(loadCloudProviderDetails, {
+      params: { cloudProviderId },
```
jruddell, line 16: @xantorres do we need to pass the cloudProviderId through in the list action and the selector as well?

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/edit-cluster-page/ClusterAddonsPage.tsx
```diff
@@ -88,13 +86,12 @@ export default function ClusterAddonsPage({
...
   const existingClusterAddons = useSelectorWithParams(clusterAddonsSelector, { clusterId })
 
-  // TODO: Use useGetAction here once that's been created
-  const [[currentAddonVersions = emptyObj], loadingAddonVersions] = useDataLoader(
-    loadAddonVersions,
-    {
-      clusterId,
-    },
-  )
+  const { loading: loadingAddonVersions } = useListAction(listAddonVersions, {
```
jruddell, line 89: keep the todo here

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/new-actions.ts
```diff
@@ -106,7 +106,7 @@ export const deleteClusterAddon = clusterAddonActions.add(
 )
 
 export const addonVersionActions = new ActionsSet<DataKeys.AddonVersions>({
-  uniqueIdentifier: 'metadata.uid',
+  uniqueIdentifier: 'metadata.id',
```
jruddell, line 109: 👍

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/selectors.ts
```diff
@@ -13,6 +13,11 @@ export const parseClusterAddonParams = (params: AddonOverrideParam[]) => {
   return params.reduce((acc, item) => ({ ...acc, [item.name]: item.value }), {})
 }
 
+export const addonVersionsSelector = createSharedSelector(
```
jruddell, line 16: What is the need for this?

src/app/plugins/kubernetes/components/infrastructure/clusters/import/RegionsChecklist.tsx
```diff
@@ -33,9 +35,8 @@ const RegionsChecklist = ({ cloudProviderId, onChange, value, className, cluster
     }
   }
 
-  const [details] = useDataLoader(loadCloudProviderDetails, {
-    cloudProviderId: cloudProviderId,
-  })
+  const { loading } = useListAction(loadCloudProviderDetails)
```
jruddell, line 38: Why aren't we passing the cloudProviderId here through as params like we do on the other instances? 
image

src/app/plugins/kubernetes/components/infrastructure/clusters/selectors.ts
```diff
@@ -210,3 +210,8 @@ export const makeClusterEventsSelector = (defaultParams = {}): AppSelector<IEven
     )(events)
   })
 }
+
+export const supportedRoleVersionsSelector = createSharedSelector(
```
jruddell, line 214: what is the purpose of this?

src/app/plugins/kubernetes/components/infrastructure/importedClusters/gke-nodes.tsx
```diff
@@ -42,7 +43,7 @@ const ListPage = ({ ListContainer }) => {
         })),
       ),
       flatten,
-    )(nodeGroups)
+    )((nodeGroups as unknown) as NodePool[])
```
jruddell, line 46: You shouldn't need to do this. We should chat about it on a call and see if theres a better way than double casting

src/app/plugins/kubernetes/components/infrastructure/regions/selectors.ts
```diff
@@ -0,0 +1,8 @@
+import getDataSelector from 'core/utils/getDataSelector'
+import { createSharedSelector } from 'core/utils/selectorHelpers'
+import DataKeys from 'k8s/DataKeys'
+
+export const regionsSelector = createSharedSelector(
```
jruddell, line 5: What is the purpose of these selectors? they aren't doing anything

src/app/plugins/kubernetes/components/onboarding/add-cloud-provider-page.tsx
```diff
@@ -27,6 +27,9 @@ const useStyles = makeStyles<Theme>((theme) => ({
   header: {
     marginTop: '0px',
   },
+  button: {
```
jruddell, line 30: lets not change things inside onboarding related to styling as this is yet to be redesigned

src/app/plugins/kubernetes/components/rbac/RolePicklist.js
```diff
@@ -1,20 +1,28 @@
...
 // We need to use `forwardRef` as a workaround of an issue with material-ui Tooltip https://github.com/gregnb/mui-datatables/issues/595
 const RolePicklist = forwardRef(
-  ({ clusterId, namespace, loading, onChange, selectFirst, showAllRoleTypes, ...rest }, ref) => {
-    const [roles, rolesLoading] = useDataLoader(roleActions.list, { clusterId, namespace })
-    const [clusterRoles, clusterRolesLoading] = useDataLoader(clusterRoleActions.list, {
-      clusterId,
+  ({ clusterId, loading, onChange, selectFirst, showAllRoleTypes, ...rest }, ref) => {
+    const { loading: rolesLoading } = useListAction(listRoles, {
+      requiredParams: ['clusterId'],
+      params: { clusterId },
```
jruddell, line 18: isn't the namespace needed here like it is in the data loader?

src/app/plugins/kubernetes/components/storage/csi-drivers/csiDriverPicklist.tsx
```diff
@@ -26,23 +28,25 @@ interface Props {
 }
 
 const CSIDriversPicklist: React.ComponentType<Props> = forwardRef<HTMLElement, Props>(
-  (
-    { clusterId, name = 'csiDriversList', label = 'CSI Drivers', formField = false, ...props },
-    ref,
-  ) => {
-    const [csiDrivers, loading] = useDataLoader(csiDriverActions.list, { clusterId })
+  ({ clusterId, name = 'csiDriversList', label = 'CSI Drivers', onChange, compact, ...props }) => {
+    type ActionParams = InferActionParams<typeof listCSIDrivers>
```
jruddell, line 32: Why are we defining a type inside of a component?

src/app/plugins/kubernetes/components/storage/csi-drivers/csiDriverPicklist.tsx
```diff
@@ -26,23 +28,25 @@ interface Props {
...
+    const requiredParams: Array<keyof ActionParams> = ['clusterId']
```
jruddell, line 33: are these `const`'s for `requiredParams` and `params` necessary? generally we just pass this through directly to the action/selector

src/app/core/utils/getDataSelector.ts
xantorres: Yeah, the change is going to optimizer things a little bit, although this is wrong, the `find` should not have been removed, just the params added

src/app/plugins/kubevirt/components/virtual-machines/add/ReviewStep.tsx
```diff
@@ -32,7 +34,8 @@ const diskSourceValues = {
 }
 
 const RenderClusterName = ({ clusterId }) => {
-  const [clusters] = useDataLoader(clusterActions.list)
+  useListAction(listClusters)
```
jruddell, line 37: I know this wasn't using a loading state before, but its probably best to include a loading affordance when waiting on something to load async

src/app/core/elements/dropdown/useStyles.ts
patil-pratik-87: We can have the `min-width` in the containers that would use the dropdowns but then everyone who consumes the dropdown will need to set a `min-width` where  1. It's a lot of work again and again and 2. It might not look consistent everywhere.

src/app/plugins/account/components/userManagement/common/RolesPicklist.js
xantorres: It would probably be better if `allRoles` would come from another selector itself, instead of having to pass it through params, that way we would avoid using a `makeMngmRolesSelector` at all

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/new-actions.ts
xantorres: To be honest I don't like using paths for uniqueIdentifiers like this, we should be mapping the id to the root level of the object instead of having to access nested properties like this.

src/app/core/elements/dropdown/useStyles.ts
jruddell: I'm saying override the default min-width in our ValidatedForm. Which would be only done in one place, and wouldn't change the current dropdown width behavior

src/app/plugins/account/components/userManagement/common/RolesPicklist.js
patil-pratik-87: yeah, the `RolesPicklist` is used in two other places where `allRoles` was not used, hence ditched it. It's probably best to remove the prop all togethor

src/app/plugins/kubevirt/components/virtual-machines/add/ReviewStep.tsx
patil-pratik-87: sure

src/app/plugins/kubernetes/components/storage/csi-drivers/csiDriverPicklist.tsx
xantorres: In general it is a much better practice to declare these `requiredParams` outside of the render method  so that way we avoid redeclaring a new array on every render.

src/app/core/utils/getDataSelector.ts
jruddell: `find(whereEq(providedIndexedParams))(storeData)`

src/app/core/elements/dropdown/useStyles.ts
patil-pratik-87: okay, got it

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/selectors.ts
patil-pratik-87: removed

src/app/plugins/kubernetes/components/infrastructure/clusters/selectors.ts
patil-pratik-87: removed

src/app/plugins/kubernetes/components/infrastructure/regions/selectors.ts
patil-pratik-87: removed

src/api-client/Qbert.ts
```diff
@@ -662,14 +662,17 @@ class Qbert extends ApiService {
...
       url,
       version: 'v4',
       options: {
         clsName: this.getClassName(),
         mthdName: 'getAddonVersions',
       },
     })
+    // converted the single object to array as the current useListAction is only meant to work with array responses.
+    // This can be reverted and used with a different ListAction which handles object repsonse,when we have one
```
jruddell, line 674: Keep the `TODO` or `fixme` text here so we can find it when searching. Also the message shouldn't say `used with a different ListAction` it should instead say `used with a GetAction`

src/app/core/components/validatedForm/ValidatedForm.js
```diff
@@ -33,6 +33,9 @@ const styles = (theme) => ({
     '& .tooltip-container': {
       maxWidth: '400px',
     },
+    '& .dropdown-container': {
+      width: '400px',
```
jruddell, line 37: dont specify widths unless there is a really specific use case for it. use `maxWidth`

src/app/plugins/kubernetes/components/app-catalog/apps/components/DeployAppModalForm.tsx
```diff
@@ -288,5 +288,9 @@ const useStyles = makeStyles<Theme>((theme) => ({
     gridTemplateColumns: '220px 164px',
     alignItems: 'end',
     gridGap: theme.spacing(2),
+    minWidth: 'auto',
+    '& .dropdown-container': {
+      width: 'auto',
```
jruddell, line 293: I suspect you needed to change this here because its `width`, is the override needed when this is changed to `maxWidth`?

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/new-actions.ts
jruddell: Were you going to do this change @patil-pratik-87 ?

src/app/plugins/kubernetes/components/onboarding/add-cloud-provider-page.tsx
jruddell: I still see the change here, did you forget to push the fix?

src/app/plugins/kubernetes/components/pods/pod-container-details-page.tsx
```diff
@@ -1,17 +1,25 @@
...
   const { id, clusterId, containerName } = match.params
-  const [pods, loading] = useDataLoader(podActions.list, { clusterId })
-  const pod = useMemo(() => pods.find((pod) => pod.id === id) || {}, [id, pods])
+  const { loading } = useListAction(listPods, {
+    params: { clusterId },
+    requiredParams: ['clusterId'],
+  })
+
+  const pods = useSelectorWithParams(podsSelector, { clusterId })
+  const pod = useMemo(() => pods.find((pod) => pod.id === id) || {}, [id, pods]) as IPodSelector
```
jruddell, line 22: do you need to cast this?

jruddell (approved):

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/new-actions.ts
patil-pratik-87: Didn't intend to make the change here, I thought it could be bigger change and best done separately

src/app/plugins/kubernetes/components/onboarding/add-cloud-provider-page.tsx
patil-pratik-87: Naah kept it intentionally. The stuff was misaligned and thought doesn't hurt to keep it here, since it's already in. Should you still think we should remove it?

src/api-client/Qbert.ts
patil-pratik-87: sure, that makes sense

src/app/core/components/validatedForm/ValidatedForm.js
patil-pratik-87: `maxWidth` will default the width to the content of the first selected option which will not solve the issue and 
`minWidth` also introduces an issue where the dropdown will take up the entire parent width since it's mentioned 100% in the dropdown. So had to use a static width

src/app/plugins/kubernetes/components/pods/pod-container-details-page.tsx
patil-pratik-87: let me try and fix the typings here, then

src/app/plugins/kubernetes/components/onboarding/add-cloud-provider-page.tsx
jruddell: we may end up deleting the whole onboarding folder. So yes lets not make any changes to the design currently till this is figured out from product

patil-pratik-87: okay

PR #1765, opened 2022-04-29
Fixed issue with dropdowns with tooltip being shrunk down, and update…

kevinylee1991:
…d vm wizard to use newer section components


src/app/core/components/validatedForm/ValidatedForm.js
```diff
@@ -22,14 +22,17 @@ const styles = (theme) => ({
...
+    alignItems: 'stretch',
     marginBottom: theme.spacing(1),
     '& .MuiFormControl-root.validatedFormInput': {
       width: '100%',
       maxWidth: '400px',
       marginTop: theme.spacing(1.5),
       marginBottom: theme.spacing(1.5),
     },
+    '& .tooltip-container': {
+      maxWidth: '400px',
```
xantorres, line 34: Tooltips are not rendered inside the elements anymore, they use a common container, so this will probably not work.

jruddell: the tooltip container wrapped around a dropdown inside our forms shrinks the dropdowns size, this mitigates that issue

jruddell (approved):

PR #1784, opened 2022-05-13
UX-1876/Disable adding an SAML group when SSO is not configured

kevinylee1991:


jruddell:
Should we also prevent SSO login from the login page when this mapping is not setup?

kevinylee1991:
> 

I can check to see if the API is available when not authenticated

kevinylee1991:
Unfortunately it seems we are not able to get that information without being authenticated

Screen Shot 2022-05-16 at 3 46 29 PM

jruddell (approved):

PR #1787, opened 2022-05-17
Fix action names by using a property instead of class name

xantorres:


kevinylee1991:
Oh I was actually just looking into this too, I had the same solution of adding a get name per Action type.

kevinylee1991 (approved):

jruddell (approved):

PR #1789, opened 2022-05-17
Fixed crashing issue with dropdown in group creation wizards

kevinylee1991:
Was crashing bc onChange was not a function in asyncDropdown (wasn't defined). To avoid crashes, should we perhaps add a defaultValue for onChange along with a console error or bugsnag error to notify us that the onChange wasn't defined?


src/app/plugins/account/components/ssoManagement/groups/GroupCustomMappingsFields.tsx
```diff
@@ -114,7 +114,7 @@ const renderCustomMappings = ({ values, setFieldValue, classes }) => (
               </div>
             </>
           }
-          multiline
```
xantorres, line 117: this is not necessary, in React just adding a prop implicitly sets it as "true"

jruddell: Also, why would we pass `true` through as a string?

kevinylee1991: `react-dom.development.js?61bb:86 Warning: Received `true` for a non-boolean attribute `multiline`.

If you want to write it to the DOM, pass a string instead: multiline="true" or multiline={value.toString()}.
    at input`

I'm fine removing it though

kevinylee1991: Looking closer it doesn't look like default input element even has a multiline attribute so i dunno why it was complaining

kevinylee1991: I changed it back

xantorres: You should just remove it entirely then, we refactored a lot of elements to new components so that property might have been deleted.

jruddell (approved):

PR #1795, opened 2022-05-19
Fixed bug with ListContainer columns where newly added columns were u…

kevinylee1991:
…nable to show up on the table

To reproduce bug, try adding a new column to a table -- the new column will not show up. With this code it will update appropriately.


xantorres (changes requested):

src/app/core/elements/grid/hooks/useGridManagedColumns.ts
```diff
@@ -146,6 +147,25 @@ export default function useGridManagedColumns<T, C extends GridManagedColumnSpec
...
+    // are added to the end of the list
+    const keysList = columnSpecs.map(({ key }) => key)
+    // Checks for any difference between columnsOrder and keysList
+    if (symmetricDifference(columnsOrder, keysList).length) {
+      const newOrderedColumns = union(intersection(columnsOrder, keysList), keysList)
+      const newVisibleColumns = union(
+        intersection(visibleColumnKeys, keysList),
+        difference(keysList, columnsOrder),
+      )
+      dispatch({ type: 'setVisibleColumnKeys', payload: { columnKeys: newVisibleColumns } })
```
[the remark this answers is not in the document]
kevinylee1991: makes sense, moved it into the initial values setter

src/app/core/elements/grid/hooks/useGridManagedColumns.ts
```diff
@@ -121,21 +147,23 @@ export default function useGridManagedColumns<T, C extends GridManagedColumnSpec
     disableColumnHiding,
   }: GridManagedColumnsConfig<T, C>,
 ): [Array<ManagedColumnsParsedGridRow<T>>, GridManagedColumnsProps] {
-  const initialState = useMemo(
-    () => ({
+  const initialState = useMemo(() => {
+    return {
```
xantorres, line 151: Why not just filter the `columnSpecs` here before doing anything else so you won't need to do the filtering twice afterwards?

src/app/core/elements/grid/hooks/useGridManagedColumns.ts
```diff
@@ -110,6 +110,32 @@ const columnsReducer: Reducer<ColumnsState, ColumnsReducerAction> = (
...
+  // If there is a difference between saved columns and column spec, remove columns
+  // that are no longer in the spec and add in any new columns in the spec
+  if (symmetricDifference(columnsOrder, keysList).length) {
+    return union(intersection(columnsOrder, keysList), keysList)
+  }
+  return columnsOrder
+}
+
+const getVisibleColumns = ({ visibleColumns, columnSpecs, columnsOrder }) => {
+  const visibleKeys = columnSpecs.reduce((acc, { display = true, key }) => {
```
xantorres, line 124: I think you could pass the already filtered `orderedColumnKeys` so that you won't have to do a symetricDifference and union again here.

kevinylee1991: orderedColumnKeys gets defined at the same time together with visibleColumnKeys though and they're only called once at the start. Maybe we could store it in a variable, but because of the conditionals in the initialState useMemo, perhaps this wouldn't be all that worth it?

PR #1800, opened 2022-05-20
UX-1867/Fix provisioner bug in create storage class form

knnguy:
When creating a storage class using in-tree provider for AWS clusters, the YAML is now automatically populated with “provisioner: [kubernetes.io/aws-ebs](http://kubernetes.io/aws-ebs)"

Screen Shot 2022-05-20 at 1 38 29 PM


src/app/plugins/kubernetes/components/storage/storage-classes/add-page/storage-class-config-fields.tsx
```diff
@@ -28,10 +28,10 @@ const cloudStorageProvisionerValues = switchCase(
   {
     [CloudProviders.Aws]: 'aws-ebs',
     [ClusterCloudPlatforms.EKS]: 'aws-ebs',
-    [CloudProviders.Azure]: 'azure-disk',
```
jruddell, line 31: I dont think we should be changing this. aws-ebs is for aws only. so these values should be correct

knnguy: https://platform9.atlassian.net/jira/software/projects/UX/boards/96?selectedIssue=UX-1867

The ticket says AWS, Azure, and GKE

jruddell: Ticket is wrong. It's just the `kubernetes/io` prefix. The storage driver is different for these providers

knnguy: Ok, fixed

jruddell (approved):

PR #1803, opened 2022-05-24
Added default params for namespaces, and added console error for miss…

kevinylee1991:
…ing dependency params


kevinylee1991:
Didn't see any other dependencies that had more required params than its parent, other than tenants/namespaces

src/app/core/actions/Action.ts
```diff
@@ -107,26 +109,27 @@ abstract class Action<
     },
   )
 
-  public async call(params: P = this.defaultParams, options: Partial<O> = {}): Promise<void | R> {
+  public async call(params: P = {} as P, options: Partial<O> = {}): Promise<void | R> {
+    const combinedParams = this.memoizedParams({ ...this.defaultParams, ...params })
```
xantorres, line 113: dWe should probably initialize `defaultParams` to `{}` in the constructor, otherwise unexpected things could happen here

xantorres (approved):

PR #1806, opened 2022-05-25
Fixed bug where clusters can be stuck loading kubevirt addons

kevinylee1991:
The old action doesn't have kubevirt addons as a dependency, so could get stuck loading clusters trying to load kubevirt addons depending on which page user hard refreshes on


src/app/core/containers/AuthContainer.tsx
```diff
@@ -271,9 +271,9 @@ const AuthContainer = () => {
   useEffect(() => {
     const getClusters = async () => {
       setLoadingClusters(true)
-      await clusterActions.list()
+      await listClusters.call({})
```
xantorres, line 274: You don't need to pass an empty obj here

jruddell (approved):

kevinylee1991:
Ended up getting resolved in another PR, closing this

PR #1809, opened 2022-05-27
UX-1833/Move scaling nodes to a modal form

knnguy:
### Scale BareOS
Screen Shot 2022-05-26 at 9 55 06 AM

Screen Shot 2022-05-26 at 9 55 27 AM

### Scale AWS/Azure
Screen Shot 2022-05-26 at 9 55 43 AM


jruddell (approved):
can you delete the block chooser component as well? I believe it was only used for these scale pages and is now obsolete

PR #1810, opened 2022-05-31
Feature/automation/local data testid v5.6

Komaljadhav12:
cluster-delete
multi-master


Komaljadhav12:
screencapture-file-C-Users-Komal-J-Documents-5-5-branch-pf9-ui-plugin-cypress-reports-html-2022-05-31-22_20_32

cypress/integration/infrastructure/single-master-cluster-create.spec.ts
```diff
@@ -10,8 +10,122 @@ describe('Single-Master Cluster Create Flow', () => {
...
+    })
 
-  it('should be able to create Single Master Cluster', () => {
-    cy.ensureSingleMasterClusterCreate()
+    cy.fixture('apiCalls/apiCalls').then((calls) => {
+      this.calls = calls
+    })
+  })
+
+  beforeEach(function() {
```
jruddell, line 23: is there supposed to be more than one beforeEach function?

Komaljadhav12: Yes. When we use .then it will change the cypress context to asynchronous JavaScript so if fetching the data and using it immediately in same hook cause the error. So we have separated out extraction and its usage in different before each hooks.

PR #1834, opened 2022-06-29
UX-1924/Fix batch cluster upgrade API body to use number instead of string for batchUpgradePercent property

kevinylee1991:


src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/PercentageClusterUpgrade.tsx
```diff
@@ -60,8 +63,10 @@ export const PercentageClusterUpgradeAddonField = ({ values, setValues, defineFi
         id="batchUpgradePercent"
         label="Percentage of Worker nodes to upgrade in parallel"
         placeholder="Specify % of nodes to upgrade in parallel"
-        value={wizardContext?.batchUpgradePercent ? wizardContext?.batchUpgradePercent : 10}
```
kevinylee1991, line 63: I had to move where this default value was set because the initial value wouldn't be populated in the wizardContext otherwise.  Lmk if you think there's a better place to set it than in the useEffect above

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/PercentageClusterUpgrade.tsx
```diff
@@ -34,6 +34,9 @@ export const PercentageClusterUpgradeAddonField = ({ values, setValues, defineFi
   //automatically as a field to the ValidatedForm
   useEffect(() => {
     defineField('batchUpgradePercent')({ validations: [requiredValidator] })
+    if (!wizardContext?.batchUpgradePercent) {
+      setWizardContext({ batchUpgradePercent: 10 })
```
jruddell, line 38: could this just be the initial value for the wizard? or does it have to be set in a specific effect?

jruddell (approved):

src/app/plugins/kubernetes/components/infrastructure/clusters/cluster-addons/PercentageClusterUpgrade.tsx
kevinylee1991: @jruddell it could be the initial value for the wizard, I actually had done that first. But the wizard is located in another file so I thought perhaps this method may be more appropriate. Any preference?

PR #1836, opened 2022-07-01
Fixed bug with global params influencing namespace picklists in wizards

kevinylee1991:
Not sure about other picklists that may be affected. For example, if there were a spot where we needed to select a pod in a picklist inside a wizard, we would need to update that too.


src/app/core/utils/getDataSelector.ts
```diff
@@ -63,6 +63,9 @@ const getDataSelector = memoizeShallow(
       selectParams,
       selectGlobalParams,
       (storeData, storeParams, params, globalParams) => {
+        // set useGlobalParams false only if explicitly set false
+        // @ts-ignore -- what's the best way to deal with unknown typing?
```
kevinylee1991, line 67: Typescript complains about using params?.useGlobalParams bc params is of unknown typing. What's the easiest way to solve this sort of typescript issue? I think I run into this quite often and was curious if there was a best way to solve

xantorres: You have to return the proper type in the `selectGlobalParams` selector

xantorres (approved):

PR #1841, opened 2022-07-05
UX-1691/Fixed delete dialog for jobs & cronjobs to handle both types

kevinylee1991:
Outdated
Screen Shot 2022-07-05 at 3 46 21 PM

New
Screen Shot 2022-07-06 at 5 40 33 PM


src/app/plugins/kubernetes/components/cronjobs/DeleteJobDialog.tsx
```diff
@@ -0,0 +1,58 @@
...
+      cluster_uuid: job?.clusterId,
+      name: job?.name,
+      namespace: job?.namespace,
+    })
+    onClose(true)
+    history.push(routes.cronjobs.list.path())
+  }, [job])
+
+  return (
+    <Dialog open onClose={onClose} onClick={stopPropagation}>
```
jruddell, line 37: we should be using our modal instead of MUI's Dialog

jruddell (approved):

PR #1843, opened 2022-07-07
Task/tweak infra styles

patil-pratik-87:


src/app/core/containers/Row.tsx
```diff
@@ -3,25 +3,27 @@ import Theme from 'core/themes/model'
...
+  minItemWidth?: string
+  gap?: number
 }
 
 const useStyles = makeStyles<Theme, StyleProps>((theme) => ({
   rowLayout: {
     display: 'grid',
     gridTemplateColumns: ({ minItemWidth }) => `repeat(auto-fill, minmax(${minItemWidth}px, 1fr))`,
-    gap: theme.spacing(1),
+    gap: ({ gap }) => gap,
```
jruddell, line 15: default gap to a value incase we don't pass one

PR #1850, opened 2022-07-14
Converted most kubernetes resources to use global params

kevinylee1991:
Didn't touch Nodes, Monitoring, or Apps.

For Nodes, it looks like we may want to keep more control over the ability to show nodes across all clusters, so I hesitated to remove the filter from the table.

For monitoring, I had the same thought, that it is meant to be controlled more by the page rather than the global parameters.

For apps, it looked like Deployed apps could be converted, but the cards and the details page would need to have information about which cluster & namespace the application belongs to. Wanted to get your feedback on this before touching this.

Currently no information is available about the cluster or namespace anywhere on the cards.
Screen Shot 2022-07-14 at 3 30 23 PM

Updated:

Deployed Apps:
Screen Shot 2022-07-19 at 3 33 04 PM
App Catalog:
Screen Shot 2022-07-19 at 3 33 20 PM
Deployed App Details:
Screen Shot 2022-07-19 at 3 32 21 PM


src/app/plugins/kubevirt/components/virtual-machines/selectors.ts
```diff
@@ -97,6 +97,7 @@ export const presetsSelector = createSharedSelector(
   getDataSelector<DataKeys.VirtualMachinePresets>(DataKeys.VirtualMachinePresets, ['clusterId']),
   allClustersSelector,
   (rawPresets, clusters) => {
+    console.log(rawPresets, 'raw presets')
```
jruddell, line 100: console.log

src/app/plugins/kubernetes/components/rbac/profiles/create/RolesForm.tsx
```diff
@@ -115,10 +114,15 @@ const RolesForm = ({ wizardContext, setWizardContext }) => {
     params: { ...params, clusterId: wizardContext.baseCluster },
     requiredParams: ['clusterId'],
   })
-  const data = useSelectorWithParams(paramsRolesSelector, {
+  const data = useSelectorWithParams(rolesSelector, {
     clusterId: wizardContext.baseCluster,
     namespace: params.namespace,
+    useGlobalParams: false,
   })
+  const filteredData =
```
jruddell, line 122: can you explain why we need to apply this filtering here vs using the global params and the selector to filter?

kevinylee1991: @jruddell just updated

jruddell (approved):

PR #1856, opened 2022-07-21
UX-1994/ create capi cluster schemas + modified wizardContext structure for capi

patil-pratik-87:


src/app/core/components/validatedForm/ValidatedForm.js
```diff
@@ -21,7 +21,7 @@ export const ValidatedFormProvider = ValidatedFormContext.Provider
 const styles = (theme) => ({
   root: {
     display: 'flex',
-    flexFlow: 'column wrap',
+    flexFlow: 'column',
```
patil-pratik-87, line 24: This was causing extra space at the bottom in the wizard steps, I 've done a basic sanity nothing is breaking, but  maybe still we can look for a better solution @jruddell

PR #1864, opened 2022-07-26
Added issues overview to infrastructure plugin dashboard

kevinylee1991:
Even if we don't get something in for the Default Clusters section, maybe we should just put a coming soon on some boxes or something underneath to take up some vertical real estate?

Screen Shot 2022-07-28 at 5 15 44 PM

------ Old ------
Screen Shot 2022-07-25 at 6 07 52 PM
Screen Shot 2022-07-25 at 6 13 11 PM


jruddell:
Go ahead and do the default clusters as the next task so we fill this dashboard up some more

src/app/plugins/infrastructure/components/DashboardPage.tsx
```diff
@@ -1,5 +1,155 @@
-import React from 'react'
+import React, { useCallback, useMemo } from 'react'
```
jruddell, line 1: now that you have a dashboard folder, you should probably put this page in there. I would also call this file `Insights.tsx`

src/app/plugins/infrastructure/components/dashboard/helpers.ts
```diff
@@ -0,0 +1,142 @@
...
+  status: string
+  message: string
+  link: string
+  resourceType: string
+  resourceName: string
+  id: string
+}
+
+export const getActiveIssues = (allClusters, nodes): Issue[] => {
+  const clusters = allClusters.filter(pathEq(['clusterType'], 'normal'))
```
jruddell, line 18: this is quite a heavy operation. can you make a new selector that handles this logic? probably best to break each type down to an individual selector (one for nodes, imported clusters, clusters). And then you can either combine in the view or have one combined selector that returns a final array of issues

jruddell (commented):
looking good! Do we need those new image files? I had asked design to use fontawesome icons, can you verify these are not in the fontawesome library? As a general pattern I'd like to prevent shipping image assets unless we have to.

src/app/plugins/infrastructure/components/dashboard/helpers.ts
xantorres: Agree, we shouldn't do any data filtering or manipulation in the view, that's what selectors are for.

kevinylee1991:
> looking good! Do we need those new image files? I had asked design to use fontawesome icons, can you verify these are not in the fontawesome library? As a general pattern I'd like to prevent shipping image assets unless we have to.

There is one for the clouds, but I couldn't find one for the other two. Should I swap the cloud one?

jruddell (approved):

PR #1865, opened 2022-07-26
Add models for CAPI clusters, AWSCluster, and NodeletControlPlane

knnguy:


src/app/plugins/infrastructure/components/clusters/capi/actions.ts
```diff
@@ -26,7 +26,7 @@ export const deleteCapiCluster = capiClusterActions.add(
   new DeleteAction<DataKeys.CapiClusters, { namespace: string; name: string }>(
     async ({ namespace, name }) => {
       Bugsnag.leaveBreadcrumb('Attempting to delete CAPI cluster', { namespace, name })
-      return qbert.deleteCapiCluster(namespace, name)
+      await qbert.deleteCapiCluster(namespace, name)
```
jruddell, line 29: shouldn't we be returning the promise here?

jruddell (approved):

src/app/plugins/infrastructure/components/clusters/capi/actions.ts
xantorres: It shouldn't be necesary, because of the `await` here

xantorres (approved):

PR #1869, opened 2022-07-26
UX-1928/Added parsed YAML templates to the WizardSummary

xantorres:
Also converted `CodeMirror` component to TS and splitted it in a base `CodeMirror` without form context and `CodeMirrorField` with it

You can provide a `yamlTemplates` prop either to the `<Wizard />` component or when creating the `<WizardStep />`'s

Also when submitting the Wizard, you can get the parsed templates as the second parameter of the `onComplete` callback provided to the `<Wizard />`


jruddell (approved):

jruddell:
Can you add a screenshot of what it looks like and add it to the PR?

src/app/core/components/wizard/WizardSummary.tsx
```diff
@@ -24,6 +25,7 @@ export default function WizardSummary<T>({
   icon,
   error,
   fields = {} as any,
+  yamlTemplates = [],
```
patil-pratik-87, line 28: @xantorres Can we have `yamlTemplates` specific to each `WizardStep` somehow? The design shows separate yamls for separate steps. Let's talk about it when we chat

patil-pratik-87 (approved):

PR #1901, opened 2022-08-23
Fixes of the smoke or regression suite and updated the json files.

Komaljadhav12:
add-pod
deployment
pod-delete
service


vijay-nandagavali (commented):
Reviewed at my end. Changed Approved.

Sanket36 (approved):

PR #1903, opened 2022-08-24
UX_2069/Enabled edit for CAPI node group YAMLS

knnguy:
Added the edit functionality to yamls in the node group details


src/app/plugins/infrastructure/components/clusters/capi/machine-deployment/model.ts
```diff
@@ -21,7 +19,7 @@ export interface IMachineDeploymentSelector extends MachineDeployment {
   configRef: MachineDeployment['spec']['template']['spec']['bootstrap']['configRef']
   machineTemplate: IAwsMachineTemplateSelector
   config: IConfigSelector
-  resources: (MachineDeployment | NodeletConfig | AwsMachineTemplate)[]
+  resources: unknown[]
```
knnguy, line 22: I had to put the types as unknown because of the typing errors

patil-pratik-87: okay

src/app/plugins/infrastructure/components/clusters/capi/aws-machine-templates/selectors.ts
```diff
@@ -26,7 +26,7 @@ export const awsMachineTemplatesSelector: AppSelector<IAwsMachineTemplateSelecto
         spotInstanceMaxPrice: mt.spec?.template?.spec?.spotMarketOptions
           ? mt.spec?.template?.spec?.spotMarketOptions.maxPrice
           : 'Max Price',
-        resources: mt,
+        resources: [mt],
```
patil-pratik-87, line 29: Why are we using an array here and some other places? It was not an array as it is a machine template and won't have references to other resources like machine deployments, machine pools or awsmanagedmachinepools

patil-pratik-87 (approved):

src/app/plugins/infrastructure/components/clusters/capi/aws-machine-templates/selectors.ts
knnguy: I was thinking the `resources` property should always be an array to consistent. In other resources like Machine Deployments, for example, the `resources` field is an array but then in a resource like this, AWSMachineTemplate, the same property, `resources` is not. Plus the property name is `resources` which implies it has multiple values. If making all the resources property an array doesn't make sense, we can change this to something like `resource`. 
It's just a bit confusing when we have to check whether or not the resources property is

PR #1920, opened 2022-08-31
UX-2093/Fixed deselect dropdown item bug

knnguy:
Issue:
User could not deselect a cluster from the global cluster filter.

Fix:
Major issue was that we were doing comparisons using object references which always returned false. 

https://user-images.githubusercontent.com/23369276/187785235-6412ad2d-b357-4847-a39c-937965477c5e.mov


patil-pratik-87 (approved):

patil-pratik-87:
@knnguy  Also make sure you have a PR for a 5.6 backport

PR #1951, opened 2022-09-07
UX-2137, UX-2138/Fixed issue where min/max props in TextFields were not working correctly

knnguy:
**Issue:**
Blank screen occurred when the user scaled down the number of Kubevirt VMs to 0. This should not even be allowed. The min number of VMs is 2.

Props like `min`, `max`, `maxLength` were being passed to the input components through the `inputProps` property but it doesn't work. The `inputProps` property is for the MUI `TextField` component but we have our own `Textfield` component now which does not accept that property

**Fix:**
Pass in the props separately to the input component


kevinylee1991 (approved):
I ran into the same for kubevirt, we might have some more occurrences elsewhere

PR #2002, opened 2022-09-12
UX-2176/Don't require cloud init script in Kubevirt config

knnguy:


src/app/plugins/kubevirt/components/virtual-machines/add/ConfigureStep.tsx
```diff
@@ -15,13 +14,38 @@ import {
...
   fullWidth: {
     width: '100% !important',
   },
 }))
 
+const moizedYamlLoad = moize(jsYaml.safeLoad, {
+  maxSize: 10,
+})
+
+export const yamlValidator = new FieldValidator(
```
kevinylee1991, line 30: Wonder if it might be better to add a condition to the yamlValidator that returns true if the input is empty? Since requiredValidator should be added to any fields that are required.

kevinylee1991 (approved):

PR #2015, opened 2022-09-13
Fix custom path string function in yaml template parser to handle mul…

kevinylee1991:
…tiple string interpolations


knnguy (approved):

src/app/utils/YamlTemplateParser.ts
```diff
@@ -74,10 +74,15 @@ export default class YamlTemplateParser {
   // Todo: add some error-proofing
   private getCustomPathString(str) {
     const customString = str.substr(17)
+    // Regex lookarounds would be more convenient, but are not in safari
     const regex = /\${(.*?)\}/g
-    const matches = regex.exec(customString)
-    if (!matches?.length) return
-    const paths = [matches[1]]
+    const matches = customString.match(regex)
```
xantorres, line 79: You could probably replace all of this function with a simple `.replace`:
`customString.replace(/\${(.*?)\}/g, (match, path) => pathStr(path, this.values) || '');`

xantorres (changes requested):

src/app/utils/YamlTemplateParser.ts
kevinylee1991: oh cool thanks, this helped me learn about regex some more (my weakness lol)

PR #2064, opened 2022-09-19
UX-2264/Fix rbac selectors by using allClustersSelector

xantorres:


knnguy (approved):

knnguy:
Were you able to get any data, like roles, clusterroles, etc. for the capi clusters? I was looking into an issue with rbac and capi and it seems like capi doesn't use the same rbac apis as the qbert clusters

xantorres:
> Were you able to get any data, like roles, clusterroles, etc. for the capi clusters? I was looking into an issue with rbac and capi and it seems like capi doesn't use the same rbac apis as the qbert clusters

Yes, I was able to list all the items

PR #2122, opened 2022-10-11
UX-2321/Add a metadata field in routes and create metadata param to display global filters

xantorres:
Also removed Storage and RBAC links from Infrastructure menu


src/app/plugins/infrastructure/index.ts
```diff
@@ -225,64 +215,12 @@ infrastructurePlugin.registerRoutes([
     link: { path: '/cloud-providers/google/(edit)?/:id?', exact: true },
     component: GoogleCloudProvidersListPage,
   },
-  {
-    name: 'CSI Drivers',
```
patil-pratik-87, line 229: Are we sure we don't need it inside `Infrastructure`?

patil-pratik-87 (approved):

src/app/plugins/infrastructure/index.ts
xantorres: Infrastructure links are already present in Kubernetes section

PR #2136, opened 2022-10-21
UX-2315/Remove the addons field from AWSManagedControlPlane y…

knnguy:
…aml template


src/app/plugins/infrastructure/components/clusters/aws/capi/schema/eks/aws-managed-control-plane.json
```diff
@@ -10,17 +10,14 @@
     "addons": [
```
patil-pratik-87, line 10: we can remove the entire `addons` property.

jruddell (approved):

PR #2140, opened 2022-10-25
UX-2067/Update the kubevirt VM & VMI Details views

kevinylee1991:
VM Details Page
Screen Shot 2022-10-25 at 3 49 46 PM
Screen Shot 2022-10-25 at 3 49 56 PM

VMI Details Page
Screen Shot 2022-10-25 at 3 50 14 PM
Screen Shot 2022-10-25 at 3 50 22 PM


src/app/core/utils/routes/kubevirtRoutes.ts
```diff
@@ -7,6 +7,10 @@ import {
 import { IVirtualMachineInstanceDetailsPageTabs } from 'app/plugins/kubevirt/components/virtual-machines/vmi-model'
 import { Route } from 'core/plugins/route'
 
+const metadata = {
```
jruddell, line 10: id rather we define this as an opt out. disable it from individual screens that we want

jruddell (approved):
looks good. in general i think we should default to always show the global filters and then opt out in screens that we dont want them. less boilerplate this way

src/app/core/utils/routes/kubevirtRoutes.ts
kevinylee1991: I think that makes sense too, was just following the pattern here to match the existing. I think will need to make another jira to change it from showGlobalFilters to hideGlobalFilters and swap things to opposite

PR #2148, opened 2022-10-28
UX-1978/Implemented Kubevirt Dashboard and enabled retrieval of data from selectors with indexes using array values

kevinylee1991:
Screen Shot 2022-10-28 at 4 45 45 PM

Different DU with 2 healthy clusters

Screen Shot 2022-10-28 at 5 18 58 PM


src/app/core/utils/getDataSelector.ts
```diff
@@ -40,6 +41,20 @@ const arrayizeIndexBy = (args: [string, string | string[]]) => {
   return [dataKey, ensureArray(indexBy)]
 }
 
+const testIndexedParamsItem = (params) => (item) => {
+  const combinations = paramsCartesianProduct(params)
```
xantorres, line 45: to prevent doing the params cartesian product on each iteration put this function call in the outer closure (the one with `params` argument)

src/app/core/utils/getDataSelector.ts
```diff
@@ -75,10 +90,13 @@ const getDataSelector = memoizeShallow(
         )(params as ParamsType)
         // If the provided params are already cached
         // Search either for the provided params or an empty object (it means we fetched for ALL items)
-        if (isNilOrEmpty(indexBy) || find(whereEq(providedIndexedParams))(storeData)) {
+        if (
+          isNilOrEmpty(indexBy) ||
+          testIndexedParamsCollection(providedIndexedParams)(storeData)
```
xantorres, line 95: You should call this before the `if` statement and assign it to a `const` that you will use both in the `if` and in the `filter` below

src/app/plugins/kubevirt/components/dashboard/KubevirtDashboard.tsx
```diff
@@ -0,0 +1,353 @@
...
+        label: 'PVC',
+        value: allPvcs?.length,
+        loading: pvcsLoading,
+      },
+      {
+        label: 'Data Volume',
+        value: allDataVolumes?.length,
+        loading: dataVolumesLoading,
+      },
+    ]
```
xantorres, line 319: I think we are doing too much data processing in the view.
Maybe this will look a bit cleaner if we put all the selectors in a unified selector instead of calling a bunch of `useSelectorWithParams` and then mixing them together in this `useMemo`.

src/app/core/utils/getDataSelector.ts
```diff
@@ -40,6 +41,28 @@ const arrayizeIndexBy = (args: [string, string | string[]]) => {
   return [dataKey, ensureArray(indexBy)]
 }
 
+const memoizeCombinations = memoize((params) => paramsCartesianProduct(params))
+
+const testIndexedParamsItem = (params) => {
```
jruddell, line 46: I'm not a huge fan of using `test` in the name of helper functions like this. Since this is more or less finding the items/combinations you should change `test` to `find`. I.e...  `testIndexedParamsItem` -> `findIndexedParamsItem`. Same goes for the function below. `testIndexedParamsCollection` -> `findIndexedParamsCollection`

jruddell (approved):

PR #2150, opened 2022-10-30
UX-1013/airgap external requests

jruddell:


src/app/core/containers/AppController.ts
```diff
@@ -1,6 +1,6 @@
 import axios from 'axios'
 import { isNilOrEmpty, pathStrOr } from 'utils/fp'
-import Bugsnag from '@bugsnag/js'
+import Bugsnag from 'utils/bugsnag'
 import { DocumentMetaCls } from 'core/components/DocumentMeta'
```
xantorres, line 4: Importing DocumentMeta here (something from a view) in a controller is an antipattern and a bad idea

src/app/core/containers/AppController.ts
```diff
@@ -58,7 +58,7 @@ const isWhitelistedUrl = (pathname): boolean => {
   return whiteListedUrls.includes(pathname)
 }
 
-const getFeatures = async () => {
+export const getFeatures = async () => {
```
xantorres, line 61: This should be in a helpers file or in its own file, having it here and imported in other files could lead to some unexpected behavior or errors.

xantorres: And probably this function should be memoized (at least for some seconds) since we will call this multiple times from different places during startup

xantorres (approved):

PR #2151, opened 2022-11-01
Add Grid rows menu

xantorres:
image

With a single action:
image

Only icon:
image

How to define:
image


jruddell (approved):
Looks great!

